PR #764, opened 2023-06-28
fix: BTC client - modify inbound scan

kevinssgh:
# Description

Modified the rate at which the blocks are scanned from bitcoin. Previously only one block was scanned and processed at a fixed interval, but now it can scan multiple blocks up to a limit if the client is trying to catch up. 

Closes: (https://github.com/zeta-chain/zeta-node/issues/759)

## Type of change

# How Has This Been Tested?

This was tested by forcing a large number of blocks to be generated in the smoketest and observe how long it takes to process. When the maximum is set to 100 blocks per scan, it takes about 2 seconds to scan through the empty blocks. 

- [x] Tested CCTX in localnet
- [x] Go integration tests

### DIFF
--- a/zetaclient/bitcoin_client.go
+++ b/zetaclient/bitcoin_client.go
@@ -67,6 +67,7 @@ const (
 	minConfirmations = 1
 	chunkSize        = 500
 	maxHeightDiff    = 10000
+	maxScanDiff      = 100
 )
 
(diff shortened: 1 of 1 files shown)


brewmaster012:
@kevinssgh I imagine this could be closed, in favor of a simpler approach that reduces observeInTx interval to something like 10s?

PR #810, opened 2023-07-14
fix: expand CctxAllPending query RPC to include pending cctxs below NonceLow

brewmaster012:
closes: #807

### DIFF
--- a/x/crosschain/keeper/keeper_cross_chain_tx.go
+++ b/x/crosschain/keeper/keeper_cross_chain_tx.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"cosmossdk.io/math"
 	"fmt"
-
 	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/query"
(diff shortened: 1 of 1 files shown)


x/crosschain/keeper/keeper_cross_chain_tx.go
```diff
@@ -136,12 +135,30 @@ func (k Keeper) CctxAllPending(c context.Context, req *types.QueryAllCctxPending
...
+	if startNonce < 0 {
+		startNonce = 0
 	}
-	//sends := k.GetAllCctxByStatuses(ctx, []types.CctxStatus{types.CctxStatus_PendingOutbound, types.CctxStatus_PendingRevert})
-	sends := make([]*types.CrossChainTx, 0)
+	for i := startNonce; i < p.NonceLow; i++ {
+		res, found := k.GetNonceToCctx(ctx, tss.TssPubkey, int64(req.ChainId), i)
+		if !found {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("nonceToCctx not found: nonce %d, chainid %d", i, req.ChainId))
+		}
```
kingpinXD, line 151: Can this condition ever happen? 
If it does happen, how to we get out of this? A cctx with an older nonce would need to be created.

brewmaster012: it should not happen unless serious mess up happened

kingpinXD (approved):

x/crosschain/keeper/keeper_cross_chain_tx.go
kingpinXD: We can just log it then instead of returning , the return will cause clients to be stuck in an infinite loop

PR #812, opened 2023-07-16
feat: defensive: bitcoin deduplication via simulating nonce

brewmaster012:
# Description

This PR introduces a mechanism to ensure that each scheduled bitcoin transaction will at most generate
one bitcoin transaction, much like on EVM clients where nonce is included into transaction so multiple
re-trys with the same nonce can only succeed at most once. 

from Charlie Chen:
1. Introduced nonce-mark utxo to align TSS signers with same nonce.
2. To bootstrap nonce-mark UTXO feature, the nonce-mark utxo checking is temporarily disabled because TSS doesn't have that utxo yet. We'll enforce that in the next upgrading.

Closes: #488 

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


kevinssgh (commented):
I really like the idea of utilizing bitcoin's inputs and outputs to enforce the sequence and uniqueness of each withdraw. There might be an edge case when we try to introduce this to the network where the nonce is already > 0 and the previous input isn't there. I suppose we will have to manually add it with a deposit.

Anyway I like the idea

kingpinXD (approved):

kingpinXD (approved):

PR #822, opened 2023-07-18
fix: Hot fix v5.0.2

kevinssgh:
# Description
Fixing crashes on zetaclient due to nil references and malformed data in database.

### DIFF
--- a/zetaclient/evm_signer.go
+++ b/zetaclient/evm_signer.go
@@ -245,6 +245,9 @@ func (signer *EVMSigner) TryProcessOutTx(send *types.CrossChainTx, outTxMan *Out
 			logger.Error().Msgf("Unknown chain: %d", send.GetCurrentOutTxParam().ReceiverChainId)
 			return
 		}
+	} else {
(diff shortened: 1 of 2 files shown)


zetaclient/evm_signer.go
```diff
@@ -245,6 +245,9 @@ func (signer *EVMSigner) TryProcessOutTx(send *types.CrossChainTx, outTxMan *Out
 			logger.Error().Msgf("Unknown chain: %d", send.GetCurrentOutTxParam().ReceiverChainId)
 			return
 		}
+	} else {
+		logger.Info().Msgf("Transaction doesn't need to be processed status: %d", send.CctxStatus.Status)
```
kingpinXD, line 249: This log line might cause a lot of spam until we fix the zetocored query

brewmaster012: probably ok as INFO level is restricted in dd. 
also there is a chance that it can resolve itself, depending on which tx was last confirmed
see https://github.com/zeta-chain/node/pull/823

kingpinXD (approved):

PR #825, opened 2023-07-19
feat: generate typescript bindings

fadeev:
`make typescript` should generate TS bindings in `./typescript`.

Using https://buf.build/blog/protobuf-es-the-protocol-buffers-typescript-javascript-runtime-we-all-deserve

We probably don't need TS bindings in the repo. This should be generated automatically on release and published on `npm`.


lumtis:
Just wondering, why is it in draft?

fadeev:
> Just wondering, why is it in draft?

I want the fullstack team to check it out and try using to see if we have everything in place before merging.

GMaiolo:
Could we remove the `_pb` suffix for prolixity? I feel it makes sense for consumers; unless it causes an issue within the node repository.

---

Is it possible to have an `index.d.ts` file that exports everything within its folder? e.g.:

```
crosschain/
  index.d.ts
  foo.d.ts
  bar.d.ts
  baz.d.ts
  
```
  
With `index.d.ts` being something like the following:

```
export * from "./foo"
export * from "./bar"
export * from "./baz"
```

This way, we can consume types directly from the known folder (`crosschain`) and we don't need to know the specific file name (`foo`, `bar`, `baz`).

fadeev:
@GMaiolo done! ✅

fadeev:
@lucas-janon can you, please, add an npm token, so we can set it up to publish automatically when a release is created.

GMaiolo:
@fadeev could we manually publish a `0.0.1` version in NPM so I can test the usage?

fadeev:
I can't publish, because I don't have an npm token. Have you tried with `yarn link` first?

lumtis:
Can we put as ready for review @fadeev ?

Makefile
```diff
@@ -161,6 +161,11 @@ proto:
 	@buf format -w
 .PHONY: proto
 
+typescript:
```
lumtis, line 164: I think we can include it into `make generate` commands

lumtis (commented):
Did you run `make typescript` with the latest changes?
I get several modified files when I run the command

Makefile
fadeev: https://github.com/zeta-chain/node/pull/825/commits/5397f6050c8a913040c1bbf494f1045d4267f27f

lumtis (approved):

fadeev:
@GMaiolo can you, please, review the code generated in this PR and let us know if this matches your requirements? Thanks!

GMaiolo (approved):
LGTM!

Issue #815, opened 2023-07-17
refactoring the keeper state for crosschain module


fadeev:
```[tasklist]
### Tasks
```

PR #829, opened 2023-07-20
refactor: migrate crosschain keeper

kingpinXD:
# Description

- Moves `Node_Accounts,` ` KeyGen,` and `Permission_Flags` from cross-chain to observer module
- Adds a check for the observer set , which executes at every block, If this check fails (the observer set changes for some reason), It will  
       - Pause Inbound Tx ( as new TSS will need to be generated) 
       - Reset KeyGen back to pending status
       
- Once the inbound pauses, the following step would need to be done
   - Modify the node account list to remove/add new observers. (Can add a Policy based TX for this in a separate PR) 
   - Set KeyGen Height for a new block to generate a new TSS 
   
Closes:
https://github.com/zeta-chain/node/issues/815
https://github.com/zeta-chain/node/issues/830

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


kingpinXD:
Planning to modify setLastBlockCount to only execute when observer set changes ,instead of every block.

x/observer/abci.go
```diff
@@ -0,0 +1,36 @@
...
+		totalObserverCountCurrentBlock += len(observer.ObserverList)
+	}
+	if totalObserverCountCurrentBlock < 0 {
+		ctx.Logger().Error("TotalObserverCount is negative at height", ctx.BlockHeight())
+		return
+	}
+	if totalObserverCountCurrentBlock == int(lastBlockObserverCount.Count) {
+		return
+	}
+	ctx.Logger().Error("LastBlockObserverCount does not match the number of observers found at current height", ctx.BlockHeight())
```
ws4charlie, line 29: Does this error mean that something bad happened in the last block? Is this an unexpected error and where/how are we handling this error for now? btw, was just curious

kingpinXD: Yeah, this line of code would only be executed when the observer set changes. The change in the validator set can be triggered from one of these hooks for example 
https://github.com/zeta-chain/zeta-node/blob/7eadeaf4339b570b9e983872cab7e7d50aa4eb37/x/observer/keeper/hooks.go#L10-L12

We are just logging the error , and pausing the inbound for now

ws4charlie: Thanks. I thought it's the code for long-term.

brewmaster012 (approved):

PR #833, opened 2023-07-21
feat: WIP - Post Blame data from TSS keysign response

kevinssgh:
# Description

Will add detailed description soon

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/observer/keeper/keeper_utils.go
```diff
@@ -8,6 +8,66 @@ import (
...
+	}
+	for _, obs := range observerMapper.ObserverList {
+		if obs == address {
+			return true, nil
+		}
+	}
+	return false, errors.Wrap(types.ErrNotAuthorized, fmt.Sprintf("address: %s", address))
+}
+
+func (k Keeper) FindBallot(ctx sdk.Context, index string, chain *common.Chain, observationType types.ObservationType) (ballot types.Ballot, isNew bool, err error) {
```
kingpinXD, line 47:
```suggestion
func (k Keeper) FindBallot(ctx sdk.Context, index string, chain *common.Chain, observationType types.ObservationType,Threshold sdk.Dec) (ballot types.Ballot, isNew bool, err error) {
```
The threshold defined in params is only for inbound and outbound TX , we can reuse it for Blame as well, but it would be better to maintain a list of thresholds in params .

Tss voter for example uses 100%  , https://github.com/zeta-chain/zeta-node/blob/2437225c0294635a5bd150e50d61853157b2e47d/x/crosschain/keeper/keeper_tss_voter.go#L61-L61

proto/observer/blame.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+package zetachain.zetacore.observer;
+
+import "observer/observer.proto";
+
+option go_package = "github.com/zeta-chain/zetacore/x/observer/types";
+
+message Node {
```
kingpinXD, line 8: Can we use the existing NodeAccount structure ? And some fields to it.

That would also allow us to check MsgAddBlame to ensure the nodes being blamed are still part of the network.

x/observer/types/events.go
```diff
@@ -0,0 +1,26 @@
...
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(BallotCreated,
+			sdk.NewAttribute(BallotIdentifier, ballot.BallotIdentifier),
+			sdk.NewAttribute(CCTXIndex, ballot.BallotIdentifier),
+			sdk.NewAttribute(BallotObservationHash, observationHash),
+			sdk.NewAttribute(BallotObservationChain, obserVationChain),
+			sdk.NewAttribute(BallotType, ballot.ObservationType.String()),
+		),
+	)
+}
```
kingpinXD, line 26: We already have a typed event for Ballot created .

x/observer/types/message_add_blame_vote.go
```diff
@@ -0,0 +1,60 @@
...
+	return TypeMsgAddBlameVote
+}
+
+func (m *MsgAddBlameVote) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	if m.ChainId < 0 {
```
kingpinXD, line 35:
```suggestion
		if common.GetChainFromChainID(m.ChainId) == nil{
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d) does not exist", m.ChainId)
	}
```

x/observer/types/message_add_blame_vote.go
```diff
@@ -0,0 +1,60 @@
...
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d)", m.ChainId)
+	}
+	return nil
```
kingpinXD, line 38:
```suggestion
        if common.GetChainFromChainID(m.ChainId) == nil{
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d) does not exist", m.ChainId)
	}
	return nil
```

x/observer/types/message_add_blame_vote.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (m *MsgAddBlameVote) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	if m.ChainId < 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d)", m.ChainId)
+	}
```
kingpinXD, line 37:
```suggestion
	if common.GetChainFromChainID(m.ChainId) == nil{
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d) does not exist", m.ChainId)
	}
```

x/observer/types/events.go
kingpinXD: https://github.com/zeta-chain/zeta-node/blob/452206ce6fa67e4364cac519bdbcd5d2e7833589/x/observer/keeper/events.go#L11-L21

kevinssgh: Yes, this is the same one that I moved to observer

kevinssgh: I see, it was moved to the keeper instead. I'll remove it

proto/observer/blame.proto
kevinssgh: The node account is defined in crosschain still. We can make this change in the next release, I'll take a note.

x/observer/keeper/keeper_utils.go
kevinssgh: We can make that change for sure but this is the first iteration of the blame implementation and I don't want to introduce this right now, it might take more time to test.

kingpinXD (approved):

PR #837, opened 2023-07-22
fix: remove SetCrossChainTx() at the first ballot creation time

brewmaster012:
This is a bug because SetCrossChainTx() not only sets Cctx, but also NonceToCctx and InTxHashToCctx two mappers. When in a revert triggered by a observation of failed outbound tx, setting NonceToCctx will have wrong chain because the Cctx has not transitioned its state to PendingRevert yet.

Also includes slight refactor that inlines two functions that are called only once, and their names are not descriptive and functionality heavily relies on side effects.

### DIFF
--- a/x/crosschain/genesis.go
+++ b/x/crosschain/genesis.go
@@ -46,7 +46,7 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) 
 
 	// Set all the send
 	for _, elem := range genState.CrossChainTxs {
-		k.SetCrossChainTx(ctx, *elem)
+		k.SetCctxAndNonceToCctxAndInTxHashToCctx(ctx, *elem)
 	}
 
 	// Set all the nodeAccount
--- a/x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go
+++ b/x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go
@@ -78,7 +78,7 @@ func (k msgServer) VoteOnObservedOutboundTx(goCtx context.Context, msg *types.Ms
 		// Set this the first time when the ballot is created
 		// The ballot might change if there are more votes in a different outbound ballot for this cctx hash
 		cctx.GetCurrentOutTxParam().OutboundTxBallotIndex = ballotIndex
-		k.SetCrossChainTx(ctx, cctx)
+		//k.SetCctxAndNonceToCctxAndInTxHashToCctx(ctx, cctx)
 	}
 	// AddVoteToBallot adds a vote and sets the ballot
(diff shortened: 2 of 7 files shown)


x/crosschain/genesis.go
```diff
@@ -46,7 +46,7 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 
 	// Set all the send
 	for _, elem := range genState.CrossChainTxs {
-		k.SetCrossChainTx(ctx, *elem)
+		k.SetCctxAndNonceToCctxAndInTxHashToCctx(ctx, *elem)
```
kingpinXD, line 49: Why are we removing the SetCCTX here? 
Also init genesis is only called once when the module is added , it is not called for at upgrades

x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go
```diff
@@ -78,7 +78,7 @@ func (k msgServer) VoteOnObservedOutboundTx(goCtx context.Context, msg *types.Ms
 		// Set this the first time when the ballot is created
 		// The ballot might change if there are more votes in a different outbound ballot for this cctx hash
 		cctx.GetCurrentOutTxParam().OutboundTxBallotIndex = ballotIndex
-		k.SetCrossChainTx(ctx, cctx)
+		//k.SetCctxAndNonceToCctxAndInTxHashToCctx(ctx, cctx)
```
kingpinXD, line 81: It might be better to split this function  `SetCctxAndNonceToCctxAndInTxHashToCctx ` into three. 
We can always set the CCTX here and keep the other two mappings the same.

SetCctxAndNonceToCctxAndInTxHashToCctx can be a helper function which calls the three, but not use it here

brewmaster012: I don't think splitting SetCctxAndNonceToCctxAndInTxHashToCctx() into three is good idea:
1. the three "sub-functions" are always grouped and called together in 6 call sites, except this one; 
2. The function is already very short; 
3. If you want to set CCTX just set CCTX. There is no need for creating additional functions.

kingpinXD: We don't have a set SetCCTX function right now, 
`SetCctxAndNonceToCctxAndInTxHashToCctx` can exist and call the three sub-functions.
The function is short but still has logic, which is different from what a setter function should do. 

Having three subfunctions will also make it easier to write unit tests.

But I understand that we are short on time and this needs to be merged. Approving for now as functionality wise the changes are minor

kingpinXD (approved):

Issue #851, opened 2023-07-26
Add `goimports` in `golangci.yml` and format imports in the codebase


lumtis:
I think formatted imports with a convention like
```
import (
  [standard packages]

  [third-party packages]

  [zetachain packages]
)
```
is beneficial for the readability of the codebase.

I suggest in this issue adding the `goimports` linter in `golangci.yml` and formatting all imports in the codebase following the above convention.

PR #859, opened 2023-07-27
fix: hotfix v6.2.0

kingpinXD:
# Description

- Hotfix v6.2.0 adds back the `evidence` , `slashing`, and `distribution` modules 
- It adds a check to reject `MsgUndelegate`

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1,6 +1,12 @@
 package app
 
 import (
+	distr "github.com/cosmos/cosmos-sdk/x/distribution"
+	"github.com/cosmos/cosmos-sdk/x/evidence"
+	evidencekeeper "github.com/cosmos/cosmos-sdk/x/evidence/keeper"
+	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
+	"github.com/cosmos/cosmos-sdk/x/slashing"
+	slashingkeeper "github.com/cosmos/cosmos-sdk/x/slashing/keeper"
 	slashingtypes "github.com/cosmos/cosmos-sdk/x/slashing/types"
 	"time"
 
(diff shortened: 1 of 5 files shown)


x/crosschain/types/expected_keepers.go
```diff
@@ -11,6 +11,8 @@ import (
 type StakingKeeper interface {
 	GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Validator)
 	GetValidator(ctx sdk.Context, addr sdk.ValAddress) (validator stakingtypes.Validator, found bool)
+	SetParams(ctx sdk.Context, params stakingtypes.Params)
+	GetParams(ctx sdk.Context) (params stakingtypes.Params)
```
lumtis, line 15: I would add a TODO to remember to remove those once we revert changes

x/crosschain/module.go
```diff
@@ -175,8 +176,10 @@ func (AppModule) ConsensusVersion() uint64 { return 1 }
 
 // BeginBlock executes all ABCI BeginBlock logic respective to the capability module.
 func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
-	if ctx.BlockHeader().Height == 946264 {
-		panic("pause at 946264; please revert back to v6.0.0 and restart")
+	if ctx.BlockHeight() == 946265 {
+		params := am.stakingKeeper.GetParams(ctx)
+		params.UnbondingTime = time.Hour * 24 * 4
```
lumtis, line 181: Something like an hour would also be safe if we want to keep the value low

app/app.go
```diff
@@ -1,6 +1,12 @@
 package app
 
 import (
+	distr "github.com/cosmos/cosmos-sdk/x/distribution"
+	"github.com/cosmos/cosmos-sdk/x/evidence"
+	evidencekeeper "github.com/cosmos/cosmos-sdk/x/evidence/keeper"
+	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
+	"github.com/cosmos/cosmos-sdk/x/slashing"
+	slashingkeeper "github.com/cosmos/cosmos-sdk/x/slashing/keeper"
```
lumtis, line 9: I think it would be better to group the imports below where they got initially commented out to be more visual.

I suggest formatting all the imports in the codebase: https://github.com/zeta-chain/node/issues/851 so maybe all of this could be handled there.

x/crosschain/types/expected_keepers.go
kingpinXD: I dont think we are planning to merge this PR , we can go to v7.0.0 directly

x/crosschain/module.go
kingpinXD: THe original plan was to have this as 4 days , I think the number we ended up with due to some sort of typo during the gov proposal

app/app.go
kingpinXD: Makes sense , I like the idea of grouping , Might not be worth the time on this PR since we are not merging , but we do this in v7

Issue #878, opened 2023-07-31
Increase `fungible` keeper test coverage


lumtis:
Part of https://github.com/zeta-chain/node/issues/643

`fungible` module manipulate tokens and therefore we should consider a high value for the code coverage of the module keeper.

PR #961, opened 2023-08-14
test(`fungible/keeper`): increase fungible keeper code coverage - `CallEVMWithData `

lumtis:
# Description

Part of https://github.com/zeta-chain/node/issues/878

Increase code coverage of the fungible keeper module:
- Implement initializer for SDK module's and EVM's test keepers
- Add option for `fungible` test keeper to mock dependent modules
- Add test for `CallEVMWithData`

This is a single test added to eventually see if there is some feedback.
This is an example of test using mocking. We're testing `CallEVMWithData` to see if the method successfully call `ApplyMessage` as part of its specs.
We add test case to check if the gas is estimated depending on if `gasLimit` is provided or not.
We finally test method failure returns an error.

Other misc changes:
- Put `systemcontract.go` content into `system_contract.go`

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


brewmaster012:
This is very cool stuff! I really like it.
If we can extend it to run contracts (no need to involve other modules yet, just as a mock EVM network), it would be fantastic.

brewmaster012 (approved):

lumtis:
> This is very cool stuff! I really like it. If we can extend it to run contracts (no need to involve other modules yet, just as a mock EVM network), it would be fantastic.

With `FungibleKeeperWithMocks` we can instantiate a fungible keeper and choose modules to be mocked.
I had a look into the message methods and for example for `DeployFungibleCoinZRC20` I think it would be best to use unmocked EVM keeper to test if the message execution deploys the contract in the EVM module.

`CallEVMWithData` is a low level method and therefore I think it's best to use all mocks to verify the method adheres to a specification (here the method run the apply function of `evm`).

In general I think the best practice is to use:
- mocked modules for low level methods to test method specification
- unmocked modules for high level methods like messages to test state changes in general
- 
Using mocks for high level methods end up in over-specified tests which increases the overhead of code maintainance

kingpinXD (approved):

PR #981, opened 2023-08-15
feat: add-observer-tx

kingpinXD:
# Description
- Add handler logic for add-observer-tx 
- Add a new policy type, `add_observer.`
- Add new permission , `IsOutBoundEnabled`

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/observer/keeper/observer_mapper.go
```diff
@@ -85,9 +86,36 @@ func (k Keeper) GetAllObserverMappersForAddress(ctx sdk.Context, address string)
...
+	totalObserverCountCurrentBlock := uint64(0)
+	for _, mapper := range observerMappers {
+		mapper.ObserverList = append(mapper.ObserverList, msg.ObserverAddress)
+		totalObserverCountCurrentBlock += uint64(len(mapper.ObserverList))
+		k.SetObserverMapper(ctx, mapper)
+	}
+	if totalObserverCountCurrentBlock < 0 {
+		return &types.MsgAddObserverResponse{}, types.ErrObserverCountNegative
+	}
+	pubkey, _ := common.NewPubKey(msg.ZetaclientGranteePubkey)
```
lumtis, line 104: I think we should still handle the error even if it is handled in message validation

x/observer/types/message_add_observer.go
```diff
@@ -43,5 +46,18 @@ func (msg *MsgAddObserver) ValidateBasic() error {
...
+	}
+	_, err = sdk.AccAddressFromBech32(msg.ZetaclientGranteeAddress)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid zetaclient grantee address (%s)", err)
+	}
+	_, err = common.NewPubKey(msg.ZetaclientGranteePubkey)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidPubKey, "invalid zetaclient grantee pubkey (%s)", err)
+	}
+	// https://github.com/zeta-chain/node/issues/988
```
lumtis, line 61: Not sure I understand, if it should be from the same pair, why not derivating the address from the pubkey?

x/observer/types/params.go
```diff
@@ -52,9 +56,11 @@ func DefaultParams() Params {
 			PolicyType: Policy_Type_update_keygen_block,
 			Address:    "zeta1afk9zr2hn2jsac63h4hm60vl9z3e5u69gndzf7c99cqge3vzwjzsxn0x73",
 		},
+		{
+			PolicyType: Policy_Type_add_observer,
+			Address:    "zeta1afk9zr2hn2jsac63h4hm60vl9z3e5u69gndzf7c99cqge3vzwjzsxn0x73",
```
lumtis, line 61: We're reusing the same hardcoded address many places, I think we should define a constant

zetaclient/bitcoin_client.go
```diff
@@ -210,7 +210,7 @@ func (ob *BitcoinChainClient) WatchInTx() {
 
 // TODO
 func (ob *BitcoinChainClient) observeInTx() error {
-	permssions, err := ob.zetaClient.GetInboundPermissions()
+	permssions, err := ob.zetaClient.GetPermissionFlags()
```
lumtis, line 213:
```suggestion
	permissions, err := ob.zetaClient.GetPermissionFlags()
```

zetaclient/evm_client.go
```diff
@@ -537,7 +537,7 @@ func (ob *EVMChainClient) ExternalChainWatcher() {
 }
 
 func (ob *EVMChainClient) observeInTX() error {
-	permssions, err := ob.zetaClient.GetInboundPermissions()
+	permssions, err := ob.zetaClient.GetPermissionFlags()
```
lumtis, line 540:
```suggestion
	permissions, err := ob.zetaClient.GetPermissionFlags()
```

lumtis (changes requested):

proto/observer/tx.proto
```diff
@@ -27,8 +27,9 @@ message MsgUpdateCoreParamsResponse {}
 
 message MsgAddObserver {
   string creator = 1;
-  int64 chain_id = 2;
-  ObservationType observationType = 3;
+  string observer_address = 3;
+  string zetaclient_grantee_address = 4;
+  string zetaclient_grantee_pubkey = 5;
```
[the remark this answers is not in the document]
kingpinXD: I would like to keep the prefix to signify that these are used by zeta clients only

app/setup_handlers.go
```diff
@@ -5,19 +5,21 @@ import (
...
 	app.UpgradeKeeper.SetUpgradeHandler(releaseVersion, func(ctx sdk.Context, plan types.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		app.Logger().Info("Running upgrade handler for " + releaseVersion)
 
-		// Updated version map to thelatest consensus versions from each module
+		// Updated version map to the latest consensus versions from each module
 		for m, mb := range app.mm.Modules {
 			vm[m] = mb.ConsensusVersion()
 		}
-
+		vm[observerTypes.ModuleName] = vm[observerTypes.ModuleName] - 1
```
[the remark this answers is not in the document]
kingpinXD: Since I am assigning the version map manually, I would need to decrement so that the from version is `1` and toVersion, which comes from `module.ConsensusVersion`, is 2.

x/observer/keeper/observer_mapper.go
kingpinXD: Can you think of a way in which validate basic can be skipped by a TX? I know it s a minute operation , but if not needed we can skip it , what are your thoughts ?

x/observer/types/message_add_observer.go
kingpinXD: answered above

kingpinXD: also updated the issue mentioned here to reflect the same

proto/observer/tx.proto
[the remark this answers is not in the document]
lumtis: Ok, not blocking I just think it's better to use the simplest name when possible

x/observer/keeper/observer_mapper.go
lumtis: We have some message handlers in the code that can be called without sending a message, wondering if it could be the case for this one. But actually even outside of this, we are supposed to consider the validity of the message a precondition for the message handler execution, so it's fine from my side.

proto/observer/events.proto
```diff
@@ -19,3 +19,11 @@ message EventKeygenBlockUpdated {
   string keygen_block = 2;
   string keygen_pubkeys = 3;
 }
+
+message EventNewObserverAdded {
+  string msg_type_url = 1;
+  string observer_address = 2;
+  string zetaclient_grantee_address = 3;
```
lumtis, line 26: I may see that `GetAddressFromPubkeyString` has been added in your new commit.

If we have this method, do we still need to keep the address here then?

x/observer/types/message_add_observer.go
kingpinXD: https://github.com/zeta-chain/zeta-node/blob/e8f8182d18d0ba4d1a64285d521e21c997994d51/common/pubkey.go#L36-L46

x/observer/types/message_add_observer_test.go
```diff
@@ -22,9 +22,7 @@ func TestMsgAddObserver_ValidateBasic(t *testing.T) {
 			err: sdkerrors.ErrInvalidAddress,
 		}, {
 			name: "valid address",
-			msg: MsgAddObserver{
-				Creator: sample.AccAddress(),
-			},
+			msg:  GenerateObserverMsg(),
 		},
```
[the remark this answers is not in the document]
kingpinXD: https://github.com/zeta-chain/zeta-node/blob/e8f8182d18d0ba4d1a64285d521e21c997994d51/x/observer/types/message_add_observer_test.go#L11-L57

proto/observer/events.proto
kingpinXD: I think we can keep it for the event . It might be easier for some consumers to filter using their addresses directly. 
don't have strong opinions about it though, that do you think?

lumtis: Sorry, haven't seen this is the event, this makes sense.

app/setup_handlers.go
```diff
@@ -5,19 +5,21 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	observerTypes "github.com/zeta-chain/zetacore/x/observer/types"
```
lumtis, line 8:
```suggestion
	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
```

x/observer/client/cli/tx_permission_flags.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func CmdUpdatePermissionFlags() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "update-permission-flags [is-inbound-enabled]",
+		Use:   "update-permission-flags [is-inbound-enabled] [is-outbound-enabled]",
 		Short: "Update PermissionFlags",
 		Args:  cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
lumtis: We have two args, I believe we should apply this

lumtis (commented):
Looks good to me, just few small comments

lumtis (approved):

app/setup_handlers.go
```diff
@@ -5,19 +5,21 @@ import (
...
 	app.UpgradeKeeper.SetUpgradeHandler(releaseVersion, func(ctx sdk.Context, plan types.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		app.Logger().Info("Running upgrade handler for " + releaseVersion)
 
-		// Updated version map to thelatest consensus versions from each module
+		// Updated version map to the latest consensus versions from each module
 		for m, mb := range app.mm.Modules {
 			vm[m] = mb.ConsensusVersion()
 		}
-
+		vm[observertypes.ModuleName] = vm[observertypes.ModuleName] - 1
```
brewmaster012, line 21: @kingpinXD you wanna keep these in v9?

proto/observer/tx.proto
```diff
@@ -27,8 +27,8 @@ message MsgUpdateCoreParamsResponse {}
 
 message MsgAddObserver {
   string creator = 1;
-  int64 chain_id = 2;
-  ObservationType observationType = 3;
+  string observer_address = 2;
+  string zetaclient_grantee_pubkey = 3;
```
brewmaster012, line 31: is observer_address the val operator address? 
is the grantee_pubkey the hotkey pubkey?

proto/observer/tx.proto
```diff
@@ -42,7 +42,8 @@ message MsgAddBlameVote {
 message MsgAddBlameVoteResponse {}
 message MsgUpdatePermissionFlags {
   string creator = 1;
-  bool isInboundEnabled = 3;
+  bool isInboundEnabled = 2;
```
brewmaster012, line 45: i would refrain from changing the numbering.

x/crosschain/types/expected_keepers.go
```diff
@@ -56,6 +56,7 @@ type ZetaObserverKeeper interface {
 	GetAllNodeAccount(ctx sdk.Context) (nodeAccounts []zetaObserverTypes.NodeAccount)
 	SetNodeAccount(ctx sdk.Context, nodeAccount zetaObserverTypes.NodeAccount)
 	IsInboundAllowed(ctx sdk.Context) (found bool)
```
brewmaster012, line 58: use the same name in proto, i.e. isInboundEnabled?

x/observer/types/keys.go
```diff
@@ -17,6 +17,8 @@ const (
 
 	// MemStoreKey defines the in-memory store key
 	MemStoreKey = "mem_observer"
+
+	GroupId1Address = "zeta1afk9zr2hn2jsac63h4hm60vl9z3e5u69gndzf7c99cqge3vzwjzsxn0x73"
```
brewmaster012, line 21: what is this address?

brewmaster012 (approved):
left some minor comments

x/observer/keeper/observer_mapper.go
brewmaster012: i agree with @lumtis , it takes **much less time** to add the error check here than all three of us trying to **think** about it. Maybe @kingpinXD you are right, but to prevent future waste of time, attention, and energy, every error in zetacored needs to be checked.

app/setup_handlers.go
kingpinXD: we have a migration for observer store 
https://github.com/zeta-chain/zeta-node/blob/31d033a3cec3872249056d7215c4428bc8622d64/x/observer/migrations/v2/migrate.go#L11-L28

proto/observer/tx.proto
kingpinXD: yes

x/observer/types/keys.go
kingpinXD: A constant used for the first group ID 1, we were using it in multiple places, just replacing it with a constant

PR #1003, opened 2023-08-20
refactor: update InitGenesis and ExportGenesis

lumtis:
# Description

Update the export and import genesis

When merged we can make a cherry pick to make the change on `develop`

I'm creating the branch now for availability. I think we could also update Validate methods and tests


x/emissions/genesis.go
```diff
@@ -10,15 +10,17 @@ import (
 // state.
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, ak types.AccountKeeper, genState types.GenesisState) {
 	k.SetParams(ctx, genState.Params)
+
+	// This create module accounts if not created
+	// TODO: find a proper way  to create module accounts?
```
kingpinXD, line 15: We can add methods for withdrawable emissions . That would need a change in proto files though

kingpinXD (approved):

x/emissions/genesis.go
lumtis: Yes, does it have a logical on-chain usage right now? It seems right now it is only used for documentation purposes.
But it should be straightforward to add.

Maybe we could try to maintain somewhere a TODO list for when adding new state components so we always add the value in the genesis in the same PR we add the type.

lumtis:
Let me know if we can confirm we don't need this to the hotfix branch, I will port it to develop instead

lumtis:
Recreated the branch for `develop` since no longer needed for the hotfix https://github.com/zeta-chain/node/pull/1023

PR #1012, opened 2023-08-21
feat: Telemetry improvements

kevinssgh:
# Description

1. Added an interface to the existing telemetry to get the Public IP of the TSS P2P node that is configured by the env Var: 
MYIP
2. Also added a check to ensure the env var is set

Closes: https://github.com/zeta-chain/node/issues/972

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet


lumtis (approved):
LGTM

CharlieMc0:
Can public IP still be set in the zetaclient_config.json file or does this force operators to use the MYIP var?

cmd/zetaclientd/start_utils.go
```diff
@@ -13,16 +13,13 @@ import (
...
-		err := os.Setenv("MYIP", cfg.PublicIP)
-		if err != nil {
-			logger.Fatal().Err(err).Msg("Error setting MYIP")
-		}
+func getMYIP(logger zerolog.Logger) (string, error) {
+	ip := os.Getenv("MYIP")
+	if ip == "" {
+		logger.Fatal().Msg("Please set MYIP environment variable")
+		return "", errors.New("empty ip address")
 	}
```
kingpinXD, line 21: why remove setting MYIP via config ?

kevinssgh: I think we need to choose one, either the env variable or the config file. It is confusing when we tried to debug whether the public IP was set correctly in each node.

kevinssgh:
> Can public IP still be set in the zetaclient_config.json file or does this force operators to use the MYIP var?

In this case we are not using the config file for public IP but I think it got confusing whether a node was setup correctly since we use both config and env variable. Just need to choose one, I'm fine with either.

cmd/zetaclientd/start.go
```diff
@@ -55,7 +55,9 @@ func start(_ *cobra.Command, _ []string) error {
 
 	masterLogger := log.Logger
 	startLogger := masterLogger.With().Str("module", "startup").Logger()
-	setMYIP(cfg, startLogger)
+	if err != nil {
```
ws4charlie, line 58: We already checked 'err' and returned either at line 44 or or 52?

Issue #1026, opened 2023-08-22
`zetacore` modify zetacore to keep track of previous tss 


kingpinXD:
- Add new data structure of type, to maintain historical TSS entries
- We can still keep the existing TSS data structure for fetching the current TSS .
- The historical TSS entires can be used during migration

Issue #1027, opened 2023-08-22
`zetacore` `zetaclient` Add tss pubkey to cctx 


kingpinXD:
- TSS pubkey can be added to CCTX 
- Zetaclient can ue this this pubkey to fetch the TSS keyshare to use for the outbound .
- Zetaclient needs to be modified to maintain a list of TSS keyshares instead of just the current .

PR #1031, opened 2023-08-23
feat: tss-history

kingpinXD:
# Description
- Add `add Tss History to crosschain store`
- Add `tss pubkey to createCCTX`
- Add a new admin function to update TSS manually instead of doing it automatically when a new TSS is generated
Closes: #1026 #1027

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


cmd/zetaclientd/keygen_tss.go
```diff
@@ -16,9 +17,9 @@ import (
 	"gitlab.com/thorchain/tss/go-tss/p2p"
 )
 
-func GenerateTss(logger zerolog.Logger, cfg *config.Config, zetaBridge *mc.ZetaCoreBridge, peers p2p.AddrList, priKey secp256k1.PrivKey, ts *mc.TelemetryServer) (*mc.TSS, error) {
+func GenerateTss(logger zerolog.Logger, cfg *config.Config, zetaBridge *mc.ZetaCoreBridge, peers p2p.AddrList, priKey secp256k1.PrivKey, ts *mc.TelemetryServer, tssList []*types.TSS) (*mc.TSS, error) {
```
lumtis, line 20:
```
tssList []*types.TSS
```
Is it the TSS history? I would rename it `tssHistoricalList` to be more explicit

lumtis, line 20: Do we need using array of pointer? It seems to me that we never use `nil` value?

cmd/zetaclientd/start.go
```diff
@@ -117,7 +117,12 @@ func start(_ *cobra.Command, _ []string) error {
 		}
 	}()
 
-	tss, err := GenerateTss(masterLogger, cfg, zetaBridge, peers, priKey, telemetryServer)
+	tssHistoricalList, err := zetaBridge.GetTssHistory()
+	if err != nil {
+		startLogger.Error().Err(err).Msg("GetTssHistory error")
+	}
+
+	tss, err := GenerateTss(masterLogger, cfg, zetaBridge, peers, priKey, telemetryServer, tssHistoricalList)
```
lumtis, line 125: Maybe we should explicitly set an empty array for `tssHistoricalList` if the previous method returned an error

common/pubkey.go
```diff
@@ -31,6 +32,18 @@ var EmptyPubKey PubKey
 
 // NewPubKey create a new instance of PubKey
 // key is bech32 encoded string
```
lumtis, line 34: Misplaced comments

common/pubkey.go
```diff
@@ -31,6 +32,18 @@ var EmptyPubKey PubKey
...
+
+func GetAddressFromPubkeyString(pubkey string) (sdk.AccAddress, error) {
```
lumtis, line 36: Maybe we could add unit test for this method

proto/crosschain/genesis.proto
```diff
@@ -24,4 +24,5 @@ message GenesisState {
 
   repeated LastBlockHeight lastBlockHeightList = 8;
   repeated InTxHashToCctx inTxHashToCctxList = 9 [(gogoproto.nullable) = false];
+  repeated TSS tss_history = 10;
```
lumtis, line 27: Any reason to not use `[(gogoproto.nullable) = false]`? I feel like using pointers increase complexity and the risk of getting nil dereference

x/crosschain/client/cli/cli_tss.go
```diff
@@ -41,6 +41,32 @@ func CmdShowTSS() *cobra.Command {
 	return cmd
 }
 
+func CmdListTssHistory() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list-tss-history",
```
lumtis, line 46: Maybe we could rename the query cmd into `TssHistory` instead of `ListTssHistory`

x/crosschain/genesis.go
```diff
@@ -42,6 +42,10 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 		k.SetTSS(ctx, *genState.Tss)
 	}
 
+	for _, elem := range genState.TssHistory {
+		k.SetTSSHistory(ctx, *elem)
```
lumtis, line 46:
```suggestion

	for _, elem := range genState.TssHistory {
		if elem != nil {
		    k.SetTSSHistory(ctx, *elem)
		}
```

x/crosschain/genesis.go
```diff
@@ -84,6 +88,11 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 		genesis.CrossChainTxs = append(genesis.CrossChainTxs, &e)
 	}
 
+	tssList := k.GetAllTSS(ctx)
+	for _, elem := range tssList {
+		e := elem
+		genesis.TssHistory = append(genesis.TssHistory, e)
+	}
```
lumtis, line 95: Can't we do `genesis.TssHistory = k.GetAllTSS(ctx)` directly?

x/crosschain/keeper/keeper_cross_chain_tx_vote_inbound_tx.go
```diff
@@ -70,6 +70,10 @@ func (k msgServer) VoteOnObservedInboundTx(goCtx context.Context, msg *types.Msg
 	if receiverChain == nil {
 		return nil, sdkerrors.Wrap(types.ErrUnsupportedChain, fmt.Sprintf("ChainID %d, Observation %s", msg.ReceiverChain, observationType.String()))
 	}
+	tss, found := k.GetTSS(ctx)
+	if !found {
+		ctx.Logger().Error("TSS not found")
```
lumtis, line 75: Shouldn't we return an error? We use `tss.TssPubkey` later which I believe would cause a panic

x/crosschain/keeper/keeper_tss.go
```diff
@@ -36,22 +45,32 @@ func (k Keeper) RemoveTSS(ctx sdk.Context) {
...
+		list = append(list, &val)
 	}
-
 	return
 }
 
+// GetPreviousTSS returns the previous tss information
+func (k Keeper) GetPreviousTSS(ctx sdk.Context) (val types.TSS, found bool) {
+	tssList := k.GetAllTSS(ctx)
+	if len(tssList) <= 2 {
```
lumtis, line 64:
```suggestion
	if len(tssList) < 2 {
```
I believe you can have a previous tss if len==2

x/crosschain/keeper/keeper_tss_voter.go
```diff
@@ -89,13 +89,15 @@ func (k msgServer) CreateTSSVoter(goCtx context.Context, msg *types.MsgCreateTSS
...
+		tss := types.TSS{
 			TssPubkey:           msg.TssPubkey,
 			TssParticipantList:  keygen.GetGranteePubkeys(),
 			OperatorAddressList: ballot.VoterList,
 			FinalizedZetaHeight: ctx.BlockHeight(),
 			KeyGenZetaHeight:    msg.KeyGenZetaHeight,
-		})
+		}
+		k.SetTSS(ctx, tss)
+		k.SetTSSHistory(ctx, tss)
```
lumtis, line 100: Maybe we could implement a method `AppendTSS` that would make the atomic operation of adding a new TSS.

x/crosschain/keeper/msg_server_whitelist_erc20.go
```diff
@@ -35,6 +36,10 @@ func (k Keeper) WhitelistERC20(goCtx context.Context, msg *types.MsgWhitelistERC
 			return nil, sdkerrors.Wrapf(types.ErrInvalidAddress, "ERC20 contract address (%s) already whitelisted on chain (%d)", msg.Erc20Address, msg.ChainId)
 		}
 	}
+	tss, found := k.GetTSS(ctx)
+	if !found {
+		return nil, errorsmod.Wrapf(types.ErrCannotFindTSSKeys, "Cannot create new admin cmd")
```
lumtis, line 41: > `Cannot create new admin cmd`

Is it the right error message? It seems not related to the error to me

zetaclient/tss_signer.go
```diff
@@ -334,23 +339,50 @@ func getKeyAddrBTCWitnessPubkeyHash(tssPubkey string) (*btcutil.AddressWitnessPu
...
+			return fmt.Errorf("pubkey %s not found in keyshare", t.TssPubkey)
+		}
+	}
+	return nil
+}
+
+func (tss *TSS) LoadTssFilesFromDirectory(tssPath string) error {
+	files, err := os.ReadDir(tssPath)
+	if err != nil {
+		fmt.Println("ReadDir error", err)
```
lumtis, line 383:
```suggestion
		fmt.Println("ReadDir error:", err.Error())
```

cmd/zetaclientd/keygen_tss.go
kingpinXD: We don't need it , but I don't see any issues with keeping it. Do you see any issues with the pointer ?

common/pubkey.go
kingpinXD: This method is not needed, I removed it.

x/crosschain/client/cli/cli_tss.go
kingpinXD: We have been using `show-XXX` and `list-XXX` for all items to show query single and multiple . 
This goes with that nomenclature

x/crosschain/keeper/keeper_cross_chain_tx_vote_inbound_tx.go
kingpinXD: We shouldn't return here, as deposit Tx does not need a TSS address. I added some conditional logic instead. Thanks for pointing this out

x/crosschain/keeper/msg_server_whitelist_erc20.go
kingpinXD: whitelistERC20 is an action of type admin cmd, we will have more of these in the future .
Added some details to the error print so that it makes more sense.

zetaclient/tss_signer.go
```diff
@@ -334,23 +339,50 @@ func getKeyAddrBTCWitnessPubkeyHash(tssPubkey string) (*btcutil.AddressWitnessPu
...
+	}
+	err = tss.VerifyKeysharesForPubkeys(tssHistoricalList)
 	if err != nil {
-		fmt.Println("ReadDir error", err)
 		return nil, err
 	}
+	return &tss, nil
+}
+
+func (tss *TSS) VerifyKeysharesForPubkeys(tssList []types.TSS) error {
```
brewmaster012, line 366: i don't think this is correct, new keysigners would not have old keyshares that they were not part of.

brewmaster012 (changes requested):

zetaclient/tss_signer.go
kingpinXD: Thanks that's a great catch , 
Modified to using only TSS that the node participated in

lumtis:
Is it ready for re-review @kingpinXD ?

brewmaster012 (approved):

x/observer/keeper/observer_mapper.go
```diff
@@ -93,24 +93,32 @@ func (k msgServer) AddObserver(goCtx context.Context, msg *types.MsgAddObserver)
 	if msg.Creator != k.GetParams(ctx).GetAdminPolicyAccount(types.Policy_Type_add_observer) {
 		return &types.MsgAddObserverResponse{}, types.ErrNotAuthorizedPolicy
 	}
+	pubkey, _ := common.NewPubKey(msg.ZetaclientGranteePubkey)
+	granteeAddress, _ := common.GetAddressFromPubkeyString(msg.ZetaclientGranteePubkey)
+	k.DisableInboundOnly(ctx)
+	// AddNodeAccountOnly flag usage
+	// True: adds observer into the Node Account list but returns without adding to the observer list
+	// False: adds observer to the observer list, and not the node account list
```
lumtis, line 101: The `only` naming is a bit confusing to me because it would imply it's an action that is always executed but only this action is executed if the flag is set.

zetaclient/bitcoin_client.go
```diff
@@ -183,6 +183,28 @@ func (ob *BitcoinChainClient) GetLastBlockHeight() int64 {
 	return height
 }
 
+func (ob *BitcoinChainClient) SetLastBlockHeightScanned(block int64) {
```
lumtis, line 186: I think we should add comment to explain difference with `GetLastBlockHeight`

zetaclient/chainclient.go
```diff
@@ -10,8 +10,8 @@ import (
 // general chain client
 
 type ChainClient interface {
-	GetLastBlockHeight() int64 // 0 means error
-	SetLastBlockHeight(int64)
+	//GetLastBlockHeight() int64 // 0 means error
+	//SetLastBlockHeight(int64)
```
lumtis, line 14: I think we should remove methods in interfaces no longer used instead of commenting them out if we plan to no longer use them. It might be confusing when reading the code.

zetaclient/tss_signer.go
```diff
@@ -359,23 +364,60 @@ func getKeyAddrBTCWitnessPubkeyHash(tssPubkey string) (*btcutil.AddressWitnessPu
...
+		if NodeWasPartOfTss(granteePubKey32, t.TssParticipantList) {
+			if _, ok := tss.Keys[t.TssPubkey]; !ok {
+				return fmt.Errorf("pubkey %s not found in keyshare", t.TssPubkey)
+			}
+		}
+	}
+	return nil
+}
+
+func NodeWasPartOfTss(granteePubKey32 string, granteeList []string) bool {
```
lumtis, line 407:
```suggestion
func WasNodePartOfTss(granteePubKey32 string, granteeList []string) bool {
```

x/observer/client/cli/tx_add_observer.go
```diff
@@ -10,7 +13,7 @@ import (
 
 func CmdAddObserver() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "add-observer [observer-address] [zetaclient-grantee-pubkey]",
+		Use:   "add-observer [observer-address] [zetaclient-grantee-pubkey] [add_node_account_only]",
```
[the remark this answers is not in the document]
kingpinXD: I agree, but this flag is temporary only, the only reason we are doing this in two steps as because we don't have the post-generation tasks ready, and we need to push the upgrade out soon. This flag will probably be removed in the next release.

kevinssgh (approved):
Looks good

Issue #1021, opened 2023-08-22
Trackers persist while outTxs have already been paid.


ws4charlie:

ws4charlie:
Take Mumbi testnet for an example. The trackers with smaller nonces (`2527, 2596, 2655...`) still exists in tracker store while pending cctx's low nonce already reaches `**18279**`. The small-nonce tracker will never get the chances to be cleaned up.
![image](https://github.com/zeta-chain/node/assets/34498985/7193b6b1-3967-4415-938c-9635bcbfd1f3)

ws4charlie:
pending cctxs and garbage trackers had been cleaned up with hotfix release [v11.0.4](https://github.com/zeta-chain/node/releases/tag/v11.0.4)

PR #1050, opened 2023-08-25
fix: clean up missed trackers whose nonces are below nonceLow

ws4charlie:
# Description

We have some missed out tx trackers never get chances to be confirmed and cleaned from tracker store while outbound transaction went through successfully. This PR is fix the issue by correctly pick up those missed issue in RPC query.

Closes: https://github.com/zeta-chain/node/issues/1021

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


x/crosschain/keeper/keeper_cross_chain_tx.go
```diff
@@ -167,17 +170,17 @@ func (k Keeper) CctxAllPending(c context.Context, req *types.QueryAllCctxPending
...
-	startNonce := p.NonceLow - 100
-	if startNonce < 0 {
-		startNonce = 0
+	outTxTrackers, err := k.GetAllTrackerByChainBeforeNonce(ctx, int64(req.ChainId), p.NonceLow, req.Pagination)
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to query out tx trackers by chain %d before nonce %d", req.ChainId, p.NonceLow))
 	}
-	for i := startNonce; i < p.NonceLow; i++ {
-		res, found := k.GetNonceToCctx(ctx, tss.TssPubkey, int64(req.ChainId), i)
+	for _, tracker := range outTxTrackers {
```
brewmaster012, line 180: one concern is that outTxTrackers is not after consensus; it's probably not a good idea to rely on outTxTrackers here. 

Also, there might be cases that no outTxTracker exists for a "forgotten" pending cctx.

ws4charlie: True. Trackesr are not trustworthy info.

Issue #1046, opened 2023-08-25
Add on-chain algorithm to increase outbound tx gas price


lumtis:
Part of https://github.com/zeta-chain/node/issues/935
Depends on https://github.com/zeta-chain/node/issues/1045

Determine an algorithm to increase the gas price of an outbound tx on-chain.

The ountbound tx gas price should be increased if no status change after a period of time.

The algorithm can be as follows:

```
epochLenght // number of blocks for gas price iteration
increaseGasPriceLength // number of blocks to wait before increasing gas price

beginBlocker:
  if blockHeight % epochLength == 0:
    for outTx = range activeOutBoundTxs():
      if blockHeight - outTx.LastUpdatedHeight > increaseGasPriceLength:
        increaseGasPrice(outTx)
        resendTx(outTx)
```

Issue #1045, opened 2023-08-25
Implement `IncreaseGasPrice` for outbound tx


lumtis:
Part of https://github.com/zeta-chain/node/issues/935

The method `IncreaseGasPrice` allows to to increase the gas price for an outbound tx to be resent. The gas increase should be taken from gas stability pool and burned.

PR #1062, opened 2023-08-30
feat(`crosschain`): increase pending cctx gas prices

lumtis:
# Description

- Add gas stability pool interface into `crosschain`
- Add in `crosschain` a BeginBlock routine that each epoch iterates the pending outbound and update the gas price if not been mined since a `RetryInterval` time
- Fund the gas stability pool with remaining fees of cctxs
- Add fungible mock in `crosschain` for testing gas stability pool funding and withdrawing
- Add `EffectiveGasLimit` in outbound param to effective real remaining gas
- Implement a temporary fix for https://github.com/zeta-chain/node/issues/1085 because otherwise smoketest fail with accounting error (more remaining fees than fees burned because effectiveGasPrice > gasPrice) this doesn't close directly the issue, I think we should return an error if the gas price is not defined on ZetaChain

### Notes

- There are 3 values that will sill require a decision for final values
  - `EpochLength` the number of blocks before doing an iteration through the pending txs (prevent doing an iteration every blocks which is a bit extensive)
  - `RetryInterval`  the time a tx must be pending before increasing the gas price
  - `GasPriceIncreasePercent` the percent of the medium gas price used for the increase. 100% means the gas price is increased by the medium gas price of the chain every step

- `lastUpdate_timestamp` from the status of the cctx is used to detect when the gas price must be increased, this value is updated when we increase the gas price of the outbound tx, it doesn't seem to have side effects to me as it is not used in ZetaClient, but we can implement another value if this one should not be update

Closes: https://github.com/zeta-chain/node/issues/1046 https://github.com/zeta-chain/node/issues/1045

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


brewmaster012 (approved):
LGTM

x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go
```diff
@@ -186,3 +199,35 @@ func (k msgServer) VoteOnObservedOutboundTx(goCtx context.Context, msg *types.Ms
...
+	if gasLimit > 0 && gasUsed > 0 && !gasPrice.IsZero() {
+		if gasLimit >= gasUsed {
+			remainingGas := gasLimit - gasUsed
+			remainingFees := math.NewUint(remainingGas).Mul(gasPrice).BigInt()
+
+			// We fund the stability pool with a portion of the remaining fees
+			remainingFees = percentOf(remainingFees, RemainingFeesToStabilityPoolPercent)
+
+			// Fund the gas stability pool
+			if err := k.fungibleKeeper.FundGasStabilityPool(ctx, chainID, remainingFees); err != nil {
```
brewmaster012, line 225: This FundGasStabilityPool will call zEVM contract, which is supposed to leave Ethereum logs; 
which also makes this cosmos txs a observable Ethereum tx from ETH JSON-RPCs. 

Does this work here?

lumtis: Yes, it calls `CallZRC20Deposit` that calls `CallEVMwithData` that emits events. I can still add a test case if we want to make sure.

kingpinXD (approved):

kingpinXD (approved):

PR #1089, opened 2023-09-05
test: Stateful smoketest

kevinssgh:
# Description

Modified smoketest to run repeatedly with an existing state.

Closes: https://github.com/zeta-chain/node/issues/1039

## Type of change

# How Has This Been Tested?

This has been tested locally using privnet flags. The change allows the smoketest to run at least twice with an existing state. for example:
```
smoketest "$SMOKETEST_CMD"
SMOKETEST_EXIT_CODE=$?
if [ $SMOKETEST_EXIT_CODE -ne 0 ]; then
  echo "smoketest failed"
  exit 1
fi
smoketest "$SMOKETEST_CMD" --deployed
if [ $SMOKETEST_EXIT_CODE -eq 0 ]; then
  echo "smoketest passed"
  exit 0
else
  echo "smoketest failed"
  exit 1
fi
```
- [x] Tested CCTX in localnet


contrib/localnet/orchestrator/smoketest/main.go
```diff
@@ -308,8 +315,8 @@ func LocalSmokeTest(_ *cobra.Command, _ []string) {
 	fmt.Printf("##   Ether on Ethereum private net\n")
 	fmt.Printf("##   ZETA on ZetaChain EVM\n")
 	fmt.Printf("##   ETH ZRC20 on ZetaChain\n")
-	// The following tests are optional tests; comment out the ones you don't want to run
-	// temporarily to reduce dev/test cycle turnaround time
+	//// The following tests are optional tests; comment out the ones you don't want to run
+	//// temporarily to reduce dev/test cycle turnaround time
```
lumtis, line 319: Why adding more slashes?

contrib/localnet/orchestrator/smoketest/main.go
```diff
@@ -110,8 +110,15 @@ var LocalCmd = &cobra.Command{
...
 
+type localArgs struct {
+	contractsDeployed bool
+}
+
+var localTestArgs = localArgs{}
+
 func init() {
 	RootCmd.AddCommand(LocalCmd)
+	LocalCmd.Flags().BoolVar(&localTestArgs.contractsDeployed, "deployed", false, "set to to true if running smoketest again with existing state")
```
lumtis, line 121: We could maybe also have persistence by default and use `reset` flag to redeploy contracts

lumtis (commented):
Just had a first look. Looks good.
I think the prefix should rather be `test` than `fix` as a general improvement for tests.

contrib/localnet/orchestrator/smoketest/main.go
kevinssgh: I think it is a good idea to be able to optionally redeploy but I noticed a lot of the tests failed for reasons that weren't so obvious. I would need some time to investigate and fix those issues. Perhaps in another task we can add this since we will be building more test workflows.

kevinssgh:
> cool are we ready to have a stateful v9 to current develop upgrade in between? how would that look like @kevinssgh 

With these changes we will be able to run any type of scripts between the smoketest runs. I'm not completely familiar with all the steps in an upgrade but it should involve something like this: 

Network will initially run with v9 or which ever version you want to start with. 

A gov proposal for an upgrade height -  A script from the orchestrator container can be run for a proposal, then it will have to ssh into the validators in order to submit a vote. 

Then a binary update will have to happen - orchestrator would have to wait for each node to stop at the given height and replace the binaries. Then restart nodes.

brewmaster012:
> > cool are we ready to have a stateful v9 to current develop upgrade in between? how would that look like @kevinssgh
> 
> With these changes we will be able to run any type of scripts between the smoketest runs. I'm not completely familiar with all the steps in an upgrade but it should involve something like this:
> 
> Network will initially run with v9 or which ever version you want to start with.
> 
> A gov proposal for an upgrade height - A script from the orchestrator container can be run for a proposal, then it will have to ssh into the validators in order to submit a vote.
> 
> Then a binary update will have to happen - orchestrator would have to wait for each node to stop at the given height and replace the binaries. Then restart nodes.

kingpinXD (approved):
LGTM , 
The upgrade script that runs in between, would that be in a separate PR?

lumtis (approved):
LGTM

ws4charlie (approved):

PR #1095, opened 2023-09-05
feat: add new query for count of total pending CCTX

kingpinXD:
# Description

- add a new query for count of PendingCCTX 

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


ws4charlie (approved):
looks good

x/crosschain/keeper/keeper_cross_chain_tx.go
```diff
@@ -157,16 +158,39 @@ func (k Keeper) CctxAllPending(c context.Context, req *types.QueryAllCctxPending
...
+		return nil, status.Error(codes.InvalidArgument, "invalid chain id")
+	}
+	pendingCrossChainTxList, err := k.GetActualPendingCCTX(ctx, int64(req.ChainId))
+	if err != nil {
+		return nil, err
+	}
+	return &types.QueryAllCctxPendingResponse{CrossChainTx: pendingCrossChainTxList}, nil
+}
+
+func (k Keeper) CctxAllPendingCount(c context.Context, _ *types.QueryAllCctxPendingCountRequest) (*types.QueryAllCctxPendingCountResponse, error) {
```
brewmaster012, line 171: this separate RPC call seems a bit redunance to me: 
on one hand, the nonceHigh-nonceLow gives the same result here per chain (which is better than lumped all pendings count together, and much cheaper)

If really need the count, a better place would be to use CctxAllPending and have it with proper pagination (the result will include a total count). This is the same cost as here as all the pending cctxs are collected.

brewmaster012 (changes requested):

x/crosschain/keeper/keeper_cross_chain_tx.go
kingpinXD: The difference between high an low would not include an cctxs which might be have nonces less tha nonce low . As  described on the comment lines.
That is the sole reason for creating this pr , if those ctxs are not needed we can go ahead and close the pr and continue using getpending nonces

As discussed with front end they jsut need the count and not the actual cctx , keeping the same cost this would avoid the need for them to query multiple pages when trying to get just the count .

kingpinXD: The difference between high an low would not include an cctxs which might be have nonces less tha nonce low . As described on the comment lines.
That is the sole reason for creating this pr , if those ctxs are not needed we can go ahead and close the pr and continue using getpending nonces

As discussed with front end they jsut need the count and not the actual cctx , keeping the same cost this would avoid the need for them to query multiple pages when trying to get just the count .

I can however improve this query by avoiding the second loop in getActualPendimg

x/crosschain/keeper/keeper_cross_chain_tx.go
```diff
@@ -3,6 +3,7 @@ package keeper
 import (
 	"context"
 	"fmt"
+	goMath "math"
```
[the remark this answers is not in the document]
lumtis: I would personally use instead
```
import (
  "math"
  sdkmath "cosmossdk.io/math"
)
```

x/crosschain/keeper/keeper_cross_chain_tx.go
lumtis, line 171: I think we should have queries in files prefixed with `grpc` to make it consistent throughout the codebase and easier to find

x/crosschain/keeper/keeper_cross_chain_tx.go
```diff
@@ -157,16 +158,39 @@ func (k Keeper) CctxAllPending(c context.Context, req *types.QueryAllCctxPending
...
+		pendingCrossChainTxList, err := k.GetActualPendingCCTX(ctx, chain.ChainId)
+		if err != nil {
+			return nil, err
+		}
+		count += int64(len(pendingCrossChainTxList))
+	}
+	return &types.QueryAllCctxPendingCountResponse{CctxPendingCount: count}, nil
+}
+
+func (k Keeper) GetActualPendingCCTX(ctx sdk.Context, chainID int64) ([]*types.CrossChainTx, error) {
```
lumtis, line 184: What does `actual` means? We should add comment to explain. Shouldn't be called `GetPendingCCTXByChainID`?

x/crosschain/keeper/keeper_cross_chain_tx.go
brewmaster012: see this http://46.4.15.110:1317/zeta-chain/crosschain/outTxTracker

at the end: 
image

This is the standard behavior of pagination; it returns a total count regardless of how many records is returned in this page. 

If you just need the count, query with page limit=1, and get the count. 
Isn't this better than a separate RPC just for count?

brewmaster012: Here's an example of outTxTracker: 
http://46.4.15.110:1317/zeta-chain/crosschain/outTxTracker?pagination.limit=1

brewmaster012: hmm... the total is 0 in this case. Wut??!

@lumtis do you know whether this is expected behavior of pagination?

Issue #1037, opened 2023-08-24
Emergency pause of ZRC-20 contract


fadeev:

lumtis:
Does a solution for this have been already designed @zeta-chain/protocol-engineering ?
I think a solution can be:
- Adding a `paused` boolean in `ForeignCoin` structure
- In `PostTx` evm hook, verify if a paused zrc20 has been called, if yes revert the tx
- Implement message for updating paused value, we could have an array to pause several contract in the same message

Issue #932, opened 2023-08-09
`observer`: Revise policy type names


lumtis:
Context: https://github.com/zeta-chain/node/pull/912#discussion_r1287334012

Make a walkthrough of the different policy types and rename them appropriately depending on their usage.

lumtis:
Did a walkthrough of the types:

- `Policy_Type_stop_inbound_cctx`: can update permission flags to disable inbound and outbound in cctx.

**Should be renamed**

- `Policy_Type_deploy_fungible_coin` - can:
  - Whitelist ERC20
  - Deploy foreign coin
  - Remove foreign coin
  - Update system contract
  - Pause ZRC20
  - Update ZRC20 bytecode
  - Update withdraw fees

**Should be renamed**

- `Policy_Type_update_client_params` - update core params, looks up-to-date although `client_params` could be renamed into `core_params`

- `Policy_Type_update_keygen_block` - update TSS address and Keygen

**Should be renamed**

- `Policy_Type_out_tx_tracker` - Add to out tx tracker or remove from out tx tracker, looks up-to-date

- `Policy_Type_add_observer` - Add observer, looks up-to-date

We might also want to merge some of the rule to simplify the authorization structure.

cc @zeta-chain/protocol-engineering

fadeev:
Create a doc with all the privileged message types. Depending on how complex this ends up being, we might consolidate the roles.

lumtis:
Update: doc set up on Notion: Engineering/Protocol Docs/Observer Policy Types

fadeev:
https://www.notion.so/zetachain/Observer-Policy-Types-4e501dd64dfc460180a361f9b8507310

PR #1096, opened 2023-09-06
feat(`fungible`): add message to pause/unpause zrc20

lumtis:
# Description

- Add `Paused` field in `ForeignCoin`
- The message `MsgUpdateZRC20PausedStatus` provides a list of ZRC20 to update. If the action is pause, all zrc20 get paused, and vice versa.
- `fungible` add a `PostTxProcessing` into `evm`, it checks addresses of contract emitting events, if an event is emitted from a paused zrc20, the tx is reverted. All actions in ZRC20 interactions emit events, therefore is prevent doing any actions on it once paused

This currently blocks actions to the ZRC20 at the smart contract level, maybe there are additional logic to add like blocking cctxs related to the zrc20?

Closes: https://github.com/zeta-chain/node/issues/1037

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


x/fungible/keeper/evm_hooks.go
```diff
@@ -0,0 +1,57 @@
...
+type EVMHooks struct {
+	k Keeper
+}
+
+func (k Keeper) EVMHooks() EVMHooks {
+	return EVMHooks{k}
+}
+
+// PostTxProcessing is a wrapper for calling the EVM PostTxProcessing hook on the module keeper
+func (h EVMHooks) PostTxProcessing(ctx sdk.Context, _ core.Message, receipt *ethtypes.Receipt) error {
```
kingpinXD, line 24: Is there a reason to not use the hooks implemented by crosschain module ?

x/fungible/keeper/msg_server_update_zrc20_paused_status.go
```diff
@@ -0,0 +1,65 @@
...
+) (*types.MsgUpdateZRC20PausedStatusResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// check message validity
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, err
+	}
+
+	// check if the sender is the admin
+	if msg.Creator != k.observerKeeper.GetParams(ctx).GetAdminPolicyAccount(zetaObserverTypes.Policy_Type_deploy_fungible_coin) {
```
kingpinXD, line 28: Maybe lets create a new policy type ? What do you think ?

kingpinXD (commented):
left some minor comments . 
I do feel it might be a good idea to concetrate all the administration related logic to observer module or a separate module improted by observer . However that would be a much bigger refactior and can be done seprately . Would love to get your thoughts on that .

x/fungible/keeper/msg_server_update_zrc20_paused_status.go
kingpinXD: Maybe even use `Policy_Type_stop_inbound_cctx`and rename it to `Policy_Update_Permissions`

brewmaster012: @kingpinXD let's consolidate the policy types. We should not have more than 2, at most 3.

x/fungible/keeper/evm_hooks.go
lumtis: I think it's proper to define it in `fungible` because it exclusively uses fungible domain. So from a modular point of view, we should be able to remove `crosschain` module and still have this functionality working.

x/fungible/keeper/msg_server_update_zrc20_paused_status.go
lumtis: Agree with this. `Policy_Type_deploy_fungible_coin` is also used for the update bytecode message.I think `Policy_Type_deploy_fungible_coin` can be renamed into something like `Policy_Type_admin_fungible_coin`.

lumtis:
>I do feel it might be a good idea to concetrate all the administration related logic to observer module or a separate module improted by observer . However that would be a much bigger refactior and can be done seprately . Would love to get your thoughts on that .

Thanks for the review. We can think about it. So far I am quite good with the message currently in `fungible`. I think we should think about modules more like a piece containing a specific state of the chain, rather than a group of actions. If a message interact with the state of a module A, so it should be in module A IMO.

brewmaster012 (approved):
LGTM
Would be nicer to have a small test in smoketest though so that we have more e2e coverage.

x/fungible/keeper/msg_server_update_zrc20_paused_status.go
lumtis: Opened this one some time ago: https://github.com/zeta-chain/node/issues/932
I think we could handle it for v10

lumtis:
@brewmaster012 
Added a smoketest that pause the ZRC20 and check:
- Can't interact with the zrc20
- Can't interact with contracts that interact with the zrc20
- Can continue to interact with other zrc20
- Can unpause and interact again with the contract

x/fungible/types/message_update_zrc20_paused_status.go
```diff
@@ -0,0 +1,64 @@
...
+		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidRequest, "invalid action (%d)", msg.Action)
+	}
+
+	if len(msg.Zrc20Addresses) == 0 {
+		return cosmoserrors.Wrap(sdkerrors.ErrInvalidRequest, "no zrc20 to update")
+	}
+
+	// check if all zrc20 addresses are valid
+	for _, zrc20 := range msg.Zrc20Addresses {
+		if ethcommon.HexToAddress(zrc20) == (ethcommon.Address{}) {
```
ws4charlie, line 59: would 'ethcommon.IsHexAddress' be a good func to do the checking?

x/fungible/keeper/evm_hooks.go
```diff
@@ -0,0 +1,57 @@
...
+			addressExist[log.Address] = struct{}{}
+			addresses = append(addresses, log.Address)
+		}
+	}
+
+	// check if any of the addresses are from a paused ZRC20 contract
+	for _, address := range addresses {
+		fc, found := k.GetForeignCoins(ctx, address.Hex())
+		if found && fc.Paused {
+			return cosmoserrors.Wrap(types.ErrPausedZRC20, address.Hex())
```
ws4charlie, line 52: will the whole evm transaction be reverted when this error is thrown? For example, a Uniswap function (e.g., swap) calls zBTC function `transferFrom` and emits a zrc20 event, the whole 'swap' will be reverted? not partially, right?

x/fungible/types/message_update_zrc20_paused_status.go
lumtis: I copied the current code. Didn't know this method exist, that's interesting!

x/fungible/keeper/evm_hooks.go
lumtis: The whole tx is reverted.
Tested here https://github.com/zeta-chain/node/blob/e04294f2700463675f46baa6f3adf4b74f7ad51a/contrib/localnet/orchestrator/smoketest/test_pause_zrc20.go#L94

kingpinXD (approved):

x/fungible/types/message_update_zrc20_paused_status.go
lumtis: b4b0f24b

x/fungible/keeper/evm_hooks.go
ws4charlie: Cool. Reverting the whole tx is a safe action.

ws4charlie (approved):
looks good

Issue #1069, opened 2023-08-31
Prevent creating an outbound tx if cctx outbound is reverted and sender is ZetaChain


lumtis:
When a cctx outbound tx fails, we create a new outbound for the revert independently if the sender chain is Zetachain or not (and pay gas):
https://github.com/zeta-chain/node/blob/3046a78693971efafaffb19ffdb5710356f77649/x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go#L145

While we don't create outbound when receiving an inbound and Zetachain is receiver:
https://github.com/zeta-chain/node/blob/3046a78693971efafaffb19ffdb5710356f77649/x/crosschain/keeper/keeper_cross_chain_tx_vote_inbound_tx.go#L162

I think we should add a condition in `keeper_cross_chain_tx_vote_outbound_tx.go` and prevent creating the outbound if the sender is Zetachain.

brewmaster012:
yeah, the cctx should probably go into aborted, because this suggests serious error (not only exceptions). 

coming to think of it, we don't really distinguish CCTX exceptions (kind of expected, such as fees not enough) from errors (contract missing, inconsistent internal state, etc). 

maybe we should add distinguish those two?

lumtis:
> because this suggests serious error (not only exceptions) as outbound request from zEVM are only transfer fungible tokens (including ZETA) which should not fail.

Don't we consider for the future more advanced uses where there could be an actual revert on the external chain?

> maybe we should add distinguish those two?

Agree with this, maybe we should first identify all the places where the distinction could exist, then think about an additional status?

PR #1107, opened 2023-09-11
fix(`crosschain`): set cctx as aborted if outbound fails and sender is zeta chain

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/1069

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/crosschain/keeper/keeper_cross_chain_tx_vote_outbound_tx.go
```diff
@@ -151,7 +151,8 @@ func (k msgServer) VoteOnObservedOutboundTx(goCtx context.Context, msg *types.Ms
 			newStatus := cctx.CctxStatus.Status.String()
 			EmitOutboundSuccess(tmpCtx, msg, oldStatus.String(), newStatus, cctx)
 		case observerTypes.BallotStatus_BallotFinalized_FailureObservation:
-			if msg.CoinType == common.CoinType_Cmd {
+			if msg.CoinType == common.CoinType_Cmd || cctx.InboundTxParams.SenderChainId == common.ZetaChain().ChainId {
```
ws4charlie, line 154: I can't tell the meaning of each status but just I'm curious.
1. As for `msg.CoinType == common.CoinType_Cmd`, I knew that it was a newly added type by Panruo meaning something special and concrete. What are the cases for `cctx.InboundTxParams.SenderChainId == common.ZetaChain().ChainId`? Does it overlap with the former?

2. 'CctxStatus_Aborted' leads to different subsequent workflow than 'CctxStatus_Reverted' or we are just setting a different flag to give it a different meaning for the moment?

3. Both line 146 and 178 uses set 'CctxStatus_Aborted' status. We can't unify the logic in one single flow?

lumtis: >As for msg.CoinType == common.CoinType_Cmd, I knew that it was a newly added type by Panruo meaning something special and concrete. What are the cases for cctx.InboundTxParams.SenderChainId == common.ZetaChain().ChainId? Does it overlap with the former?

They are unrelated, both are just two conditions when there is no revert and the cctx should become aborted.

>'CctxStatus_Aborted' leads to different subsequent workflow than 'CctxStatus_Reverted' or we are just setting a different flag to give it a different meaning for the moment?

We set it aborted because there is no revert tx when the sender is ZetaChain, also this scenario should not occur under the current logic.

 >Both line 146 and 178 uses set 'CctxStatus_Aborted' status. We can't unify the logic in one single flow?

Agree on that, although I think the general flow of vote_outbound_tx and vote_inbound_tx could be simplified/factorized. I would consider it for a larger refactoring PR.

ws4charlie: Thank you for explaining it.

ws4charlie (approved):
I'm okay with it. It's better if some one really knows the status can take a look.

kingpinXD (approved):

Issue #1103, opened 2023-09-08
Decide on parameters for cctx outbound gas price automatic increase


lumtis:
https://github.com/zeta-chain/node/blob/b2e152316e6eba4ada82771dd48a0aa26e761a5b/x/crosschain/keeper/abci.go#L15

We define the following values for gas automatic increase:
- `EpochLength`: the number of blocks before iterating pending outbound for gas price increase (current: 100 blocks)
- `RetryInterval`: the time an outbound must be pending getting gas price increase (current: 10 minutes)
- `GasPriceIncreasePercent`: the increase of the gas price each step in percent of median gas price (current: 100%)

Current values have been set more like placeholders without fully knowledgeable decision

We need to settle for those initial values.

Issue #1119, opened 2023-09-12
Rename `PermissionFlags`


lumtis:
I think the name `PermissionFlags` is misleading since it doesn't contain flag about permissions.

Current values: 
```
IsInboundEnabled
IsOutboundEnabled
```

PR #1120, opened 2023-09-12
feat(`observer`): rename `PermissionFlags` to `CrosschainFlags` and add gas price increase flags to it 

lumtis:
# Description

- Rename `PermissionFlags` into `CrosschainFlags` as a general structure for crosschain functionalities flags
  - Note: keep the same key prefix in order to be state backward compatible
- Define `GasPriceIncreaseFlags` for the gas price functionality parameters
- Add `GasPriceIncreaseFlags` to the message
- Add tests to the message
- Add event in message

**NOTE**: I also removed returning an error if the flag is not found:
https://github.com/zeta-chain/node/blob/037dc0be504e5187adfd96cd5d2fcc098cea9f3c/x/observer/keeper/msg_server_permission_flags.go#L23
This logic sounded strange to me because it would mean if the flags are not set then it can never be set ever and we always use default value? 

Closes: https://github.com/zeta-chain/node/issues/1103 https://github.com/zeta-chain/node/issues/1119

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/observer/keeper/msg_server_update_crosschain_flags.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// check if the value exists
+	flags, isFound := k.GetCrosschainFlags(ctx)
+	if !isFound {
+		flags = *types.DefaultCrosschainFlags()
+	}
+
+	// update values
+	flags.IsInboundEnabled = msg.IsInboundEnabled
+	flags.IsOutboundEnabled = msg.IsOutboundEnabled
```
kingpinXD, line 28: Maybe replace similar behavior for inbound and outbound flags as well, so that the user can just set one if needed?

kingpinXD (approved):

x/observer/keeper/msg_server_update_crosschain_flags.go
lumtis: We can do that

lumtis: We can't directly have a bool pointer in proto, we would need to define a wrapper or have another bool like `IsInboundEnabledUpdated` which I think would complexify the API.

lumtis: But we could define a structure
```
struct TxEnabled{
  IsInboundEnabled bool
  IsOutboundEnabled bool
}
```
To give the ability to update the gas price increase params without touching the inbound/outbound enable status

proto/observer/crosschain_flags.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+package zetachain.zetacore.observer;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "github.com/zeta-chain/zetacore/x/observer/types";
+
+message GasPriceIncreaseFlags {
+  int64 epochLength = 1;
```
ws4charlie, line 10: do we allow negative epochLength (e.g., special meaning)?

ws4charlie (approved):
looks good.

proto/observer/crosschain_flags.proto
lumtis: 3553a17a

PR #1135, opened 2023-09-14
test: Stateful upgrade

kevinssgh:
# Description

Added new target in makefile that will run the smoketest twice and perform a binary update between. The old and new versions to be tested can be passed into the makefile target as environment variables. 

* zetavisor and upgrade binaries are integrated in dockerfile
* new script added to start network and create gov proposal for zetacore upgrade - zetavisor will do this automatically
* updated smoketest to optionally start at a particular height

ex. 

stateful-upgrade:
	@echo "--> Starting stateful smoketest"
	$(DOCKER) build --build-arg old_version=v9.0.0 --build-arg new_version=v9.0.6-test -t zetanode -f ./Dockerfile-versioned .
	$(DOCKER) build -t orchestrator -f contrib/localnet/orchestrator/Dockerfile.fastbuild .
	cd contrib/localnet/ && $(DOCKER) compose -f docker-compose-stateful.yml up -d

Closes: https://github.com/zeta-chain/node/issues/1117

TODO: Need to create script that restarts zetaclients after upgrade.

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


contrib/localnet/orchestrator/smoketest/main.go
```diff
@@ -295,8 +303,9 @@ func LocalSmokeTest(_ *cobra.Command, _ []string) {
 	smokeTest.TestMessagePassingRevertSuccess()
 	smokeTest.CheckZRC20ReserveAndSupply()
 
-	smokeTest.TestPauseZRC20()
-	smokeTest.CheckZRC20ReserveAndSupply()
+	//DISABLED temporarily since older binaries might not work with this due to admin account
```
ws4charlie, line 306: I remember Bertrant created these two tests and he made them working on develop?

ws4charlie (approved):
looks good.

contrib/localnet/orchestrator/smoketest/main.go
lumtis: I think `TestPauseZRC20` and `TestUpdateBytcode` can be commented out since it only concerns isolated actions and don't need to be checked everytime in the CI.

But I added in the genesis generation of smoke tests the account for foreign coin admin, I'm not sure what would cause it to not work then

lumtis:
Is it still WIP @kevinssgh ?

contrib/localnet/orchestrator/smoketest/main.go
kevinssgh: I think your change is not included in older versions. This particular test path would run the smoketest on an older version, do an upgrade then run the smoketest again. It isn't guaranteed that the account will be present in the older versions.

kingpinXD (approved):

contrib/localnet/orchestrator/smoketest/main.go
lumtis: We should comment `smokeTest.TestUpdateBytecode()` for the same reason, also use the admin account.

I think we could keep those disabled since it represents tests for administration, we might want to have a specific smoketest for it.

Issue #1114, opened 2023-09-11
Scheduling of outbound transaction keysign.


fadeev:

ws4charlie:
Here's the original conversation: https://zetachain.slack.com/archives/C05QUKKM8L8/p1694280188748279?thread_ts=1694239099.400909&cid=C05QUKKM8L8
![image](https://github.com/zeta-chain/node/assets/34498985/f58641e4-6541-433a-b2a5-1d1bca1be513)

PR #1131, opened 2023-09-13
feat: add new query for pending nonces using chain id

kingpinXD:
# Description
Adds anew query `PendingNoncesByChain`

Closes: <PD-XXXX>

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

- [ ] Tested CCTX in localnet
- [ ] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

PR #1147, opened 2023-09-19
fix: improve Bitcoin keysign performance to target outbound TPM of 5

ws4charlie:
# Description

Background:
Bitcoin outTx keysign is slow in Athens-3 because it's been dealt with using same keysign strategy (given a nonce, we schedule keysign every 30 blocks) with EVM chains. A different keysign strategy needs to be figured out and to better fit Bitcoin.

1. Added special handling logic in scheduler for Bitcoin:
    1> schedule one keysign on each ticker (1 block).
    2> schedule keysign only when nonce-mark UTXO is available.
    3> stop processing when pendingNonce/lookahead is reached.
2. Added an artificial pending nonce to assist keysign scheduler.
3. Enabled connectivity check between adjacent outTx.
4. Some code adjustments and comments updating.

Close: https://github.com/zeta-chain/node/issues/1114

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


proto/crosschain/query.proto
```diff
@@ -293,7 +306,7 @@ message QueryAllCctxResponse {
 }
 
 message QueryAllCctxPendingRequest {
-  uint64 chainId = 1;
+  int64 chainId = 1;
```
kingpinXD, line 309: Maybe use chain_id?

x/crosschain/keeper/keeper_pending_nonces.go
```diff
@@ -74,13 +74,31 @@ func (k Keeper) RemovePendingNonces(ctx sdk.Context, pendingNonces types.Pending
...
 func (k Keeper) RemoveFromPendingNonces(ctx sdk.Context, tssPubkey string, chainID int64, nonce int64) {
-	p, found := k.GetPendingNonces(ctx, tssPubkey, uint64(chainID))
+	p, found := k.GetPendingNonces(ctx, tssPubkey, chainID)
 	if found && nonce >= p.NonceLow && nonce <= p.NonceHigh {
 		p.NonceLow = nonce + 1
 		k.SetPendingNonces(ctx, p)
 	}
 }
 
+func (k Keeper) PendingNoncesByChain(c context.Context, req *types.QueryPendingNoncesByChainRequest) (*types.QueryPendingNoncesByChainResponse, error) {
```
kingpinXD, line 84: This is already implemented here : https://github.com/zeta-chain/node/pull/1131

zetaclient/bitcoin_client.go
```diff
@@ -623,23 +637,69 @@ func (ob *BitcoinChainClient) fetchUTXOS() error {
 	return nil
 }
 
+// This func tries increasing the artificial pending nonce of outTx (if lagged behind).
```
lumtis, line 640:
```suggestion
// refreshPendingNonce tries increasing the artificial pending nonce of outTx (if lagged behind).
```

zetaclient/bitcoin_client.go
```diff
@@ -623,23 +637,69 @@ func (ob *BitcoinChainClient) fetchUTXOS() error {
...
+		if txid == "" {
+			return "", fmt.Errorf("getOutTxidByNonce: cannot find outTx txid for nonce %d", nonce)
+		}
+		// make sure it's a real Bitcoin txid
+		_, getTxResult, err := ob.GetTxResultByHash(txid)
+		if err != nil {
+			return "", errors.Wrapf(err, "getOutTxidByNonce: error getting outTx result for nonce %d hash %s", nonce, txid)
+		}
+		if getTxResult.Confirmations <= 0 { // just a double check
+			return "", fmt.Errorf("getOutTxidByNonce: outTx txid %s for nonce %d is not confirmed", txid, nonce)
```
lumtis, line 698:
```
is not confirmed
```
Shouldn't we say `is not included`?

zetaclient/zetacore_observer.go
```diff
@@ -124,8 +128,7 @@ func (co *CoreObserver) startSendScheduler() {
 							continue
 						}
 						signer := co.signerMap[*c]
-						chainClient := co.clientMap[*c]
-						sendList, err := co.bridge.GetAllPendingCctx(uint64(c.ChainId))
+						sendList, err := co.bridge.GetAllPendingCctx(c.ChainId)
```
lumtis, line 131: I think we should rename send into cctx to be clearer

zetaclient/zetacore_observer.go
```diff
@@ -160,7 +163,18 @@ func (co *CoreObserver) startSendScheduler() {
 						for idx, send := range sendList {
 							params := send.GetCurrentOutTxParam()
 							if params.ReceiverChainId != c.ChainId {
-								log.Warn().Msgf("mismatch chainid: want %d, got %d", c.ChainId, params.ReceiverChainId)
+								co.logger.ZetaChainWatcher.Warn().Msgf("mismatch chainid: want %d, got %d", c.ChainId, params.ReceiverChainId)
+								continue
+							}
+							currentHeight := uint64(bn)
```
lumtis, line 169:
```suggestion
							}
							// #nosec G701 range is verified
							currentHeight := uint64(bn)
```

zetaclient/zetacore_observer.go
```diff
@@ -160,7 +163,18 @@ func (co *CoreObserver) startSendScheduler() {
...
+							nonce := params.OutboundTxTssNonce
+							outTxID := fmt.Sprintf("%s-%d-%d", send.Index, params.ReceiverChainId, nonce) // should be the outTxID?
```
lumtis, line 171: What does the comment mean?

x/crosschain/keeper/keeper_pending_nonces.go
ws4charlie: Nice. Do you want to merge your PR into mine or you want to merge it into develop first?

zetaclient/bitcoin_client.go
```diff
@@ -623,23 +637,69 @@ func (ob *BitcoinChainClient) fetchUTXOS() error {
...
+	ob.mu.Lock()
+	pendingNonce := ob.pendingNonce
+	ob.mu.Unlock()
+
+	if p.NonceLow > 0 && uint64(p.NonceLow) >= pendingNonce {
+		// get the last included outTx hash
+		txid, err := ob.getOutTxidByNonce(uint64(p.NonceLow)-1, false)
+		if err != nil {
+			ob.logger.ChainLogger.Error().Err(err).Msg("refreshPendingNonce: error getting last outTx txid")
+		}
```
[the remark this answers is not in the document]
ws4charlie: You're right. This refreshing tries to update the pending nonce without caring about errors (it will succeed sooner or later). As I put in the `refreshPendingNonce` func's comment, there could be many (unpredictable) reasons for a pending nonce lagging behind, for example:
1. The zetaclient gets restarted.
2. The tracker is missing in zetacore.
image

zetaclient/zetacore_observer.go
ws4charlie: Cool

zetaclient/zetacore_observer.go
ws4charlie: This one was pretty old when some one unsure about the best pattern of the ID to be used in `OutTxProcessorManager`. I can rephrase the comment in case we need to improve it in future.

lumtis: I don't think we have a value `outTxID`, do you mean `outTxHash`? I think it could be used think we can safely assume it is unique

zetaclient/zetacore_observer.go
```diff
@@ -227,6 +234,40 @@ func (co *CoreObserver) startSendScheduler() {
...
+func (co *CoreObserver) processBitcoinOutTx(outTxMan *OutTxProcessorManager, idx int, send *types.CrossChainTx, signer ChainSigner, ob ChainClient, currentHeight uint64) bool {
+	params := send.GetCurrentOutTxParam()
+	nonce := params.OutboundTxTssNonce
+	lookahead := uint64(ob.GetCoreParams().OutboundTxScheduleLookahead)
+	outTxID := fmt.Sprintf("%s-%d-%d", send.Index, params.ReceiverChainId, nonce)
+
+	// start go routine to process outtx
+	outTxMan.StartTryProcess(outTxID)
+	co.logger.ZetaChainWatcher.Debug().Msgf("Sign bitcoin outtx %s with value %d\n", outTxID, params.Amount)
+	go signer.TryProcessOutTx(send, outTxMan, outTxID, ob, co.bridge, currentHeight)
```
lumtis, line 252: It's a bit hard to visualize how concurrency is managed with this Go routine. Is the Go routine communicating with `outTxMan`? What is the process to wait/determine the out tx is processed?

zetaclient/zetacore_observer.go
ws4charlie: Sorry, it's an identifier that only used in zetaclient. Here is how outTxID is calculated: https://github.com/zeta-chain/node/blob/btc-keysign-schedule/zetaclient/chainclient.go#L24
image

zetaclient/zetacore_observer.go
ws4charlie: The logic here is to make sure there will be only one go routine processing the cctx with identifier `outTxID`. It's controlled by a defer function inside function `TryProcessOutTx` (both EVM and Bitcoin chains)
https://github.com/zeta-chain/node/blob/btc-keysign-schedule/zetaclient/btc_signer.go#L205

zetaclient/zetacore_observer.go
```diff
@@ -155,17 +158,29 @@ func (co *CoreObserver) startSendScheduler() {
...
+								co.logger.ZetaChainWatcher.Error().Msgf("mismatch chainid: want %d, got %d", c.ChainId, params.ReceiverChainId)
+								continue
+							}
+							// #nosec G701 range is verified
+							currentHeight := uint64(bn)
+							nonce := params.OutboundTxTssNonce
+							outTxID := fmt.Sprintf("%s-%d-%d", cctx.Index, params.ReceiverChainId, nonce) // would outTxID a better ID?
+
+							// Process Bitcoin OutTx
+							if common.IsBitcoinChain(c.ChainId) && !outTxMan.IsOutTxActive(outTxID) {
```
kevinssgh, line 175: Are you processing and retrying bitcoin outbound txns every iteration? Will this cause a lot of unnecessary tss signing?

zetaclient/zetacore_observer.go
```diff
@@ -227,6 +234,40 @@ func (co *CoreObserver) startSendScheduler() {
 	}
 }
 
+// Bitcoin outtx is processed in a different way
+// 1. schedule one keysign on each ticker
+// 2. schedule keysign only when nonce-mark UTXO is available
```
kevinssgh, line 239: Is it possible the same nonce mark utxo will be available for more than one iteration? Once the txn has been broadcasted, I guess the utxo must be spent right away for this to work.

kevinssgh (commented):
Just a few comments

zetaclient/bitcoin_client.go
```diff
@@ -623,23 +637,69 @@ func (ob *BitcoinChainClient) fetchUTXOS() error {
...
+func (ob *BitcoinChainClient) refreshPendingNonce() {
+	// get pending nonces from zetacore
+	p, err := ob.zetaClient.GetPendingNoncesByChain(ob.chain.ChainId)
+	if err != nil {
+		ob.logger.ChainLogger.Error().Err(err).Msg("refreshPendingNonce: error getting pending nonces")
+	}
+
+	// increase pending nonce if lagged behind
+	ob.mu.Lock()
+	pendingNonce := ob.pendingNonce
```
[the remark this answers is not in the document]
ws4charlie: It is initialized in here (right below the code you commented):
image
https://github.com/zeta-chain/node/blob/btc-keysign-schedule/zetaclient/bitcoin_client.go#L666

zetaclient/zetacore_observer.go
ws4charlie: It's a good point. It schedules at most 1 keysign per Zeta block now to target TPM of 5. This part still has room to be optimized in some way. (e.g., if we already have a broadcasted outTx, not included yet, we should wait?)

zetaclient/zetacore_observer.go
ws4charlie: Yeah. This is the way I had assumed that the Bitcoin node should work. The bitcoin node should be able to check double-spending on our nonce-mark UTXO. Once it get spent (confirmation == 0), it should disappear from unspent UTXO list (by query).

zetaclient/zetacore_observer.go
ws4charlie: @kevinssgh [cca3a](https://github.com/zeta-chain/node/pull/1147/commits/cca3a22d92c7b7a2ccbb6556d77ceacbdf398d39)

kevinssgh (approved):
Looks Good

PR #1164, opened 2023-09-21
fix: Initialized prometheus metrics for tss module

kevinssgh:
# Description

There is a crash when the keygen fails and the blame counter attempts to increment. This is caused by an uninitialized metrics object. It was initialized after the keygen. 

Closes: https://github.com/zeta-chain/node/issues/1165

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet


kevinssgh:
Duplicate fix

PR #1214, opened 2023-09-28
refactor: add more extensive check for reverted contract

lumtis:
# Description

Right now, to check if a contract is reverted, we check if the returned `res.Failed() == true`.
There can be situation where when `CallEVM` fail, the returned `res` will be `nil` and the error with contain the description `execution reverted`.
Since `depositAndCall` logic could be updated in the future and create a situation where `res` could become `nil` after reverted execution, this PR adds a more extensive check for the contract reverted taking this situation into account.

The PR also add more smoke test to test `DepositAndCall` with ethers.

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


brewmaster012:
i vaguely remember this patch might be already incorporated in another PR, is that right? @lumtis

lumtis:
> i vaguely remember this patch might be already incorporated in another PR, is that right? @lumtis

This PR is the result of the initial investigation I had on Denis' non-reverted cctx issue. I realize that the check for reverted tx we have doesn't handle all cases (the tx might revert and the check doesn't detect). It doesn't fix a current issue we have in the node right now (always detect reverted) but it might become an issue if we change the `DepositAndCall` logic in the future

lumtis:
Fix for https://github.com/zeta-chain/node/issues/1347 : https://github.com/zeta-chain/node/pull/1214/commits/b3af0bdf41c86ea6ff247e96fe3d5d10f521b7cc

ws4charlie (approved):
looks good

brewmaster012 (commented):
~Does this PR address this issue? #1347~
OK saw your previous comment. Is it possible to test the error message (evm reverted: ret 0x???) in the smoketest?

lumtis:
> Alternatively, if the mocks in unit tests can simulate a remote contract call and result then that'll be great too.

I have one unit test here that calls a reverted function:
https://github.com/zeta-chain/node/blob/9a7a7812fef15198f57934fb9b337c19fe1e7a89/x/fungible/keeper/evm_test.go#L358

If we already have the smoke test calling a reverted function, I can add an assert in the smoke test

lumtis:
@brewmaster012 added check in the smoke test
[0b2f828](https://github.com/zeta-chain/node/pull/1214/commits/0b2f828a9b38d58afa051fffd70ffa627c7a7ae1)
[9735294](https://github.com/zeta-chain/node/pull/1214/commits/9735294bfea5c7aa3d1c860e955f8f793dedec9d)

brewmaster012 (approved):
LGTM!

PR #1260, opened 2023-10-09
feat(`fungible`): add ability to update gas limit

lumtis:
# Description

Update `UpdateZRC20WithdrawFee` to add the ability to update the gas limit of a zrc20 along with the protocol fee.
If one of the value is not defined, the value is not updated for the zrc20

Closes: https://github.com/zeta-chain/node/issues/1256

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


brewmaster012 (approved):
LGTM
seems a more appropriate name for the MsgUpdateZRC20WithdrawFee is needed
could be in another PR though.

lumtis:
> LGTM seems a more appropriate name for the MsgUpdateZRC20WithdrawFee is needed could be in another PR though.

`ProtocolFee` and `GasLimit` are both used to calculate the final withdraw fee, that's why I was thinking this is convenable but I agree we can think about another name

PR #1263, opened 2023-10-10
feat: Bitcoin block header and merkle proof

ws4charlie:
# Description

1. Added Bitcoin block header support in zetacore.
1. Added support for unpriviledged AddOutTxTracker for observed Bitcoin outTx.
2. Added a RPC to check Bitcoin merkle proofs.
3. Added a smoketest to test the Bitcoin merkle proof generation and use the prove RPC to verify proof validity

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


common/headers_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/rpcclient"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/zeta-chain/zetacore/common"
+
+	"testing"
```
lumtis, line 17:
```suggestion
	"bytes"
	"encoding/base64"
	"fmt"
	"log"
	"testing"
	"time"

	"github.com/btcsuite/btcd/blockchain"
	"github.com/btcsuite/btcd/chaincfg"
	"github.com/btcsuite/btcd/chaincfg/chainhash"
	"github.com/btcsuite/btcd/rpcclient"
	"github.com/btcsuite/btcd/wire"
	"github.com/zeta-chain/zetacore/common"
```

common/headers_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+func TestTrueBitcoinHeader(t *testing.T) {
+	blocks := LoadTestBlocks()
+
+	for _, b := range blocks.Blocks {
+		// Deserialize the header bytes from base64
+		headerBytes, err := base64.StdEncoding.DecodeString(b.HeaderBase64)
+		if err != nil {
+			t.Error(err)
+		}
```
lumtis, line 30: I think we could use `github.com/stretchr/testify/require` like we use in tests throughout the codebase 
```
require.Error(t, err)
```

common/headers_test.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+func BitcoinHeaderValidationLiveTest(t *testing.T) {
+	client := createBTCClient(t)
+	bn, err := client.GetBlockCount()
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("Verifying block headers in block range [%d, %d]\n", bn-numHeadersToTest+1, bn)
```
lumtis, line 60: We don't need `Printf` in unit tests since we can debug those

common/proof_test.go
```diff
@@ -2,14 +2,183 @@ package common_test
```
lumtis, line 2: Same comment for the file, I think we can use `github.com/stretchr/testify/require` otherwise the tests look good

common/proof_test.go
```diff
@@ -2,14 +2,183 @@ package common_test
...
+	HeaderBase64 string `json:"headerBase64"`
+	BlockBase64  string `json:"blockBase64"`
+}
+
+type Blocks struct {
+	Blocks []Block `json:"blocks"`
+}
+
+func LoadTestBlocks() Blocks {
+	file, err := os.Open("./bitcoin/test_blocks.json")
```
lumtis, line 43: We should use `TempDir` for test file
```
	tempDir := os.TempDir()
	file, err := os.Create(filepath.Join(tempDir, "test_blocks.json"))
```

common/utils.go
```diff
@@ -0,0 +1,51 @@
+package common
+
+import (
+	"encoding/hex"
+	"fmt"
+	math "math"
```
lumtis, line 6:
```suggestion
	"math"
```

common/utils.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	DustUTXOOffset = 2000
+)
+
+// A very special value to mark current nonce in UTXO
+func NonceMarkAmount(nonce uint64) int64 {
+	if nonce > math.MaxInt64-DustUTXOOffset {
+		panic("nonce mark is too large")
```
lumtis, line 19: We should use `panic` in functions in `common`

contrib/localnet/orchestrator/smoketest/test_bitcoin.go
```diff
@@ -165,6 +168,87 @@ func (sm *SmokeTest) DepositBTC() {
...
+		panic("should verify merkle proof")
+	}
+
+	// verify merkle proof through RPC
+	res, err := sm.observerClient.Prove(context.Background(), &observertypes.QueryProveRequest{
+		ChainId:   common.BtcRegtestChain().ChainId,
+		TxHash:    txHash.String(),
+		BlockHash: blockHash.String(),
+		Proof:     common.NewBitcoinProof(txBytes, path, index),
+		TxIndex:   0, // bitcoin doesn't use txIndex
```
lumtis, line 243: >// bitcoin doesn't use txIndex

Maybe it's an attribute we can add in the generic `Proof` object then. Not needed for this PR

x/crosschain/keeper/keeper_out_tx_tracker.go
```diff
@@ -271,6 +268,76 @@ func (k msgServer) AddToOutTxTracker(goCtx context.Context, msg *types.MsgAddToO
...
+	}
+	if len(tx.MsgTx().TxOut) < 1 {
+		return fmt.Errorf("outTx should have at least one output")
+	}
+	if tx.MsgTx().TxOut[0].Value != common.NonceMarkAmount(msg.Nonce) {
+		return fmt.Errorf("want nonce mark %d, got %d", tx.MsgTx().TxOut[0].Value, common.NonceMarkAmount(msg.Nonce))
+	}
+	if tx.MsgTx().TxHash().String() != msg.TxHash {
+		return fmt.Errorf("want tx hash %s, got %s", tx.MsgTx().TxHash(), msg.TxHash)
+	}
```
lumtis, line 337: Can we create an issue for the test implementation of those files?
Are `ValidateEVMOutTxBody` and `ValidateBTCOutTxBody`? It seems to me there could be made more generalized to be used with the in tx tracker as well

x/observer/keeper/grpc_query_prove.go
```diff
@@ -4,41 +4,63 @@ import (
...
-	eth "github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/zeta-chain/zetacore/x/observer/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+// Prove simply checks two things:
+// 1. the block header is available
+// 2. the proof is good
```
lumtis, line 19:
```suggestion
// 2. the proof is valid
```

x/observer/keeper/grpc_query_prove.go
```diff
@@ -4,41 +4,63 @@ import (
...
+	txBytes, err := req.Proof.Verify(res.Header, int(req.TxIndex))
 	if err != nil && !common.IsErrorInvalidProof(err) {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if err == nil {
-		var txx ethtypes.Transaction
-		err = txx.UnmarshalBinary(val)
-		if err != nil {
-			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to unmarshal transaction: %s", err))
+		if common.IsEVMChain(req.ChainId) {
```
lumtis, line 42: I think the current Ethereum proof system is for Ethereum specifically and not EVM, and we should use `IsEthereumChain` instead

Maybe @brewmaster012 can confirm

Same comment for the other places where we use `IsEVMChain`

zetaclient/bitcoin_client.go
```diff
@@ -67,12 +69,14 @@ type BitcoinChainClient struct {
...
 	ts     *TelemetryServer
+
+	BlockCache *lru.Cache
 }
 
 const (
 	minConfirmations = 0
 	maxHeightDiff    = 10000
-	dustOffset       = 2000
+	btcBlocksPerDay  = 144
```
lumtis, line 79: Is it the same value for testnet?

common/headers_test.go
ws4charlie: I'll use `require.NoError(t, err)` to avoid log prints.

common/proof_test.go
ws4charlie: sure

common/proof_test.go
ws4charlie: test_blocks.json is a piece of test data file (located in `common/test_data/`) in the repo.

common/utils.go
ws4charlie: Cool. This panic can impact zetacore. I'll remove this panic and just perform a static cast assuming the nonce will always fall in range of `MaxInt64` which means almost infinite number of transactions.

contrib/localnet/orchestrator/smoketest/test_bitcoin.go
ws4charlie: That's a good point. Here, the `index` is the sibling index (flags) on the path to merkle root. It IS a general param in merkle proof but we can wait until it becomes general when Zetachain gets connected to more types of chains (more types of merkle trees involved).

x/crosschain/keeper/keeper_out_tx_tracker.go
ws4charlie: That's a cool idea. As far as I can see, the `ChainID`, `TxHash`, `Nonce (evm only)` checking are general for both outTx and inTx. Other parameters' checking can be different (). I can reach out to @kingpinXD and see what checks he's going to perform.

x/observer/keeper/grpc_query_prove.go
ws4charlie: Good question. Current code base only accepts Ethereum block headers at [ValidateBasic](https://github.com/zeta-chain/node/blob/bitcoin-header-N-proof/x/observer/types/messages_add_block_header.go#L55)
We should explicitly restrict which chains are enabled in the source code in a consistent manner (on messages of `MsgAddBlockHeader`, `QueryProveRequest) and `MsgAddToOutTxTracker`)

lumtis: >We should explicitly restrict which chains are enabled in the source code in a consistent manner (on messages of MsgAddBlockHeader, QueryProveRequest) and MsgAddToOutTxTracker`)

I meant it would solve it if we replace `IsEVMChain` with `IsEthereumChain` in the different locations, isn't it?

zetaclient/bitcoin_client.go
ws4charlie: I think there is no defined rule on `dustOffset` for neither of them (testnet, mainnet). The block time is 10 mins on average in mainnet and very variable in testnet3 (1sec to 30 mins). Looking back to this variable, the LRU cache probably can have any reasonable length (e.g. 10, 100, 500, 1000) to avoid unnecessary RPC queries as long as it won't take too much resources and impact performance. I just put a meaningful name `btcBlocksPerDay` be a bit clear for people.

lumtis: So having `btcBlocksPerDay  = 144` would not be an issue when using ZetaClient with Bitcoin testnet?

x/crosschain/keeper/keeper_out_tx_tracker.go
lumtis: We can see during the merge

lumtis (approved):
Looking good with the smoke tests
We'll have to do excperimentation

x/observer/keeper/grpc_query_prove.go
brewmaster012: Current Ethereum proof is based and tested on Ethereum only, so most relaible on Ethereum. 
Most likely they also work for other EVM chains, if they are not messed up too much (deviating from go-ethereum too much). Klaytn might have problems, Polygon is unknown, BSC might be ok. 

For now I think IsEVMChain() because it basically means Ethereum/BSC.

common/proof.go
```diff
@@ -37,6 +41,20 @@ func NewEthereumProof(proof *ethereum.Proof) *Proof {
...
+				TxBytes: txBytes,
+				Path:    path,
+				// #nosec G701 always in range
+				Index: uint32(index),
+			},
+		},
+	}
+}
+
 // Verify verifies the proof against the header
```
brewmaster012, line 58: might need to add "returns the verified tx in bytes"

x/crosschain/keeper/keeper_out_tx_tracker.go
```diff
@@ -268,6 +265,76 @@ func (k msgServer) AddToOutTxTracker(goCtx context.Context, msg *types.MsgAddToO
...
+	if txx.Nonce() != msg.Nonce {
+		return fmt.Errorf("want nonce %d, got %d", txx.Nonce(), msg.Nonce)
+	}
+	if txx.Hash().Hex() != msg.TxHash {
+		return fmt.Errorf("want tx hash %s, got %s", txx.Hash().Hex(), msg.TxHash)
+	}
+	return nil
+}
+
+// ValidateBTCOutTxBody validates the SetWit sender address, nonce and chain ID.
```
brewmaster012, line 300: typo: SegWit

brewmaster012 (approved):
Excellent PR!
Looking forward to its merge and deployment and the stuff it enabled us to do.

PR #1264, opened 2023-10-10
fix: Blame index update

kevinssgh:
# Description

* Updated format of blame index: chainId-nonce-msgdigest-height
* Added query for blame records by chainId and nonce
* Added cmd for new query

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


proto/observer/query.proto
```diff
@@ -211,6 +216,15 @@ message QueryAllBlameRecordsResponse {
   repeated Blame blame_info = 1;
 }
 
+message QueryBlameByChainAndNonceRequest {
+  uint32 chain_id = 1;
```
ws4charlie, line 220: can we unify chain_id type as `int64` like other in other `query.proto` files? The `QueryProveRequest` one was was already updated in one of my pending PRs.

ws4charlie (approved):
LGTM

x/observer/keeper/blame.go
```diff
@@ -40,6 +41,21 @@ func (k Keeper) GetAllBlame(ctx sdk.Context) (BlameRecords []*types.Blame, found
 	return
 }
 
+func (k Keeper) GetBlameByChainAndNonce(ctx sdk.Context, chainID int64, nonce int64) (BlameRecords []*types.Blame, found bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.BlameKey))
+	blamePrefix := fmt.Sprintf("%d-%d", chainID, nonce)
```
lumtis, line 46: I looked into the code, I see that ZetaClient index the blame index with:
```
func digestReq(request keygen.Request) (string, error) {
	bytes, err := json.Marshal(request)
	if err != nil {
		return "", err
	}

	hasher := sha3.NewLegacyKeccak256()
	hasher.Write(bytes)
	digest := hex.EncodeToString(hasher.Sum(nil))

	return digest, nil
}
```
Therefore, I'm not sure to understand how they would be indexed with `[chainID]-[nonce]`

lumtis (commented):
We should use plural blames since the query returns a list of blames

x/observer/keeper/blame.go
kevinssgh: This particular digest is for keygen. There won't be any chain or nonce associated with it, I can however add a different prefix to this like "keygen-" to differentiate it. 

The change was really intended for keysign failures.

x/observer/keeper/blame_test.go
```diff
@@ -0,0 +1,49 @@
...
+	index := fmt.Sprintf("%d-%d-%s-%d", chainId, nonce, digest, 123)
+
+	keeper.SetBlame(ctx, &types.Blame{
+		Index:         index,
+		FailureReason: "failed to join party",
+		Nodes:         nil,
+	})
+
+	blameRecords, found := keeper.GetBlame(ctx, index)
+	assert.True(t, found)
```
lumtis, line 27: I think we can use `require` insert of `assert` everywhere

zetaclient/tss_signer.go
```diff
@@ -112,7 +112,7 @@ func (tss *TSS) Sign(digest []byte, height uint64, chain *common.Chain, optional
 		log.Warn().Msgf("keysign status FAIL posting blame to core, blaming node(s): %#v", ksRes.Blame.BlameNodes)
 
 		digest := hex.EncodeToString(digest)
-		index := fmt.Sprintf("%s-%d", digest, height)
+		index := fmt.Sprintf("%d-%d-%s-%d", chain.ChainId, nonce, digest, height)
```
lumtis, line 115: We generally have the code to calculate prefixes in `keys.go`, `x/observer/types/keys.go`
Maybe we could implement:
```
GetBlameIndex(chainID, nonce, digest, height)
GetBlamePrefix(chainID, nonce)
```
to use it here and in the query, so the low level aspect of calculating the prefix index is abstracted

zetaclient/tss_signer.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
+	observerTypes "github.com/zeta-chain/zetacore/x/observer/types"
```
lumtis, line 8:
```suggestion
	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
```

lumtis, line 8: We should run `make generate`

lumtis (commented):
Minor comments, looks good

lumtis (approved):

Issue #1288, opened 2023-10-13
`gosec` issues raised


lumtis:
Some `gosec` issues that needs to be fixed are raised in https://github.com/zeta-chain/node/pull/1279 (although not related to the PR)

The first two are false positive and we can use `nosec`, we should look into the two others

```
[/github/workspace/x/emissions/client/tests/suite.go:90] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    89: 			BallotThreshold:      sdk.MustNewDecFromStr("0.66"),
  > 90: 			BallotStatus:         ballotStatus[rand.Intn(max-min)+min],
    91: 			BallotCreationHeight: 0,



[/github/workspace/x/observer/module_simulation.go:14] - G101 (CWE-798): Potential hardcoded credentials (Confidence: LOW, Severity: HIGH)
    13: const (
  > 14: 	opWeightMsgUpdateClientParams          = "op_weight_msg_update_client_params"
    15: 	defaultWeightMsgUpdateClientParams int = 100



[/github/workspace/rpc/websockets.go:120] - G114 (CWE-676): Use of net/http serve function that has no support for setting timeouts (Confidence: HIGH, Severity: MEDIUM)
    119: 		} else {
  > 120: 			err = http.ListenAndServeTLS(s.wsAddr, s.certFile, s.keyFile, ws)
    121: 		}



[/github/workspace/rpc/websockets.go:118] - G114 (CWE-676): Use of net/http serve function that has no support for setting timeouts (Confidence: HIGH, Severity: MEDIUM)
    117: 		if s.certFile == "" || s.keyFile == "" {
  > 118: 			err = http.ListenAndServe(s.wsAddr, ws)
    119: 		} else {
```

PR #1290, opened 2023-10-13
fix: gosec issues

lumtis:
# Description

Fix the new gosec issues in the CI

Using nosec for false positive.
Adding timeout for websocket server.

Need confirmation for the following values:
 - read timeout - 15sec
 - write timeout - 15sec
 - idle timeout - 1 minute

Pinging @zeta-chain/devops in case

Closes: https://github.com/zeta-chain/node/issues/1288

### DIFF
--- a/rpc/websockets.go
+++ b/rpc/websockets.go
@@ -25,6 +25,7 @@ import (
 	"net"
 	"net/http"
 	"sync"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


ws4charlie (approved):
I'm not 100% sure if the `idleTimeout` will impact log filter users. We can give it a try and see.

lumtis:
>I'm not 100% sure if the idleTimeout will impact log filter users. We can give it a try and see.

Maybe we can increase it to 5min?

CharlieMc0 (approved):

Issue #1313, opened 2023-10-17
Set limit for gas price increase


lumtis:
We set gas price increase options in `CrosschainFlags`.

These options determine gas price increase in percent at certain interval when a outbound doesn't get mined. We need to add a limit for the gas price increase to avoid having fees depleting the gas stability pools.

Since gas price value depends on the chain, we could consider a limit in percent: gas price can't be more than 500% of the initial gas price for the chain.

PR #1322, opened 2023-10-19
feat(`crosschain`): add gas price increase limit 

lumtis:
# Description

Add a value `gasPriceIncreaseMax` in `GasPriceIncreaseFlags` to control the max gas price. The value is represented in percent of the median gas price for the chain

Set the default gas price max at 500% of the median gas price --> value to agree on

Closes: https://github.com/zeta-chain/node/issues/1313

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


proto/observer/crosschain_flags.proto
```diff
@@ -13,6 +13,10 @@ message GasPriceIncreaseFlags {
     (gogoproto.stdduration) = true
   ];
   uint32 gasPriceIncreasePercent = 3;
+
+  // Maximum gas price increase in percent of the median gas price
+  // Default is used if 0
+  uint32 gasPriceIncreaseMax = 4;
```
ws4charlie, line 19: Do we differentiate EVM-chain from BTC-chain on default `gasPriceIncreaseMax`

lumtis: Right now `GasPriceIncreaseFlags` is general for all chains. This is why we use percentage, the `medianGasLimit` will be specific for each chains so the final max number will be different for each chains.

For Bitcoin, gas price increase is disabled.

ws4charlie: Sounds cool. Thanks.

x/crosschain/keeper/abci.go
```diff
@@ -85,6 +84,23 @@ func (k Keeper) CheckAndUpdateCctxGasPrice(
 	}
 	gasPriceIncrease := medianGasPrice.MulUint64(uint64(flags.GasPriceIncreasePercent)).QuoUint64(100)
 
+	// compute new gas price
+	currentGasPrice, err := cctx.GetCurrentOutTxParam().GetGasPrice()
+	if err != nil {
+		return math.ZeroUint(), math.ZeroUint(), err
+	}
+	newGasPrice := math.NewUint(currentGasPrice).Add(gasPriceIncrease)
```
ws4charlie, line 92: Is it possible that this `currentGasPrice` is outdated and significantly deviates (higher or lower) from the `medianGasPrice`? Are we `adjusting` the "gasPrice" based on (by percentage) the `currentGasPrice` or directly based on the dynamic `medianGasPrice`?

ws4charlie (approved):

brewmaster012 (approved):

x/crosschain/keeper/abci.go
lumtis: `currentGasPrice` can be higher than `medianGasPrice` if it has already been increased by the same method previously. The gas price increase is calculated from the median gas price.

Issue #1278, opened 2023-10-11
More advanced key management for ZetaClient (hotkey)


lumtis:
We currently use the keyring backend test for the key used by ZetaClient
https://github.com/zeta-chain/node/blob/f51f6e0cd34248f1fb160e4367a1b60dda9ae879/zetaclient/keys.go#L95

We need to allow more advanced keys management with the ZetaClient

lumtis:
An improvement is to use `ckeys.BackendOS`

The password would need to be remove from hardcoded

`broadcast.go` needs additional logic for password prompting
```
ctx  = ctx.WithInput(r)
r.Write(fmt.Sprintf("%s\n%s\n", p,  p))
```

PR #1323, opened 2023-10-20
refactor(`zetaclient`): use keyring backend file for hotkey

lumtis:
# Description

- Remove the password from the ZetaClient config 
- Add option `KeyringBackend` in ZetaClient config for the keyring backend used for the key
- Add the File keyring backend for hot key for encryption
- Use Keyring Backend file in the smoketests

When using keyring backend file, the password must be provided through the env var `HOTKEY_PASSWORD`

```
export HOTKEY_PASSWORD=foobarfoo
zetaclientd ... --keyring-backend file
```

Closes: https://github.com/zeta-chain/node/issues/1278

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/keys.go
```diff
@@ -164,3 +182,28 @@ func (k *Keys) GetPubKeySet() (common.PubKeySet, error) {
...
+// returns empty if no password is needed
+func (k *Keys) GetHotkeyPassword() (string, error) {
+	if k.GetKeybase().Backend() == ckeys.BackendFile {
+		return getHotkeyPassword()
+	}
+	return "", nil
+}
+
+// getHotkeyPassword retrieves the HOTKEY_PASSWORD environment variable
+// and returns an error if it's not defined or shorter than 8 characters.
```
CryptoFewka, line 196: I strongly encourage us to require a password that is longer than 8 characters, especially since it is something that is going to be almost exclusively programmatically used from an environment variable. A longer passphrase will provide more entropy than an exceptionally complex, but only eight character long password. 

I feel like in this instance we should encourage this better practice, even though ultimately the password will get scraped from the environment variable by an attacker. 

Actual protection for the secret likely would only come from a hardware security module.

lumtis: The 8 characters is the requirement for the Cosmos keyring, this is why I put this specific check.

I agree that HSM would be the optimal security model in practice. This PR is mainly to introduce the option to use a different keyring backend than test, we plan to add the HSM feature on top of it.

.github/workflows/build.yml
```diff
@@ -125,6 +125,7 @@ jobs:
             echo "Smoke Test Failed"
             exit 1
           fi
+          sleep 10
```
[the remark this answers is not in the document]
lumtis: There is a bug where the ZetaClient waits indefinitely without using this sleep. I can provide more log. There might be an eventual issue to fix that I think could be fixed in another PR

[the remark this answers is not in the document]
CryptoFewka: OK sweet. Yea an indefinite wait is far more pricey than 10 seconds =)

zetaclient/keys.go
CryptoFewka: 8 characters is probably fine, and if we're matching a Cosmos keyring requirement then perhaps it shouldn't differ

kingpinXD (approved):

cmd/zetacored/get_pubkey.go
```diff
@@ -10,18 +10,19 @@ import (
...
 		Args:  cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clientCtx := client.GetClientContextFromCmd(cmd)
 			tssKeyName := args[0]
-			if len(args) == 1 {
-				args = append(args, "")
+			password := ""
+			if len(args) > 1 {
+				password = args[1]
 			}
```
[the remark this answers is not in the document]
lumtis: You can provide spaces in the argument, single or double quotes `get-pubkey myTssKey "This is a very secure passphrase."` also works.

cmd/zetacored/get_pubkey.go
```diff
@@ -10,18 +10,19 @@ import (
 	"github.com/zeta-chain/zetacore/common/cosmos"
 )
 
-func AddTssToGenesisCmd() *cobra.Command {
+func GetPubKeyCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "get-pubkey [tssKeyName] [Password]",
+		Use:   "get-pubkey [tssKeyName] [password]",
 		Short: "Get you node account",
```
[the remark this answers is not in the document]
lumtis: 3d2cb02f

.github/workflows/build.yml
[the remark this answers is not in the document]
lumtis: I removed the 10 seconds sleep, and somehow the issue is no longer here. It may have been related to Github or somehow I introduced a fix at some point.

zetaclient/keys.go
```diff
@@ -164,3 +182,28 @@ func (k *Keys) GetPubKeySet() (common.PubKeySet, error) {
...
+func getHotkeyPassword() (string, error) {
+	password := os.Getenv(HotkeyPasswordEnvVar)
```
brewmaster012, line 198: As further improvement, i think storing the password in process memory is better than storing it in environment variable. The reason is that envivronment variable are in known places in a process (or maybe just the control block of a process?). I.e., I can on linux simply find the all environment variables in a running process: by
`$ cat /proc/2328594/environ `

If the password is stored in a regular process variable it will be much harder to find even if a shell is obtained by an attacker on that node that runs `zetaclientd`. 

This is not blocking this PR though so feel free to comment and resolve.

brewmaster012 (approved):

zetaclient/keys.go
lumtis: It could also be added as an additional option as well.
The `KeyringBackend` options we define in ZetaClient don't need to match exactly the `KeyringBackend` options of Cosmos (maybe should rename the option then).
We could have for example `KeyringBackendFilePasswordEnv` and `KeyringBackendFilePasswordMemory`.
I may think env var option can be convenient for teams that want to run a ZetaClient with a fullly Infrastructure as Code design.

CryptoFewka: I agree that ENV should certainly be an option, but when we publish documentation we really should educate folks on the safety of their keys. Perhaps we should even go as far as to present a ranked list of options or flowchart.

I am still encountering people who think their keys/cryptocurrency is safe on a hardware wallet and therefore do not need to protect/worry about the mnemonic's security.

.github/workflows/build.yml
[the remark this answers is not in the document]
lumtis: Added back the 10 seconds as it was failing again... Created an issue for this.

lumtis:
I will wait on a review from @kevinssgh before merging if you can have a look sometime. Thanks

kevinssgh (approved):

zetaclient/keys.go
lumtis: https://github.com/zeta-chain/node/issues/1368

CryptoFewka (approved):

PR #1350, opened 2023-10-27
chore: use team name for codeowners

lumtis:
# Description

Update the codeowners file by replacing team members handle by team handle.


lumtis:
Closing because it looks like from the `tss-lib` repo that using `Protocol-Engineering` as a codeowner ask for a review from a single person in the team. Which is therefore not relevant since we require 2 reviews from the protocol team.

Issue #1352, opened 2023-10-30
`zetaclient` : Add a new thread which runs the supply check


kingpinXD:
https://www.notion.so/zetachain/Zeta-supply-checks-3d0f82b36d294662a985ae5c8be61698?pvs=4

kingpinXD:
The problem with PendingOutbound and PendingRevert

The Following uses PendingOutbound as an example, but we can also use a similar design for PendingRevert. 
- Pending Outbound covers two stages of a CCTX.
    - Pending outbound: This is when the outbound scheduler has not picked up the TX, and the outbound has not been broadcasted.
    - Awaiting confirmation: This is when the TX is broadcasted and awaiting confirmation. The amount has already been minted or released. So, total supply or total locked values would already reflect this amount.

PR #1358, opened 2023-10-31
feat: zeta supply checks

kingpinXD:
# Description

- Adds helper functions and a thread to the zetaclient to perform the zeta supply checks.

Closes: <#1352>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


app/setup_handlers.go
```diff
@@ -19,6 +19,7 @@ func SetupHandlers(app *App) {
 		}
 		vm[observertypes.ModuleName] = vm[observertypes.ModuleName] - 1
 		return app.mm.RunMigrations(ctx, app.configurator, vm)
+
```
lumtis, line 22:
```suggestion
```

cmd/zetaclientd/start.go
```diff
@@ -240,6 +241,14 @@ func start(_ *cobra.Command, _ []string) error {
 	mo1 := mc.NewCoreObserver(zetaBridge, signerMap, chainClientMap, metrics, tss, masterLogger, cfg, telemetryServer)
 	mo1.MonitorCore()
 
+	zetaSupplyChecker, errStartingzetaSupplyChecker := mc.NewZetaSupplyChecker(cfg, zetaBridge, masterLogger)
```
lumtis, line 244:
```suggestion
	zetaSupplyChecker, errStartingZetaSupplyChecker := mc.NewZetaSupplyChecker(cfg, zetaBridge, masterLogger)
```

lumtis, line 244:
```suggestion
	zetaSupplyChecker, err := mc.NewZetaSupplyChecker(cfg, zetaBridge, masterLogger)
```
Why not just `err`?

zetaclient/zeta_supply_checker_test.go
```diff
@@ -0,0 +1,52 @@
...
+			zetaInTransit:            MustNewIntFromString("1000000000000000000"),
+			externalChainTotalSupply: MustNewIntFromString("9000000000000000000"),
+			genesisAmounts:           MustNewIntFromString("1000000000000000000"),
+			zetaTokenSupplyOnNode:    MustNewIntFromString("1000000000000000000"),
+			ethLockedAmount:          MustNewIntFromString("10000000000000000000"),
+			validate: func(t assert.TestingT, b bool, i ...interface{}) bool {
+				return assert.True(t, b, i...)
+			},
+		},
+		// Todo add more scenarios
```
lumtis, line 43: We should track issue for TODOs

lumtis (commented):
As far as I can tell `CctxByStatus` would be called every ticks?
It would be a very intensive operation iterating the cctxs in this query.

If we just need this query for aborted CCTXs we should think about a new indexing like we do for pending cctxs, or storing the result we want for the aborted cctxs directly onchain.

Another solution is to use a backend indexer for the data but in this case we should no longer consider observers posting invariant failure to the protocol because they should only rely on blockchain data in order to do this.

zetaclient/zeta_supply_checker.go
```diff
@@ -0,0 +1,230 @@
...
+			return err
+		}
+		totalSupply, err := zetatokenNonEth.TotalSupply(nil)
+		if err != nil {
+			return err
+		}
+		totalSupplyInt, ok := sdkmath.NewIntFromString(totalSupply.String())
+		if !ok {
+			zs.logger.Error().Msgf("error parsing total supply for chain %d", chain.ChainId)
+			continue
```
ws4charlie, line 113: should we `break` like above when conversion fails?

zetaclient/zeta_supply_checker.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	logger.Info().Msgf("zeta supply check passed, lhs : %s , rhs : %s", lhs.String(), rhs.String())
+	logger.Info().Msgf("--------------------------------------------------------------------------------")
+	return true
+}
+
+func (zs *ZetaSupplyChecker) AbortedTxAmount() sdkmath.Int {
+	cctxList, err := zs.zetaClient.GetCctxByStatus(types.CctxStatus_Aborted)
+	if err != nil {
+		panic(err)
```
ws4charlie, line 174: maybe allowing `AbortedTxAmount` throwing an error and avoid panic is better.

zetaclient/zeta_supply_checker.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	amount := sdkmath.ZeroUint()
+	for _, cctx := range cctxList {
+		amount = amount.Add(cctx.GetCurrentOutTxParam().Amount)
+	}
+	amountInt, ok := sdkmath.NewIntFromString(amount.String())
+	if !ok {
+		panic("error parsing amount")
```
ws4charlie, line 182: avoid panic here too?

zetaclient/zeta_supply_checker.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+		}
+
+		trackers, err := zs.zetaClient.GetAllOutTxTrackerByChain(chain, Ascending)
+		if err != nil {
+			continue
+		}
+		for _, tracker := range trackers {
+			zs.logger.Info().Msgf("tracker exists for nonce: %d , removing from supply checks", tracker.Nonce)
+			delete(nonceToCctxMap, tracker.Nonce)
```
ws4charlie, line 221: the purpose here is to exclude pending cctcs who have a tracker?

zetaclient/zeta_supply_checker.go
kingpinXD: I guess we can continue adding up the supplies to see if the final check passes

zetaclient/zeta_supply_checker.go
kingpinXD: Yes

kingpinXD:
> As far as I can tell `CctxByStatus` would be called every ticks? It would be a very intensive operation iterating the cctxs in this query.
> 
> If we just need this query for aborted CCTXs we should think about a new indexing like we do for pending cctxs, or storing the result we want for the aborted cctxs directly onchain.
> 
> Another solution is to use a backend indexer for the data but in this case we should no longer consider observers posting invariant failure to the protocol because they should only rely on blockchain data in order to do this.

Yeah, I thought of this; since we only really need the amount, storing that separately might be a good idea. But we can let this run for sometime on athens3 to see how well it holds and implement a service to keep track of this amount only.
The GetCctxByStatus is just a good-to-have query, and I agree that it should not be used for this purpose.

zetaclient/zeta_supply_checker.go
```diff
@@ -0,0 +1,236 @@
...
+	sdkmath "cosmossdk.io/math"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/rs/zerolog"
+	"github.com/zeta-chain/zetacore/common"
+	"github.com/zeta-chain/zetacore/x/crosschain/types"
+	"github.com/zeta-chain/zetacore/zetaclient/config"
+)
+
+type ZetaSupplyChecker struct {
```
lumtis, line 17: Can we define an interface for `ZetaSupplyChecker`?
We're going to use interfaces for ZetaClient unit tests which will be one of the next milestone for the ZetaClient

lumtis (commented):
>Yeah, I thought of this; since we only really need the amount, storing that separately might be a good idea. But we can let this run for sometime on athens3 to see how well it holds and implement a service to keep track of this amount only.


I think we can keep it as it for experimentation but we should have a more optimized solution for poroduction.

> The GetCctxByStatus is just a good-to-have query, and I agree that it should not be used for this purpose.

I think we should avoid having queries reading extensively in the store, here going through all cctxs. We should use paginated read in the store which is here not possible since we read and filter the object. I think the best is to have an additional indexation for the specific cctx status that we need.

lumtis (approved):

ws4charlie (approved):
looks good

Issue #1255, opened 2023-10-09
Cleaning up in/outbound trackers


fadeev:
```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/1376
- [ ] https://github.com/zeta-chain/node/issues/1377
- [ ] https://github.com/zeta-chain/node/issues/1378
- [ ] https://github.com/zeta-chain/node/issues/1379
- [ ] https://github.com/zeta-chain/node/issues/1380
- [ ] https://github.com/zeta-chain/node/issues/1381
- [ ] https://github.com/zeta-chain/node/issues/1382
```

kingpinXD:
Hi, can you add some details to the issue, is this for existing trackers or new trackers that are created?

ws4charlie:
The notion doc: https://www.notion.so/zetachain/Garbage-outtx-tracker-d1be5f1249834450b020c29f634dcc14

ws4charlie:
All 7 sub tasks done.

PR #1370, opened 2023-11-03
fix!: unconfirmed pending cctx should prevent further outbound tx signing

brewmaster012:
# Description

1. Modify the AllPending RPC to return all pending txs
2. Bound the outtx tracker size
3. Allow add to outtx tracker only if a corresonding cctx is pending
4. A MsgAddToOutTxTracker can now remove an existing tracker if no pending cctx is found. 

related: #1255 

## upgrade compatibility

1. consensus breaking
2. no need for state migration when upgrade
3. (on athens3) need to cleanup old out tx trackers to not have performance issues

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/zetacore_observer.go
```diff
@@ -178,6 +178,10 @@ func (co *CoreObserver) startSendScheduler() {
 								co.logger.ZetaChainWatcher.Error().Msgf("mismatch chainid: want %d, got %d", c.ChainId, params.ReceiverChainId)
 								continue
 							}
+							if params.OutboundTxTssNonce > cctxList[0].GetCurrentOutTxParam().OutboundTxTssNonce+120 {
```
lumtis, line 181: I think we should define a constant variable for the number `120`

proto/crosschain/tx.proto
```diff
@@ -75,7 +75,9 @@ message MsgAddToOutTxTracker {
   int64 tx_index = 7;
 }
 
-message MsgAddToOutTxTrackerResponse {}
+message MsgAddToOutTxTrackerResponse {
+  bool is_removed = 1; // if the tx was removed from the tracker due to no pending cctx
```
lumtis, line 79: It seems it is not used in the message handler returns

brewmaster012: most cases defaults to false which is correct; 
there is one place it is set to true. 
Can add more explicit assignment but maybe not littering is better as the is_remove=true is really special case.

zetaclient/zetacore_observer.go
brewmaster012: [1275932](https://github.com/zeta-chain/node/pull/1370/commits/127593241e54a40750fb9b14650bebe8683791a1)

proto/crosschain/tx.proto
lumtis: I agree to not specify the value if it is false.
But it seems it is not used when it should be true: https://github.com/zeta-chain/node/pull/1370/files#diff-9ef5905d374779457a649b7c98c6b5ed348612bcb1a43a30d0923fecb27e86bfR64

ws4charlie (approved):
looks good

proto/crosschain/tx.proto
brewmaster012: ah, yes
fixed here https://github.com/zeta-chain/node/pull/1370/commits/9edf3e9afc6772a3d747fc4ca7c1235286fb46c3

lumtis (approved):

PR #1402, opened 2023-11-16
refactor: add nonce check before adding tracker 

kingpinXD:
# Description

Add an additional check for nonce low when creating an out-tx tracker or adding a hash to it.

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/crosschain/keeper/msg_server_add_to_outtx_tracker.go
```diff
@@ -64,6 +67,18 @@ func (k msgServer) AddToOutTxTracker(goCtx context.Context, msg *types.MsgAddToO
...
+	pendingNonces, found := k.GetPendingNonces(ctx, tss.TssPubkey, msg.ChainId)
+	if !found {
+		return nil, cosmoserrors.Wrap(types.ErrCannotFindPendingNonces, fmt.Sprintf("chain_id %d, ", chain.ChainId))
+	}
+	if pendingNonces.NonceLow < 0 {
+		return nil, cosmoserrors.Wrap(types.ErrInvalidNonce, fmt.Sprintf("NonceLow is negative %d, ", chain.ChainId))
+	}
+	nonceLowUint := uint64(pendingNonces.NonceLow)
+	if nonceLowUint > msg.Nonce {
+		return nil, cosmoserrors.Wrap(types.ErrInvalidNonce, fmt.Sprintf("NonceLow %d is higher that nonce provided %d", nonceLowUint, msg.Nonce))
```
ws4charlie, line 79: I observed a few early/old cctxs having their tracker missing (never created) in chain state. So I found out the outTx hashes using eyeballs and manually create the tracker with CLI. I think we should allow belated trackers.

kingpinXD: Since Athens 3 has been running for a while, do you think those cctxs might have occurred before we made all the improvements to tracker creation ?  
If so this might be related to only athens 3 and not any other environment .

kingpinXD: Is so , we can maybe change the logic to only allow admins to create old trackers , and observers cant

ws4charlie: I think they'll occur due to the way we increase pending nonces (by jumping directly). Currently, zetaclients look back pending cctxs by 100 nonces to see if there are missed pending cctxs, and they post trackers to zetacore.

PR #1432, opened 2023-12-18
chore: merge internal repo changes

lumtis:
## ⚠️ do not `squash and merge` use regular merge to merge the PR

Merge changes from internal repo during code freeze into `node` now that the code freeze has ended.


Dockerfile-versioned-source
```diff
@@ -0,0 +1,72 @@
+FROM golang:1.19-alpine
```
CharlieMc0, line 1: What is the purposes of this dockerfile? And why is it a different version than the others?

lumtis: I created it, it's to perform stateful smoke tests using current source instead of a version, didn't realize Golang version was different

ws4charlie (approved):

CharlieMc0 (approved):

Dockerfile-versioned-source
lumtis: https://github.com/zeta-chain/node/issues/1433

PR #1435, opened 2023-12-19
refactor(observer): put data from observerParams in coreParams

lumtis:
# Description

- Add field from `ObserverParams` into `CoreParams` to keep all the chain parameters at the same place, updatable through a group proposal
- Keep `ObserverParams` as deprecated to simplify migration
- Refactor methods reading observer params to use core params
- Refactor `UpdateCoreParams` to allow adding new core params with new chain ID (previous implementation onoly allowed updating existant core params and therefore required a state upgrade to add a new chain which is not flexible)
- Add `RemoveCoreParams` message to allow officially removing support to a chain
- Add migration script for `ObserverParams` -> `CoreParams`


lumtis:
Closing for https://github.com/zeta-chain/node/pull/1436

PR #1436, opened 2023-12-19
refactor(observer): put data from observerParams in coreParams (cherry picked)

lumtis:
# Description

This is https://github.com/zeta-chain/node/pull/1435 but cherry picked commits to have a proper history

- Add field from `ObserverParams` into `CoreParams` to keep all the chain parameters at the same place, updatable through a group proposal
- Keep `ObserverParams` as deprecated to simplify migration
- Refactor methods reading observer params to use core params
- Refactor `UpdateCoreParams` to allow adding new core params with new chain ID (previous implementation onoly allowed updating existant core params and therefore required a state upgrade to add a new chain which is not flexible)
- Add `RemoveCoreParams` message to allow officially removing support to a chain
- Add migration script for `ObserverParams` -> `CoreParams`


changelog.md
```diff
@@ -70,6 +71,7 @@
 
 * [1391](https://github.com/zeta-chain/node/pull/1391) - consolidate node builds
 * update `MsgUpdateContractBytecode` to use code hash instead of contract address
+>>>>>>> develop
```
brewmaster012, line 74: i think you wanna remove this

proto/observer/params.proto
```diff
@@ -19,11 +20,21 @@ message CoreParams {
...
   int64 outbound_tx_schedule_lookahead = 13;
+  string ballot_threshold = 14 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+  string min_observer_delegation = 15 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+  bool is_supported = 16;
```
brewmaster012, line 33: while you are at it, where is `is_supported` used? This wording kind of bothers me (because if it's not supported why do we have an entry here), maybe something like `is_active` is better?

proto/observer/params.proto
```diff
@@ -19,11 +20,21 @@ message CoreParams {
...
   string connector_contract_address = 9;
   string erc20_custody_contract_address = 10;
-  int64 chain_id = 11;
   int64 outbound_tx_schedule_interval = 12;
   int64 outbound_tx_schedule_lookahead = 13;
+  string ballot_threshold = 14 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+  string min_observer_delegation = 15 [
```
brewmaster012, line 29: Bringing @kingpinXD here. 
Do you think it makes sense to have a per-chain min_observer_delegation? 
If no i'd suggest moving it out of `CoreParams`

brewmaster012 (approved):
Minor comments.

proto/observer/params.proto
lumtis: I think it is good to have a bool we can toggle to quickly stop supporting a chain with a group proposal.
I agree about the renaming into `IsActive` https://github.com/zeta-chain/node/issues/1523

proto/observer/params.proto
lumtis: I think we can create an issue if it's a change that makes sense.

proto/observer/params.proto
```diff
@@ -11,6 +11,7 @@ message CoreParamsList {
   repeated CoreParams core_params = 1;
 }
 message CoreParams {
```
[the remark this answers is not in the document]
lumtis: I was thinking about renaming as well in the initial issue. I can do this in this PR.

changelog.md
lumtis: [d05b9b0](https://github.com/zeta-chain/node/pull/1436/commits/d05b9b0c8c4f42d840f835bb7984384bd6f9db22)

proto/observer/params.proto
kingpinXD: makes sense to remove the per chain , min_delgation amount as we are moving to a unified observer set now

proto/observer/params.proto
```diff
@@ -19,11 +20,21 @@ message CoreParams {
   string zeta_token_contract_address = 8;
   string connector_contract_address = 9;
   string erc20_custody_contract_address = 10;
-  int64 chain_id = 11;
   int64 outbound_tx_schedule_interval = 12;
   int64 outbound_tx_schedule_lookahead = 13;
+  string ballot_threshold = 14 [
```
kingpinXD, line 25: I think ballot threshold , can be moved out this struct as well. Considering its not related to chains at all .
We can consider a mapping between tx types , and ballot_threshold in the future , or just assign thresholds while creating ballots which we are doing now

x/crosschain/client/integrationtests/inbound_voter_test.go
```diff
@@ -268,6 +268,11 @@ func (s *IntegrationTestSuite) TestCCTXInboundVoter() {
 				}
 				signedTx := BuildSignedInboundVote(s.T(), val, s.cfg.BondDenom, account, message)
 				out, err = clitestutil.ExecTestCLICmd(broadcaster.ClientCtx, authcli.GetBroadcastCommand(), []string{signedTx.Name(), "--broadcast-mode", "sync"})
+
+				outStr := out.String()
```
kingpinXD, line 272: do we need the print here?

x/crosschain/client/integrationtests/inbound_voter_test.go
```diff
@@ -268,6 +268,11 @@ func (s *IntegrationTestSuite) TestCCTXInboundVoter() {
...
+				if strings.Contains(outStr, "failed") {
+					fmt.Println(outStr)
+				}
```
kingpinXD, line 275: do we need the print here?

x/observer/keeper/core_params.go
```diff
@@ -0,0 +1,73 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/zeta-chain/zetacore/common"
+	"github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func (k Keeper) SetCoreParamsList(ctx sdk.Context, coreParams types.CoreParamsList) {
+	store := ctx.KVStore(k.storeKey)
+	b := k.cdc.MustMarshal(&coreParams)
+	key := types.KeyPrefix(fmt.Sprintf("%s", types.AllCoreParams))
```
kingpinXD, line 14: Is there a reason that we can't add chain-id to the key ?
That might improve the `GetCoreParamsByChainID` slightly

proto/observer/params.proto
```diff
@@ -11,6 +11,7 @@ message CoreParamsList {
   repeated CoreParams core_params = 1;
```
kingpinXD, line 11: Can we remove this struct all together ?

proto/observer/params.proto
lumtis: @kingpinXD can you create an issue for it?

proto/observer/params.proto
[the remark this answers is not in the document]
lumtis: [6fb3b10](https://github.com/zeta-chain/node/pull/1436/commits/6fb3b10dc33f1c9fb7fe5b210800f7ce3f6afc53)

proto/observer/params.proto
lumtis: That makes sense, I think we can create an issue and group it with the idea of setting a single `min_self_delegation` as well as a general task for observer simplification

x/crosschain/client/integrationtests/inbound_voter_test.go
lumtis: I was debugging stuff I forgot to remove

x/observer/keeper/core_params.go
lumtis: It's based on the current structure right where the params are kept as a unified list, it could be separated with a different prefix but I think it is fine right now, the performance gain would be minimal since the list of core/chain params will stay rather small depending on chain supported. Performance gain would be more if we had +10k values

proto/observer/params.proto
lumtis: What do you mean? Is it related to your comment above?

x/observer/keeper/core_params.go
kingpinXD: I agree the performance gain would be minimal, but I do feel. functions like add /remove or update  would look cleaner .

proto/observer/params.proto
kingpinXD: Yes , resolving this

x/observer/keeper/core_params.go
lumtis: I think it makes sense.
The PR is already big and necessary for v12. I created a separate issue for it: https://github.com/zeta-chain/node/issues/1541

kingpinXD: sounds good

kingpinXD (approved):

PR #1438, opened 2023-12-19
refactor(`smoke-test`): smoke test general improvements: optimize execution, clean up outputs, introduce zetae2e CLI

lumtis:
# Description

### => reduce smoke tests execution from 25 min to around 10 min

- Use concurrency to optimize smoke tests, run general setup, then test types in different Goroutines
- Create `zetae2e` command in `cmd`, this is a copy of the `smoketest` package. `smoketest` is still in the source code, can be removed in a future PR after approval of the changes and for backward compatibility
- Started renaming smoke tests into e2e tests
- Implement a logger to replace fmt.print, allow coloring output based on Goroutine
- Replace all `context.Background` with a initialized parent context
- Add general improvement in smoke tests, like waiting on tx receipt after sending parallel txs
- Fix some race conditions in smoke tests

Note: admin functions (upgrade bytecode, zrc20 pause, etc...) have been removed of the normal tests execution. The issue is that those tests concern specific changes in PR but slow down execution because those can't be run in parallel, we might want another workflow specifically for these tests.

Example execution output:


lumtis:
Opening it for review.

Started working on the command to run smoke tests with external network but this can be completed in another PR. The main change, faster smoke tests execution is ready for review.

The smoke tests seems to fail time to time because a cctx takes too long to be found. I'm still not sure about the reason for this.

.github/workflows/build.yml
```diff
@@ -79,7 +79,7 @@ jobs:
 
   smoke-test:
     runs-on: ["zeta-runners"] # 25 minutes
```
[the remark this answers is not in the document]
lumtis: [4711eae](https://github.com/zeta-chain/node/pull/1438/commits/4711eae1e9f0cfb522424d0feadb574120cbc245)

ws4charlie (approved):

brewmaster012 (approved):
LGTM

.github/workflows/sast-linters.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         with:
           version: v1.54
           skip-cache: true
-          args: --timeout=20m
+          args: --timeout=20m --out-format=json
```
CharlieMc0, line 76: The time out here is longer than the timeout for the job on Line 57. I think we should use the timeout at the job level only and remove it from here.

cmd/zetae2e/config/clients.go
```diff
@@ -0,0 +1,120 @@
...
+	connCfg := &rpcclient.ConnConfig{
+		Host:         rpc,
+		User:         "smoketest",
+		Pass:         "123",
+		HTTPPostMode: true,
+		DisableTLS:   true,
+		Params:       "testnet3",
+	}
+	return rpcclient.New(connCfg, nil)
+}
```
CharlieMc0, line 74: When running the e2e tester on a live network is there a way to pass in a different UN/PW?

lumtis: Not sure what it has been resolved but we can put these values in the config file. I started work on the e2e tester specifically for the EVM now but not yet ready.

CharlieMc0: I see the other issues for them now. https://github.com/zeta-chain/node/issues/1482
Thanks!

Issue #1503, opened 2023-12-27
High CPU/RAM usage by the v11.0.0 node


fadeev:
> After upgrade the node to v11.0.0, there is a issue about node CPU/RAM usage. The node CPU/RAM usage will reach its peak, when the node api QPS is almost 50. Have there been similar cases recently?

> There are some logs about “WARNING: DATA RACE” in my node. Pls help check whether it’s normal.

```
2:22AM INF burned tokens from module account amount=179691726306901010azeta from=evm module=x/bank
2:22AM INF minted coins from module account amount=179691726306901010azeta from=evm module=x/bank
2:22AM INF executed block height=2960111 module=state num_invalid_txs=13 num_valid_txs=33 server=node
2:22AM INF Timed out dur=3000 height=2960111 module=consensus round=0 server=node step=3
2:22AM INF commit synced commit=436F6D6D697449447B5B34342033332035322031353820372031343720313733203233203932203135372031333220323036203133322031353620383520302032313220313532203230372033302033322038332031383220373120363020353120393020313134203234372035312032332039385D3A3244324145467D
2:22AM INF committed state app_hash=2C21349E0793AD175C9D84CE849C5500D498CF1E2053B6473C335A72F7331762 height=2960111 module=state num_txs=46 server=node
2:22AM INF indexed block exents height=2960111 module=txindex server=node
2:22AM INF Timed out dur=65.224756 height=2960112 module=consensus round=0 server=node step=1
2:22AM INF Timed out dur=3000 height=2960112 module=consensus round=0 server=node step=3
2:22AM INF commit is for a block we do not know about; set ProposalBlock=nil commit=F5B32631243CF96CDFB57BB1FDDF9AD5D97CEEED5DFC973A5C9FBFB92224A37B commit_round=1 height=2960112 module=consensus proposal={} server=node
2:22AM INF received complete proposal block hash=F5B32631243CF96CDFB57BB1FDDF9AD5D97CEEED5DFC973A5C9FBFB92224A37B height=2960112 module=consensus server=node
2:22AM INF finalizing commit of block hash={} height=2960112 module=consensus num_txs=13 root=2C21349E0793AD175C9D84CE849C5500D498CF1E2053B6473C335A72F7331762 server=node
2:22AM INF executed block height=2960112 module=state num_invalid_txs=1 num_valid_txs=12 server=node
[...]

PR #1505, opened 2023-12-27
fix: remove race flag from build

lumtis:
# Description

`-race` flag decrease performance and should not be used in production.

Closes: It closes in theory https://github.com/zeta-chain/node/issues/1503 however we might want to let the issue open to check if validators rebuilding the node fix the issue


lumtis:
Duplicate of https://github.com/zeta-chain/node/pull/1504

Issue #1451, opened 2023-12-18
Move inbound/outbound validation logic from `crosschain` to `observer`


lumtis:
As part of the serie of refactoring to simplify `crosschain` and create better separation of concern.

We should move the voting logic to validate inbound/outbound in the `observer` module since this is the module responsible for observer voting. This allows `crosschain` module to be only managing cctx state lifecycle.

In `VoteOnObservedInboundTx` and `VoteOnObservedOutboundTx`, the logic until the following line:
```
	_, isFinalized := k.zetaObserverKeeper.CheckIfFinalizingVote(ctx, ballot)
	if !isFinalized {
		// Return nil here to add vote to ballot and commit state
		return &types.MsgVoteOnObservedInboundTxResponse{}, nil
	}
```
Should be moved to `observer`

We could also adds test for the functionality in the same PR: https://github.com/zeta-chain/node/issues/1316

PR #1511, opened 2024-01-02
refactor(`crosschain`): move ballot voting logic into `observer`

lumtis:
# Description

Move the voting logic for inbound and outbound to the `observer` module and add unit tests to the voting functions.

Closes: https://github.com/zeta-chain/node/issues/1451


brewmaster012:
similar issue:
https://github.com/zeta-chain/node/issues/1515

x/observer/keeper/vote_outbound.go
```diff
@@ -0,0 +1,52 @@
...
+		return false, false, ballot, "", observertypes.ErrInvalidStatus
+	}
+
+	// check if voter is authorized
+	if ok := k.IsAuthorized(ctx, voter); !ok {
+		return false, false, ballot, "", observertypes.ErrNotObserver
+	}
+
+	// fetch or create ballot
+	ballot, isNew, err = k.FindBallot(ctx, ballotIndex, observationChain, observertypes.ObservationType_OutBoundTx)
```
kingpinXD, line 39: Since we are already refactoring some of the ballot logic, it would be good to add the `voter list` and `threshold` as part of the parameter list so that the TSS voter can also use this function.

This can, however, be done as a separate PR, too, if you want to limit the changes on this one.

lumtis: I might think this is a bit related to the first comment? https://github.com/zeta-chain/node/issues/1806

ws4charlie (approved):
looks good

x/observer/keeper/vote_outbound_test.go
```diff
@@ -0,0 +1,193 @@
...
+			observer,
+		)
+		require.NoError(t, err)
+
+		// ballot should be finalized since there is only one observer
+		require.False(t, isFinalized)
+		require.False(t, isNew)
+		expectedBallot, found := k.GetBallot(ctx, "index")
+		require.True(t, found)
+		require.Equal(t, expectedBallot, ballot)
```
kingpinXD, line 191: maybe we can add a test in which the ballot is not finalized , but vote gets added . 
We can set the observer set to a higher number

x/observer/keeper/vote_inbound_test.go
```diff
@@ -0,0 +1,323 @@
...
+			common.CoinType_ERC20,
+			observer,
+			"index",
+			"inTxHash",
+		)
+		require.NoError(t, err)
+
+		// ballot should not be finalized as the threshold is not reached
+		require.False(t, isFinalized)
+		require.False(t, isNew)
```
kingpinXD, line 321: Same comment as the one for outbound TX , 
Maybe we can add unit test which adds vote but doesn't finalize the ballot

brewmaster012 (approved):
LGTM

x/crosschain/keeper/msg_server_vote_inbound_tx.go
```diff
@@ -57,74 +55,60 @@ import (
...
-	}
-
 	index := msg.Digest()
-	// Add votes and Set Ballot
-	// GetBallot checks against the supported chains list before querying for Ballot
-	ballot, isNew, err := k.zetaObserverKeeper.FindBallot(ctx, index, observationChain, observationType)
+
+	// vote on inbound ballot
+	// use a temporary context to not commit any ballot state change in case of error
+	tmpCtx, commit := ctx.CacheContext()
```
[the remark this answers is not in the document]
lumtis: Added it to make https://github.com/zeta-chain/node/blob/cfd27a705ab5d299e8eaa0e078bd6f2360a80ab2/x/crosschain/keeper/msg_server_vote_inbound_tx_test.go#L150 passing.
It is not necessary in practice now since the message would cancel the entire state transition but might become if we use the message internally. So it keep the specs: if the inbound finalized check doesn't pass, no ballot state change

x/observer/keeper/vote_inbound_test.go
lumtis: You mean the opposite? The vote is added to an existing ballot but it is not finalized?

lumtis: Looks good to me to add it. We'll need to add the extensive testing for voting logic here: https://github.com/zeta-chain/node/issues/1806

lumtis: e0395310

lumtis: Oh, I see, ballot created but not finalized? I can add this as well

lumtis: [580e897](https://github.com/zeta-chain/node/pull/1511/commits/580e897f52e5e630746686c06567784f9c00064a)

x/observer/keeper/vote_outbound_test.go
lumtis: [580e897](https://github.com/zeta-chain/node/pull/1511/commits/580e897f52e5e630746686c06567784f9c00064a)

kingpinXD (approved):

PR #1519, opened 2024-01-03
docs: add authorization note to messages

fadeev:
I'm adding a section about admin groups to the docs and I need a way to be able to say "To learn about permissions of different admin policy accounts, refer to the docs on messages" or something. Having an "Authorized" note in the code will hopefully help keep it up to date.

Related: https://github.com/zeta-chain/docs/pull/216


lumtis (commented):
Did you use the doc on Notion as a reference? This looks good to me otherwise.

For maintaining it up-to-date, something I'm thinking could be to maintain a table directly in the codebase that maps message type with authorization, so that possible the docs could be rendered automatically from this table. Could also simplify maintaining authorization and eventually make it upgradable through governance.

Not possible right now because we have messages with a different authorization depending on the action in the message itself but it could be sperated into two messages

lumtis (approved):

brewmaster012 (approved):
LGTM

fadeev:
> Did you use the doc on Notion as a reference? 

@lumtis I didn't 🤦‍♂️

> For maintaining it up-to-date, something I'm thinking could be to maintain a table directly in the codebase that maps message type with authorization, so that possible the docs could be rendered automatically from this table. Could also simplify maintaining authorization and eventually make it upgradable through governance.

I like this!

kingpinXD (approved):

Issue #1515, opened 2024-01-02
separation of voting and execution of external observation


brewmaster012:
Currently, the voting msgs of external observations from observers are tied up with execution. The vote msg that happens to be at the threshold of 2/3 happen to execute the observation and triggers next steps in the cross-chain transaction. 

This model has the following short term and longer term **problems**: 
1. **Short term**: Each voting msg must carry a gas limit that can execute the observation, even though only 1 such msg may actually execute the observation.  This is a problem because executing the observation can cost up to 4.5M gas and this limits the number of observation throughput severely. 

Separating the voting and execution will have the following **additional benefits**
1. Potentially simplifying observability in examining 
2. Make it easier to do observer accounting an dincentives; 
3. Paves way to transition/augmenting observation with merkle proof

**Potential problems** of separating voting and execution: 
1. Changing some of the core logic in the cross-chain transaction and also a bit of zetaclient

**Impact**: 
This change can be made with minimum impact on users, dApps, node operators. It will have some minor impact on explorer and analytics services such as the auditor service.

lumtis:
Does it have a further scope compared to https://github.com/zeta-chain/node/pull/1511?

brewmaster012:
> Does it have a further scope compared to #1511?

Seems the same. The only thing that I see right now is that we'd probably want this in v12. The limitation on observation throughput might be severe.

lumtis:
> > Does it have a further scope compared to #1511?
> 
> Seems the same. The only thing that I see right now is that we'd probably want this in v12. The limitation on observation throughput might be severe.

I added the Epic in v12, this is why the low level tasks were removed https://github.com/zeta-chain/node/issues/1512

> The limitation on observation throughput might be severe.

Not sure to get this, the refactoring in the PR doesn't improve the throughput

brewmaster012:
> > > Does it have a further scope compared to #1511?
> > 
> > 
> > Seems the same. The only thing that I see right now is that we'd probably want this in v12. The limitation on observation throughput might be severe.
> 
> I added the Epic in v12, this is why the low level tasks were removed #1512
> 
> > The limitation on observation throughput might be severe.
> 
> Not sure to get this, the refactoring in the PR doesn't improve the throughput

The idea is that once the separation is done, then observation txs (+ execution msg) consume less gasLimit. 
For example, currently observation inTx has gasLimit 4.5M, block gas limit is 500M
[...]

PR #1536, opened 2024-01-08
fix: inbound index

kingpinXD:
# Description

- Adds an index to keep track of finalized inbounds 
- Adds a new field to the CCTX struct to track the finalization state of a tx 

Closes: (https://github.com/zeta-chain/security/issues/44)

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/crosschain/keeper/finalized_inbounds.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/zeta-chain/zetacore/x/crosschain/types"
+)
+
+func (k Keeper) AddFinalizedInbound(ctx sdk.Context, intxHash string, chainID int64, eventIndex uint64) {
+	finalizedInboundIndex := fmt.Sprintf("%s-%d-%d", intxHash, chainID, eventIndex)
```
lumtis, line 12: Can we create a method in `type` package to compute the key?

lumtis (commented):
It looks good to me.
I'm thinking in anticipation of https://github.com/zeta-chain/node/issues/1515 if we could consider instead of having a boolean `IsFinalized` introduce a `inboundStatus`
So we can have:
- nonFinalized: vote still ongoing
- finalizing: vote finished but not executed yet
- finalized: voted and executed

We could introduce the status value and just keep nonFinalized and finalized for now

kingpinXD:
> It looks good to me. I'm thinking in anticipation of #1515 if we could consider instead of having a boolean `IsFinalized` introduce a `inboundStatus` So we can have:
> 
> * nonFinalized: vote still ongoing
> * finalizing: vote finished but not executed yet
> * finalized: voted and executed
> 
> We could introduce the status value and just keep nonFinalized and finalized for now

It makes sense to do that,however we can consider the following situations 
- if 1515 is going to be part of v12 or the first mainnet release , it would be cleaner to add that as part of the implementation PR 
- if 1515 is going to be implemented post mainnet, we can modify the cross-chain struct now and avoid migrating cctxs on mainnet

x/crosschain/keeper/finalized_inbounds.go
kingpinXD: 9224255304d3b886d646a8ba42fc0c246655e076

brewmaster012:
> [zeta-chain/security#44](https://github.com/zeta-chain/security/issues/44)

What's the relationship to this issue?

lumtis:
To me it seems a correct way to handle it. We can highlight the fix with TrailOfBits in our next call.

x/crosschain/types/keys.go
```diff
@@ -75,6 +76,10 @@ func (m CrossChainTx) LogIdentifierForCCTX() string {
 
 }
 
+func FinalizedInboundKey(intxHash string, chainID int64, eventIndex uint64) string {
+	return fmt.Sprintf("%s-%d-%d", intxHash, chainID, eventIndex)
```
ws4charlie, line 80: Would it be a good idea to reorder the key as `chainID-intxHash-eventIndex`. In future we may have a query to easily collect all finalized inTxs by chain and to know which ones are processed which ones are not.

lumtis: Agree on this!

x/crosschain/keeper/msg_server_vote_inbound_tx.go
```diff
@@ -98,12 +99,16 @@ func (k msgServer) VoteOnObservedInboundTx(goCtx context.Context, msg *types.Msg
...
 
-	_, isFinalized := k.zetaObserverKeeper.CheckIfFinalizingVote(ctx, ballot)
-	if !isFinalized {
+	_, isFinalizedInThisBlock := k.zetaObserverKeeper.CheckIfFinalizingVote(ctx, ballot)
+	if !isFinalizedInThisBlock {
 		// Return nil here to add vote to ballot and commit state
 		return &types.MsgVoteOnObservedInboundTxResponse{}, nil
 	}
 
+	if k.IsFinalizedInbound(ctx, msg.InTxHash, msg.SenderChainId, msg.EventIndex) {
```
lumtis, line 108:
```suggestion
	// Check if the transaction has already been processed
	// Prevent the same transaction to be processed in another vote
	if k.IsFinalizedInbound(ctx, msg.InTxHash, msg.SenderChainId, msg.EventIndex) {
```
Can we add some context on this check?

x/crosschain/keeper/finalized_inbounds.go
```diff
@@ -0,0 +1,19 @@
+package keeper
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/zeta-chain/zetacore/x/crosschain/types"
+)
+
+func (k Keeper) AddFinalizedInbound(ctx sdk.Context, intxHash string, chainID int64, eventIndex uint64) {
```
lumtis, line 9: Can we add unit tests for these?

lumtis, line 9: We should add genesis export/init logic for the entry.

x/crosschain/keeper/msg_server_vote_inbound_tx.go
ws4charlie, line 108: It would be better to place this check on an earlier line (e.g. before line 89 `k.zetaObserverKeeper.FindBallot`), as we don't expect observers resubmit message (with modified fields) to create garbage (new) ballots.

kingpinXD: We still need to allow observers to add votes to ballot even after the tx is finalized

x/crosschain/keeper/msg_server_vote_outbound_tx.go
```diff
@@ -211,6 +211,7 @@ func (k msgServer) VoteOnObservedOutboundTx(goCtx context.Context, msg *types.Ms
 	if err != nil {
 		// do not commit tmpCtx
 		cctx.CctxStatus.ChangeStatus(types.CctxStatus_Aborted, err.Error())
+		cctx.GetCurrentOutTxParam().TxFinalizationStatus = types.TxFinalizationStatus_Executed
```
ws4charlie, line 214: where does this `TxFinalizationStatus_Executed` enum get read (or checked against)? Isn't `CctxStatus_Aborted/CctxStatus_Reverted` enough to tell a cctx's status is finalized

x/crosschain/keeper/msg_server_vote_inbound_tx.go
ws4charlie: Maybe I mean something like this
image

x/crosschain/keeper/finalized_inbounds.go
kingpinXD: a79b057e6f84ef9166ff5b826aebc3267ba0802f

kingpinXD: a79b057e6f84ef9166ff5b826aebc3267ba0802f

x/crosschain/keeper/msg_server_vote_inbound_tx.go
kingpinXD: Yeah, 
I am. I was just saying that this line needs to be executed before the isFinalizedInbound check. 
https://github.com/zeta-chain/zeta-node/blob/a79b057e6f84ef9166ff5b826aebc3267ba0802f/x/crosschain/keeper/msg_server_vote_inbound_tx.go#L97-L100

x/crosschain/keeper/msg_server_vote_outbound_tx.go
kingpinXD: It's not used as of now, because Finalization and Execution are the same process, it will be used when we separate it out in the future

x/crosschain/keeper/msg_server_vote_inbound_tx.go
kingpinXD: fe7339c798cad8727b88aee742df334ddce0f355

kingpinXD: Added a test 
https://github.com/zeta-chain/zeta-node/blob/5547674b6ba79781ff6e479b7671de74b2202050/x/crosschain/keeper/msg_server_vote_inbound_tx_test.go#L149-L150

ws4charlie (approved):

lumtis (approved):

PR #1538, opened 2024-01-08
refactor(`e2e`): improve stateful e2e testing

lumtis:
# Description

- Make stateful e2e testing directly with one Makefile command without manual interaction

Closes:

### DIFF
--- a/contrib/localnet/scripts/genesis-stateful.sh
+++ b/contrib/localnet/scripts/genesis-stateful.sh
@@ -58,7 +58,7 @@ source ~/os-info.sh
 if [ $HOSTNAME != "zetacore0" ]
 then
   echo "Waiting for zetacore0 to create genesis.json"
-  sleep $((7*NUMOFNODES))
+  sleep 10
   echo "genesis.json created"
 fi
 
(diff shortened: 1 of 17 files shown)


ws4charlie (approved):
looks nice

contrib/localnet/scripts/genesis-stateful.sh
```diff
@@ -91,7 +91,7 @@ then
 
 # 2. Add the observers , authorizations and required params to the genesis.json
   zetacored collect-observer-info
-  zetacored add-observer-list
+  zetacored add-observer-list --keygen-block 55
```
kingpinXD, line 94: Can we use a lower height to speed up the process ?

lumtis: I was failing with the previous number, 20.
I reuse the same as one the regular test.

lumtis: 20 seems actually to be fine. Maybe recents change in the tests fix this. Let me do some more testing and maybe change this value everywhere.

kingpinXD (approved):

PR #1555, opened 2024-01-11
fix(`rpc`): reduce websocket message limit to 10MB

lumtis:
# Description

Closes: https://github.com/zeta-chain/security/issues/50


kingpinXD (approved):
LGTM , 
We should consider some testing with the consumers of the websocket though , to see if it affects them by any chance ?

lumtis:
>to see if it affects them by any chance ?
Very unlikely I think, we were using a very high value, 10MB is the regular limit

ws4charlie (approved):
makes sense to me if GoLang accepts 10MB by default on the HTTP service

PR #1556, opened 2024-01-11
fix(`crosschain`): add emptiness check for topic array in event parsing

lumtis:
# Description

Add a check for topic length

Closes: https://github.com/zeta-chain/security/issues/49

### DIFF
--- a/x/crosschain/keeper/evm_hooks.go
+++ b/x/crosschain/keeper/evm_hooks.go
@@ -275,6 +275,9 @@ func (k Keeper) ParseZRC20WithdrawalEvent(ctx sdk.Context, log ethtypes.Log) (*z
 	if err != nil {
 		return nil, err
 	}
+	if len(log.Topics) == 0 {
+		return nil, fmt.Errorf("ParseZRC20WithdrawalEvent: invalid log - no topics")
+	}
 	event, err := zrc20ZEVM.ParseWithdrawal(log)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


kingpinXD (approved):

x/crosschain/keeper/evm_hooks.go
```diff
@@ -275,6 +275,9 @@ func (k Keeper) ParseZRC20WithdrawalEvent(ctx sdk.Context, log ethtypes.Log) (*z
 	if err != nil {
 		return nil, err
 	}
+	if len(log.Topics) == 0 {
```
ws4charlie, line 278: I think we might expect `len(log.Topics) == 2` (signature + from) if you want a more strict check.
https://github.com/code-423n4/2023-11-zetachain/blob/b237708ed5e86f12c4bddabddfd42f001e81941a/repos/protocol-contracts/contracts/zevm/ZRC20.sol#L262

x/crosschain/keeper/evm_hooks.go
```diff
@@ -306,6 +309,9 @@ func ParseZetaSentEvent(log ethtypes.Log, connectorZEVM ethcommon.Address) (*con
...
```
ws4charlie, line 312: For ZRC20 `ZetaSent` event, the number of topics would be `3` (signature + zetaTxSenderAddress + destinationChainId)
https://github.com/code-423n4/2023-11-zetachain/blob/b237708ed5e86f12c4bddabddfd42f001e81941a/repos/protocol-contracts/contracts/zevm/ZetaConnectorZEVM.sol#L67C20-L67C20

lumtis: The fix is just to prevent the panic, I think we already check for the format of the event

ws4charlie (approved):

Issue #1162, opened 2023-09-21
Remove `increaseAllowance`,`decreaseAllowance` checks in EVM hook


lumtis:
Remove the check:
https://github.com/zeta-chain/node/blob/bc0e26d3888088c369c5fa77a9bf8604f825abe7/x/crosschain/keeper/evm_hooks.go#L56

The reason is we will have new patched contracts in mocknet and we have the ability to upgrade bytecodes.

PR #1557, opened 2024-01-11
fix(`crosschain`): remove `decreaseAllowance` and `increaseAllowance` checks

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/1162


brewmaster012 (approved):
If mainnet zrc20 does not have those two faulty functions then this is good to merge.

PR #1588, opened 2024-01-16
fix(`zetaclient`): fix chain params comparison logic

lumtis:
# Description

The chain params were updated every time because the comparison check was not correct.


brewmaster012 (approved):
LGTM

ws4charlie (approved):
lgtm

PR #1605, opened 2024-01-20
fix: exempt system txs from min gas check and fee deduction

brewmaster012:
# Descriptions

The following *System txs* types from observers are now exempt from minimum gas price check
and also will not be charged transaction fees. 

```
MsgGasPriceVoter
MsgVoteOnObservedInboundTx
MsgVoteOnObservedOutboundTx
MsgAddToOutTxTracker
MsgCreateTSSVoter
MsgAddBlockHeader
MsgAddBlameVote
```

Tested in smoketest, by setting gas price in `broadcast.go` to 1 WEI. Smoketest passed despite minimum gas price set
to 1GWEI. 

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/app/ante/ante.go
+++ b/app/ante/ante.go
@@ -17,10 +17,10 @@ package ante
 
 import (
 	"fmt"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
 	"runtime/debug"
 
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
-	ethante "github.com/evmos/ethermint/app/ante"
 	cctxtypes "github.com/zeta-chain/zetacore/x/crosschain/types"
 
 	tmlog "github.com/tendermint/tendermint/libs/log"
@@ -29,9 +29,10 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	errortypes "github.com/cosmos/cosmos-sdk/types/errors"
 	authante "github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 )
 
-func ValidateHandlerOptions(options ethante.HandlerOptions) error {
+func ValidateHandlerOptions(options HandlerOptions) error {
 	if options.AccountKeeper == nil {
 		return errorsmod.Wrap(errortypes.ErrLogic, "account keeper is required for AnteHandler")
 	}
@@ -47,6 +48,9 @@ func ValidateHandlerOptions(options ethante.HandlerOptions) error {
 	if options.EvmKeeper == nil {
 		return errorsmod.Wrap(errortypes.ErrLogic, "evm keeper is required for AnteHandler")
 	}
+	if options.ObserverKeeper == nil {
+		return errorsmod.Wrap(errortypes.ErrLogic, "observer keeper is required for AnteHandler")
+	}
 	return nil
 }
 
@@ -54,7 +58,7 @@ func ValidateHandlerOptions(options ethante.HandlerOptions) error {
 // Ethereum or SDK transaction to an internal ante handler for performing
 // transaction-level processing (e.g. fee payment, signature verification) before
 // being passed onto it's respective handler.
-func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
+func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 	if err := ValidateHandlerOptions(options); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


app/ante/handler_options.go
```diff
@@ -92,19 +114,21 @@ func newCosmosAnteHandler(options ethante.HandlerOptions) sdk.AnteHandler {
...
 
 // this applies to special cosmos tx that calls EVM, in which case the EVM overrides the gas limit
-func newCosmosAnteHandlerNoGasLimit(options ethante.HandlerOptions) sdk.AnteHandler {
+func newCosmosAnteHandlerNoGasFee(options HandlerOptions) sdk.AnteHandler {
 	return sdk.ChainAnteDecorators(
 		ethante.RejectMessagesDecorator{}, // reject MsgEthereumTxs
 		NewAuthzLimiterDecorator(options.DisabledAuthzMsgs...),
 		NewVestingAccountDecorator(),
-		NewSetUpContextDecorator(),
+		ante.NewSetUpContextDecorator(),
```
brewmaster012, line 122: revert back to use stock gas limit checker.

app/ante/ante.go
```diff
@@ -94,28 +98,55 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
 			}
-			if len(tx.GetMsgs()) == 1 && found {
-				// this differs newCosmosAnteHandler only in that it doesn't check gas limit
-				// by using an Infinite Gas Meter.
-				anteHandler = newCosmosAnteHandlerNoGasLimit(options)
-			} else {
-				anteHandler = newCosmosAnteHandler(options)
+
+			isAuthorize := options.ObserverKeeper.IsAuthorized
+			if mm, ok := innerMsg.(*cctxtypes.MsgGasPriceVoter); ok && isAuthorize(ctx, mm.Creator) {
```
brewmaster012, line 117: use the NoFee antehandler only for system txs that actually come from observers.

kingpinXD:
Hi @brewmaster012 I am already working on a branch for this 
`antehandler-system-tx` 

Overall I think instead of making it free we can adjust the gas price based on tx type , so that system tx cost similar to a standard tx . I am using `21000` * CurrentGasPrice as the standard fee , but it can be easily adjusted .
The same gas price adjuster can be called by both the client (to calculate the gas price used in broadcast )and antehandler to calculate the fees 

Checking for observers would still open it up to a dos attack from one of our observers if they try to .

brewmaster012:
> Hi @brewmaster012 I am already working on a branch for this `antehandler-system-tx`

I looked at it, and we need to pick one to work on and get deployed quickly. 
In this PR there are several changes that needs to be done regardless of which one to pick:
1. reverting back to the stock  `ante.NewSetUpContextDecorator()` 
2. should handle both MsgExec wrapped/not wrapped system txs 

I think this PR is a better base, would you like to work on it together?

> 
> Overall I think instead of making it free we can adjust the gas price based on tx type , so that system tx cost similar to a standard tx . I am using `21000` * CurrentGasPrice as the standard fee , but it can be easily adjusted . The same gas price adjuster can be called by both the client (to calculate the gas price used in broadcast )and antehandler to calculate the fees
> 

> Checking for observers would still open it up to a dos attack from one of our observers if they try to .

Yes: how about this: 

allow the system txs to pass with 5% of min gas price. (this roughly fixes the discrapency between cosmos txs vs ethereum txs gas meter for our system txs)

kingpinXD:
> llow the system txs to pass with 1% of min gas price. (this roughly fixes the discrapency between cosmos txs vs ethereum txs gas meter for our system txs).

That makes sense. I can work on this branch as well; 
Applying the same adjustment across all systems tx's can make some TXs substantially cheaper than others .

app/ante/ante.go
```diff
@@ -94,28 +97,48 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
+			} else if mm, ok := innerMsg.(*observertypes.MsgAddBlockHeader); ok && isAuthorize(ctx, mm.Creator) {
+				anteHandler = newCosmosAnteHandlerNoGasFee(options)
+				break
+			} else if mm, ok := innerMsg.(*observertypes.MsgAddBlameVote); ok && isAuthorize(ctx, mm.Creator) {
+				anteHandler = newCosmosAnteHandlerNoGasFee(options)
+				break
+			} else if _, ok := innerMsg.(*stakingtypes.MsgCreateValidator); ok && ctx.BlockHeight() == 0 {
+				anteHandler = newCosmosAnteHandlerNoGasFee(options)
+				break
 			}
```
lumtis, line 140: Would be proper to use a switch statement
```
switch msgType := innerMsg.(type) {
		case *cctxtypes.MsgVoteOnObservedOutboundTx:
		case *cctxtypes.MsgVoteOnObservedInboundTx:
		case ...:
			if isAuthorize(ctx, mm.Creator) {\
				anteHandler = ...
			}
		}
```

app/ante/ante.go
```diff
@@ -94,28 +97,48 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
-					}
+			anteHandler = newCosmosAnteHandler(options)
+			if len(tx.GetMsgs()) != 1 {
+				break
+			}
+
+			msg := tx.GetMsgs()[0] // now we must have len(tx.GetMsgs()) == 1
+			var innerMsg sdk.Msg
+			innerMsg = msg
+			if mm, ok := msg.(*authz.MsgExec); ok { // authz tx; look inside it
```
lumtis, line 108: Need more comment for details on this

app/ante/ante.go
```diff
@@ -94,28 +97,48 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
-					break
-				case *stakingtypes.MsgCreateValidator:
-					if ctx.BlockHeight() == 0 {
-						found = true
-						break
-					}
+			anteHandler = newCosmosAnteHandler(options)
+			if len(tx.GetMsgs()) != 1 {
+				break
+			}
```
lumtis, line 103: Need comment why this break when len  != 1

app/ante/ante.go
```diff
@@ -94,28 +97,48 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
 			}
-			if len(tx.GetMsgs()) == 1 && found {
-				// this differs newCosmosAnteHandler only in that it doesn't check gas limit
-				// by using an Infinite Gas Meter.
-				anteHandler = newCosmosAnteHandlerNoGasLimit(options)
-			} else {
-				anteHandler = newCosmosAnteHandler(options)
+
+			isAuthorize := options.ObserverKeeper.IsAuthorized
+			if mm, ok := innerMsg.(*cctxtypes.MsgGasPriceVoter); ok && isAuthorize(ctx, mm.Creator) {
```
lumtis, line 116: Need more comments on the behavior.
Why these messages are checked, why `isAuthorize` is used

app/ante/ante.go
```diff
@@ -94,28 +97,48 @@ func NewAnteHandler(options ethante.HandlerOptions) (sdk.AnteHandler, error) {
...
 				}
 			}
-			if len(tx.GetMsgs()) == 1 && found {
-				// this differs newCosmosAnteHandler only in that it doesn't check gas limit
-				// by using an Infinite Gas Meter.
-				anteHandler = newCosmosAnteHandlerNoGasLimit(options)
-			} else {
-				anteHandler = newCosmosAnteHandler(options)
+
+			isAuthorize := options.ObserverKeeper.IsAuthorized
```
lumtis, line 115:
```suggestion
			isAuthorized := options.ObserverKeeper.IsAuthorized
```

zetaclient/broadcast.go
```diff
@@ -41,6 +41,7 @@ func (b *ZetaCoreBridge) Broadcast(gaslimit uint64, authzWrappedMsg sdktypes.Msg
 	if baseGasPrice == 0 {
 		baseGasPrice = DefaultBaseGasPrice // shoudn't happen, but just in case
 	}
+	baseGasPrice = int64(float64(baseGasPrice) * 0.011)
```
lumtis, line 44: Why do we do this?
Need comments

app/ante/fees.go
```diff
@@ -99,6 +99,8 @@ func (mpd MinGasPriceDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
 		}
 	}
 
+	minGasPrice = minGasPrice.Mul(sdk.NewDecWithPrec(1, 2)) //
```
lumtis, line 102: Why do we do this? 
Need comments.

lumtis (commented):
- We need to add unit tests for the added functionality.

I think we can implement a separate function taking the message and `isAuthorized` as an argument and returning a bool if the other anteHandler should be used.
And we can add unit test for this function.

app/ante/ante.go
kingpinXD: Why do we need to limit all transactions to single message , can we add this check after we are able to determine that the TX contains a system message .

app/ante/ante.go
kingpinXD: Does this mean only authz-wrapped system messages would be able to get reduced gas prices? 
If, for some reason, an operator wants to submit a TX manually using the operator key, the reduction won't be triggered.

zetaclient/broadcast.go
kingpinXD: We should declare the constants in the common package and used to calculate the fee both here and in the antehandle . `feetx.GetGas()` provides access to the gas limit provided to a tx

kingpinXD (commented):
This looks like great start , 

I do think , if we can take some and implement a function , which ,adjusts the gas prices based on the gas_limit , when the tx matches the following condition
- IsAuthorized 
- IsSystemTx 
would be good .

app/ante/ante.go
lumtis: It just break and use the regular anteHandler. So system message need to be unique in the transaction to use the no gas ante handler, which I think is ok.

app/ante/ante.go
brewmaster012: > Does this mean only authz-wrapped system messages would be able to get reduced gas prices? If, for some reason, an operator wants to submit a TX manually using the operator key, the reduction won't be triggered.

No, 
Look one line above: regardless of whether it's MsgExec wrapped, the innerMsg will point to the unwrapped Msg.

app/ante/ante.go
brewmaster012: not possible, because you would need type assertion to be able to get `mm.Creator`.

lumtis: We can use `GetSigners` from `innerMsg` to get the address of creator

brewmaster012:
> * We need to add unit tests for the added functionality.
> 
> I think we can implement a separate function taking the message and `isAuthorized` as an argument and returning a bool if the other anteHandler should be used. And we can add unit test for this function.

Factored out

zetaclient/broadcast.go
brewmaster012: factored out the contant and use it in both zetacored and zetaclientd

app/ante/ante.go
brewmaster012: ok

brewmaster012: refactored accordingly.

app/ante/ante.go
```diff
@@ -145,3 +145,41 @@ func Recover(logger tmlog.Logger, err *error) {
 		}
 	}
 }
+
+// IsSystemTx determines whether tx is a system tx that's signed by an authorized signer
+// system tx are special types of txs (see in the switch below), or such txs wrapped inside a MsgExec
+// the parameter isAuthorizedSigner is a caller specified function that determines whether the signer of
+// the tx is authorized.
+func IsSystemTx(tx sdk.Tx, isAuthorizedSigner func(string) bool) bool {
```
lumtis, line 153: Can we add unit tests for this method?

brewmaster012: done e032572cc841aa40416a0f9c4a0a7ac7b56d7bfc

lumtis (approved):

kingpinXD (approved):

Issue #1663, opened 2024-01-29
Some blocks produced in Mumbai can't be correctly fetched by ethclient


ws4charlie:
Example problematic Mumbai block: https://mumbai.polygonscan.com/block/44854832
ethclient library has sanity check on block and returned error message `server returned non-empty transaction list but block header indicates no transactions`.

For inbound observation, we can just skip the block if ethclient returns above error because there is no transactions to observe for that block.

Issue #1662, opened 2024-01-29
go-ethereum v1.10.26 can't handle BlobTxType after Goerli Dencun upgrade


ws4charlie:
The go-ethereum version `v1.10.26` cannot fetch block/transaction that contains the new `BlobTxType` introduced in Dencun upgrade. APIs like `BlockByNumber`/ `TransactionByHash` returns error message `transaction type not supported`.
![image](https://github.com/zeta-chain/node/assets/34498985/a2566e83-ddbb-46cc-bf51-c9f306565eda)
![image](https://github.com/zeta-chain/node/assets/34498985/c5fcf7c7-b280-46bd-8680-4424d690c852)

PR #1665, opened 2024-01-29
fix: skip Goerli BlobTxType transaction and Mumbai empty block

ws4charlie:
# Description

Closes: [1663](https://github.com/zeta-chain/node/issues/1663)
Closes: [1662](https://github.com/zeta-chain/node/issues/1662)

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in Athens3
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/tss_signer.go
```diff
@@ -273,14 +273,16 @@ func (tss *TSS) SignBatch(digests [][]byte, height uint64, nonce uint64, chain *
...
-			log.Error().Err(err).Msg("error sending blame data to core")
-			return [][65]byte{}, err
-		}
+		//zetaHash, err := tss.CoreBridge.PostBlameData(&ksRes.Blame, chain.ChainId, index)
+		//if err != nil {
+		//	log.Error().Err(err).Msg("error sending blame data to core")
+		//	return [][65]byte{}, err
+		//}
+		//
+		//log.Info().Msgf("keysign posted blame data tx hash: %s", zetaHash)
```
lumtis, line 285: Can we create an issue to put this back or remove the comment completely if no longer relevant?

zetaclient/evm_client.go
```diff
@@ -92,7 +96,8 @@ type EVMChainClient struct {
 	params                     observertypes.ChainParams
 	ts                         *TelemetryServer
 
-	BlockCache *lru.Cache
+	BlockCache  *lru.Cache
```
kevinssgh, line 99: Do we need to persist these caches to the db? When you restart zetaclient you it will just have to query the block and header again.

zetaclient/tss_signer.go
kevinssgh: Better to have a flag to enable/disable

zetaclient/ethrpc/types.go
```diff
@@ -0,0 +1,162 @@
+package ethrpc
+
+import (
+	"bytes"
+	"math/big"
+	"unsafe"
+)
+
```
kevinssgh, line 8: Do we need to maintain these types or can they be shared with the other rpc client?

zetaclient/ethrpc/types.go
```diff
@@ -0,0 +1,162 @@
+package ethrpc
```
kingpinXD, line 1: Can we consider moving these types to the common folder? 
In case we plan to build other debug tools which would use these types as well. 

Does not have to be done in this PR . just adding an issue would be enough

zetaclient/ethrpc/ethrpc.go
```diff
@@ -0,0 +1,137 @@
+package ethrpc
```
kingpinXD, line 1: Adding to the comment below, I think the entire ethRPC can be moved into common

zetaclient/evm_client.go
```diff
@@ -70,6 +73,7 @@ type EVMChainClient struct {
 	*ChainMetrics
 	chain                      common.Chain
 	evmClient                  EVMRPCClient
+	rpcClient                  *ethrpc.EthRPC // a fallback rpc client
```
kingpinXD, line 76:
```suggestion
	evmAlternateRpcClient                  *ethrpc.EthRPC // a fallback rpc client
```

zetaclient/evm_client.go
```diff
@@ -1411,15 +1476,38 @@ func (ob *EVMChainClient) GetTxID(nonce uint64) string {
...
 	if err != nil {
-		return nil, err
+		if strings.Contains(err.Error(), "block header indicates no transactions") {
+			return nil, nil, false, true, err // it's ok skip empty block
+		} else if strings.Contains(err.Error(), "transaction type not supported") {
+			rpcBlock, err := ob.rpcClient.EthGetBlockByNumber(blockNumber, true)
+			if err != nil {
+				return nil, nil, true, false, err // fall back on ethRPC but still fail
+			}
+			return nil, rpcBlock, true, false, nil // fall back on ethRPC without error
```
kingpinXD, line 1506: Should we also create a cache for the new block type, wdyt ?

zetaclient/evm_client.go
```diff
@@ -756,36 +768,56 @@ func (ob *EVMChainClient) checkConfirmedTx(txHash string, nonce uint64) (*ethtyp
...
+	if err != nil {
+		log.Error().Err(err).Msgf("confirmTxByHash: checkTxInclusion error for txHash %s nonce %d", txHash, nonce)
+		return nil, nil, false
+	}
+
 	return receipt, transaction, true
 }
 
+// checkTxInclusion returns nil only if tx is included in the block at blockNumber and txIndex
+func (ob *EVMChainClient) checkTxInclusion(tx *ethtypes.Transaction, blockNumber uint64, txIndex uint) error {
```
kingpinXD, line 790: Can we define a new type instead of using a bool for fallback?
If so, we can use the switch instead of the `if, else` 
The code would look cleaner IMO , but don't feel too strongly about it

zetaclient/evm_client.go
brewmaster012: we don't need to persist is my understanding

ws4charlie: no need.

zetaclient/ethrpc/ethrpc.go
ws4charlie: Let me make a comment in borrowed files from 3rd-party library.
https://github.com/onrik/ethrpc/blob/main/ethrpc.go

zetaclient/ethrpc/types.go
ws4charlie: let me make a comment for borrowed file
https://github.com/onrik/ethrpc/blob/main/types.go

zetaclient/ethrpc/types.go
ws4charlie: let me make a comment in borrow file
https://github.com/onrik/ethrpc/blob/main/types.go

zetaclient/ethrpc/types.go
ws4charlie: It's a temp solution and be used by zetaclient only

zetaclient/evm_client.go
ws4charlie: do you mean switch on types or something?

zetaclient/ethrpc/types.go
ws4charlie: the [commit](https://github.com/zeta-chain/node/pull/1665/commits/003e851e10e5f76fee146ac76cc4bb564b6c523b)

zetaclient/ethrpc/types.go
ws4charlie: the [commit](https://github.com/zeta-chain/node/pull/1665/commits/003e851e10e5f76fee146ac76cc4bb564b6c523b)

zetaclient/ethrpc/ethrpc.go
ws4charlie: the [commit](https://github.com/zeta-chain/node/pull/1665/commits/003e851e10e5f76fee146ac76cc4bb564b6c523b)

zetaclient/evm_client.go
ws4charlie: the [commit](https://github.com/zeta-chain/node/pull/1665/commits/003e851e10e5f76fee146ac76cc4bb564b6c523b)

zetaclient/evm_client.go
ws4charlie: the [commit](https://github.com/zeta-chain/node/pull/1665/commits/003e851e10e5f76fee146ac76cc4bb564b6c523b)

kevinssgh (approved):

zetaclient/evm_client.go
kingpinXD: I meant something like this 
`type EthRPC int64
	var (
		regular  EthRPC = 1
		fallback EthRPC = 2
	)`

kingpinXD (approved):

Issue #1682, opened 2024-02-01
Prevent gas stability pool mechanism to stop if it fails for one chain


lumtis:
Gas stability pool mechanism should still run for other chains if it fails for a specific chain.

PR #1687, opened 2024-02-01
fix(`crosschain`): only use EVM supported chains for gas stability pool and prevent errors to stop iterating other chains

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/1682


ws4charlie (approved):
LGTM

x/crosschain/keeper/abci.go
```diff
@@ -28,38 +42,59 @@ func (k Keeper) IterateAndUpdateCctxGasPrice(ctx sdk.Context) error {
...
+			// iterate through all pending cctx
+			for _, pendingCctx := range res.CrossChainTx {
+				if pendingCctx != nil {
+					gasPriceIncrease, additionalFees, err := updateFunc(ctx, k, *pendingCctx, gasPriceIncreaseFlags)
+					if err != nil {
+						ctx.Logger().Info("GasStabilityPool: updating gas price for pending cctx failed",
+							"cctxIndex", pendingCctx.Index,
+							"err", err.Error(),
+						)
+						continue IterateChains
```
kingpinXD, line 75: we can probably add a check for the error message and break for cases like " unable to get the median gas price, or withdrawal failed." Because in such situations, the subsequent cctx's would fail.

x/crosschain/keeper/abci.go
```diff
@@ -28,38 +42,59 @@ func (k Keeper) IterateAndUpdateCctxGasPrice(ctx sdk.Context) error {
...
+					}
+					if !gasPriceIncrease.IsNil() && !gasPriceIncrease.IsZero() {
+						ctx.Logger().Info("GasStabilityPool: updated gas price for pending cctx",
```
kingpinXD, line 78: This might cause a lot of log lines, a max of 200 lines every 100 blocks.
Maybe we can aggregate all this data into a struct and emit that in an event?

x/crosschain/keeper/abci.go
lumtis: `continue IterateChains` means all the next pending cctx for the current chains are skipped but we iterate cctxs in the next chain. 
I agree on doing error filtering but I don't see any returned  error in `CheckAndUpdateCctxGasPrice` that would prevent us trying to iterate the cctxs for the next chains.

x/crosschain/keeper/abci.go
lumtis: [7c2084b](https://github.com/zeta-chain/node/pull/1687/commits/7c2084bb0dc1c2aebe605ee6e49439c976190ffe)

kingpinXD (approved):

Issue #1690, opened 2024-02-02
double watched gas price and fix btc scheduler


ws4charlie:

ws4charlie:
https://github.com/zeta-chain/node/pull/1693

PR #1693, opened 2024-02-02
fix: cherry-picked fixes, double watched gas price and fix btc scheduler

ws4charlie:
# Description

Cherry picked `zetaclients` modifications that had been applied to mainnet

Closes: [1690](https://github.com/zeta-chain/node/issues/1690)

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/tx.go
```diff
@@ -51,6 +51,8 @@ func (b *ZetaCoreBridge) WrapMessageWithAuthz(msg sdk.Msg) (sdk.Msg, AuthZSigner
 }
 
 func (b *ZetaCoreBridge) PostGasPrice(chain common.Chain, gasPrice uint64, supply string, blockNum uint64) (string, error) {
+	// double the gas price to avoid gas price spike
+	gasPrice = gasPrice * 2
```
lumtis, line 55: As a long term solution, is it something we could consider as part of the chain params so we can update this value without upgrade.

`GasPriceMultiplier`
Used for chains where the gas price must be unstable or where the gas price increase mechanism can't work.

cc @brewmaster012

lumtis, line 55: We do it for every chains?
I think we should just keep it for Bitcoin once we fix the gas price increase mechanism.

ws4charlie: Yeah. The decision was yesterday to apply this hard coded multiplier `2` to every chain.

ws4charlie: @lumtis maybe we can track it here: https://github.com/zeta-chain/node/issues/1700
This cherry-pick PR is purely to sync `develop` and `mainnet`

kingpinXD: Yeah, for this cherry-pick, this looks fine, but eventually, it might be an excellent idea to mode to chain params; we can always keep the values as 1 if we are not using it.

kingpinXD (approved):

zetaclient/tx.go
lumtis: This is a cherry-pick but we're also fixing the gas stability pool for Ethereum and for BSC the gas price is stable, this will cause unnecessary high fees. I think we can add a condition to only trigger for Bitcoin

Issue #1712, opened 2024-02-06
Increase outTx inclusion timeout to 20 minutes


ws4charlie:
We've decided to increase EVM outtx inclusion timeout from `10 mins` to `20 mins` for mainnet to mitigate missing outtx tracker issue.

lumtis:
Mainnet label was used for issue for mainnet launch, we can remove this label. We should consider that these changes in the codebase are for all networks

Issue #1707, opened 2024-02-05
'EstimateSmartFee' returning very high fee rate compared to 3rd party estimators, for instance the https://mempool.space/


ws4charlie:
`EstimateSmartFee` returning very high fee rate compared to 3rd estimators, for instance the https://mempool.space . 
Additionally, doubling watched gas price makes TSS spend even more gas fee in Bitcoin outbound transactions.

PR #1713, opened 2024-02-06
fix: fix bitcoin fee rate and increase evm outtx inclusion timeout

ws4charlie:
# Description
The economical fee rate `sometimes` can be half of the conservative fee rate (forgot about taking a screenshot for that. There other day was around `49sats/vB` vs. `26 sats/vb`).

image

monitored a bit more
image
image

blocks like 829506, 829530, the `Fee Span` range was quite wide and `Conservative` gas rate was `3~4` times of
`Economical` rate.
image
image

Closes: [1707](https://github.com/zeta-chain/node/issues/1707)
Closes: [1712](https://github.com/zeta-chain/node/issues/1712)

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/bitcoin_client_test.go
```diff
@@ -158,6 +175,63 @@ func (suite *BitcoinClientTestSuite) Test3() {
...
+func TestBitcoinFeeRate(t *testing.T) {
+	// mainnet config
+	connCfg := &rpcclient.ConnConfig{
+		Host:         "127.0.0.1:8332",
+		User:         "username",
+		Pass:         "password",
+		Params:       "mainnet",
+		HTTPPostMode: true,
+		DisableTLS:   true,
+	}
```
lumtis, line 192: We can't use it for unit test
What about creating a mock for the client?
It doesn't seem we need anything more than a mock for `EstimateSmartFee`

ws4charlie: [commit](https://github.com/zeta-chain/node/pull/1713/commits/7ebaf3805a1dd7284917117e7e9b2f009c2fe037)

lumtis: Can we add a comment how this test is supposed to be run? We might consider adding this to the E2E test lists?

lumtis (approved):
LGTM, just one comment on the test

zetaclient/bitcoin_client_test.go
ws4charlie: I tried it on the Bitcoin testnet. The API returned me same number regardless of what `confTarget` `estimateMode` passed. What we expect in the test case is that the higher the `confTarget` passed, the lower the rate; and `Conservative` must give a higher gas rate than `Economical` if same `confTarget` is given. These two assertions will break in Testnet (I guess private network will also break the assertions).
image

ws4charlie:
> LGTM, just one comment on the test

The way I used to run `LiveTestSomeTest` is to remove the prefix `Live` of the function (`TestBitcoinFeeRate` in this case) to make it an executable unit test. It's not an ideal way TBH.

brewmaster012 (approved):

Issue #1716, opened 2024-02-06
Cleaning up `x/{module_name}/types/expected_keepers.go`


lumtis:
Most of the modules expected keeper interfaces exposes more methods than necessary. We should remove all non-necessary methods to make clear what is the interface between modules.

PR #1754, opened 2024-02-13
chore: cleanup expected keepers

skosito:
# Description

Cleaned up unused methods from expected keepers interfaces. Also executed `make mocks` to remove not needed mocks.

One small inconsistency I noticed is that cosmos-sdk modules keepers (for example bank, account keeper etc) are sometimes added as dependency through `NewAppModule`, and sometimes when module keeper is initialized with `NewXKeeper`, not sure if there is some particular reason for this, but i tried to clean that up too.

Closes: #1716 

## Type of change

Cleanup

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -48,6 +48,7 @@
 ### Chores
 
 * [1729](https://github.com/zeta-chain/node/pull/1729) - add issue templates
+* [1754](https://github.com/zeta-chain/node/pull/1754) - cleanup expected keepers
 
 ## Version: v12.2.4
 


app/app.go
```diff
@@ -474,10 +475,10 @@ func New(
 		groupmodule.NewAppModule(appCodec, app.GroupKeeper, app.AccountKeeper, app.BankKeeper, interfaceRegistry),
 		evm.NewAppModule(app.EvmKeeper, app.AccountKeeper, evmSs),
 		feemarket.NewAppModule(app.FeeMarketKeeper, feeSs),
-		crosschainmodule.NewAppModule(appCodec, app.ZetaCoreKeeper, app.StakingKeeper, app.AccountKeeper),
-		observermodule.NewAppModule(appCodec, *app.ZetaObserverKeeper, app.AccountKeeper, app.BankKeeper),
-		fungibleModule.NewAppModule(appCodec, app.FungibleKeeper, app.AccountKeeper, app.BankKeeper),
-		emissionsModule.NewAppModule(appCodec, app.EmissionsKeeper, app.AccountKeeper),
+		crosschainmodule.NewAppModule(appCodec, app.ZetaCoreKeeper),
```
kingpinXD, line 478: Not required, it might be a good idea to rename `app.ZetaCoreKeeper` to `app.CrosschainKeeper`

kingpinXD (approved):
lgtm , left a  minor comment

app/app.go
lumtis: Yes, this is correct, we should remove all Zeta prefix and keep `<ModuleName>Keeper` for the convention. For the pointer, I think we can handle it in another PR

x/fungible/types/expected_keepers.go
```diff
@@ -1,12 +1,11 @@
 package types
 
 import (
-	"context"
+	context "context"
```
lumtis, line 4:
```suggestion
	"context"
```
I don't think it is necessary?

lumtis (approved):
Looks good! one minor comment

x/fungible/types/expected_keepers.go
lumtis: Goland? It is sometimes doing weird things to me as well sometime.

kingpinXD (approved):
you would need to add a line in the changelog.mg to get the changelog check to pass

PR #1758, opened 2024-02-14
fix: add error check in btc decode

kingpinXD:
# Description

- Adds error check for in BTC address decode 
- Logs events in cases where an event is created but cannot be parsed 

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/crosschain/keeper/evm_hooks.go
```diff
@@ -55,6 +55,7 @@ func (k Keeper) PostTxProcessing(
 // ProcessLogs post-processes logs emitted by a zEVM contract; if the log contains Withdrawal event
 // from registered ZRC20 contract, new CCTX will be created to trigger and track outbound
 // transaction.
+// TODO : https://github.com/zeta-chain/node/issues/1759
```
lumtis, line 58:
```suggestion
// TODO: implement unit tests
// https://github.com/zeta-chain/node/issues/1759
```
Better to add description for each TODO

x/crosschain/keeper/evm_hooks.go
```diff
@@ -72,12 +73,25 @@ func (k Keeper) ProcessLogs(ctx sdk.Context, logs []*ethtypes.Log, emittingContr
 				return err
 			}
 		}
+		// We were able to parse the ZRC20 withdrawal event. However, we were unable to process it as the information was incorrect
+		// This means that there are some funds locked in the contract which cannot have an outbound , this can be a candidate for a refund
+		// TODO : Consider returning error or auto refunding the funds to the user
+		if err != nil && eventWithdrawal != nil {
```
lumtis, line 79:
```suggestion
		} else if eventWithdrawal != nil {
		// We were able to parse the ZRC20 withdrawal event. However, we were unable to process it as the information was incorrect
		// This means that there are some funds locked in the contract which cannot have an outbound , this can be a candidate for a refund
		// TODO : Consider returning error or auto refunding the funds to the user
```
Would be clearer, `err != nil` is already true if `err == nil` doesn't check

x/crosschain/keeper/evm_hooks.go
```diff
@@ -72,12 +73,25 @@ func (k Keeper) ProcessLogs(ctx sdk.Context, logs []*ethtypes.Log, emittingContr
...
+		}
+
 		eZeta, err := ParseZetaSentEvent(*log, connectorZEVMAddr)
 		if err == nil {
 			if err := k.ProcessZetaSentEvent(ctx, eZeta, emittingContract, txOrigin); err != nil {
 				return err
 			}
 		}
+		// We were able to parse the ZetaSent event. However, we were unable to process it as the information was incorrect
+		if err != nil && eZeta != nil {
```
lumtis, line 91:
```suggestion
		} else if eZeta != nil {
		// We were able to parse the ZetaSent event. However, we were unable to process it as the information was incorrect
```

x/crosschain/keeper/evm_hooks.go
kingpinXD: 5e5e25dc7bdcf78eee2a5ea5fe09160251a0188e

x/crosschain/keeper/evm_hooks.go
kingpinXD: 5e5e25dc7bdcf78eee2a5ea5fe09160251a0188e

x/crosschain/keeper/evm_hooks.go
lumtis: Can
```
if err := k.ProcessZRC20WithdrawalEvent(ctx, eventWithdrawal, emittingContract, txOrigin); err != nil {
	return err
}
```
modify `eventWithdrawal`? If not I still think we should use a `else` to make the flow of the program more clear

lumtis (commented):
Adding one more comment regarding the `else` otherwise looks good

x/crosschain/keeper/evm_hooks.go
lumtis: same comment as above

common/address.go
```diff
@@ -72,6 +72,9 @@ func DecodeBtcAddress(inputAddress string, chainID int64) (address btcutil.Addre
 		return nil, fmt.Errorf("chain params not found")
 	}
 	address, err = btcutil.DecodeAddress(inputAddress, chainParams)
+	if err != nil {
+		return nil, fmt.Errorf("decode address failed: %s", err.Error())
```
ws4charlie, line 76: should we include the `inputAddress` in error message?

x/crosschain/keeper/evm_hooks.go
```diff
@@ -72,12 +74,25 @@ func (k Keeper) ProcessLogs(ctx sdk.Context, logs []*ethtypes.Log, emittingContr
 				return err
 			}
 		}
+		// We were able to parse the ZRC20 withdrawal event. However, we were unable to process it as the information was incorrect
+		// This means that there are some funds locked in the contract which cannot have an outbound , this can be a candidate for a refund
+		// TODO : Consider returning error or auto refunding the funds to the user
+		if eventWithdrawal != nil {
```
ws4charlie, line 80: shouldn't it be an `else if eventWithdrawal != nil {` ?

x/crosschain/keeper/evm_hooks.go
```diff
@@ -72,12 +74,25 @@ func (k Keeper) ProcessLogs(ctx sdk.Context, logs []*ethtypes.Log, emittingContr
...
+		}
+
 		eZeta, err := ParseZetaSentEvent(*log, connectorZEVMAddr)
 		if err == nil {
 			if err := k.ProcessZetaSentEvent(ctx, eZeta, emittingContract, txOrigin); err != nil {
 				return err
 			}
 		}
+		// We were able to parse the ZetaSent event. However, we were unable to process it as the information was incorrect
+		if eZeta != nil {
```
ws4charlie, line 92: same else if?

x/crosschain/keeper/evm_hooks.go
```diff
@@ -293,7 +310,7 @@ func ParseZetaSentEvent(log ethtypes.Log, connectorZEVM ethcommon.Address) (*con
 	}
 
 	if event.Raw.Address != connectorZEVM {
-		return nil, fmt.Errorf("ParseZetaSentEvent: event address %s does not match connectorZEVM %s", event.Raw.Address.Hex(), connectorZEVM.Hex())
+		return event, fmt.Errorf("ParseZetaSentEvent: event address %s does not match connectorZEVM %s", event.Raw.Address.Hex(), connectorZEVM.Hex())
```
ws4charlie, line 313: I think it's okay to return `nil` if you don't want to print too many logs of `Error processing Zeta Sent event , from Address ...` from the caller's side. It's an irreverent `ZetaSent` event emitted from non-connector contract.

x/crosschain/keeper/evm_hooks.go
kingpinXD: aaea061766d4b155b32783fcb417ae9388328e95

x/crosschain/keeper/evm_hooks.go
kingpinXD: aaea061766d4b155b32783fcb417ae9388328e95

x/crosschain/keeper/evm_hooks.go
kingpinXD: Remove the check, returning nil for the address mismatch event. This check is useless if we do that

x/crosschain/keeper/evm_hooks.go
kingpinXD: aaea061766d4b155b32783fcb417ae9388328e95

x/crosschain/keeper/evm_hooks.go
brewmaster012: We want to error out when the Withdrawal event needs to be processed, but the event contains invalid input from user such as unsupported address. 

Currently k.ParseZRC20WithdrawalEvent() returns non-nil error when the Withdrawal event needs to be processed **and** is correct. 

We need to have a condition here that says the Withdrawal event needs to be processed and it needs to be correct, otherwise error out so the EVM tx will fail. 

Thinking either `k.ParseZRC20WithdrawalEvent` returns an additional field that indicates this is a relevant Withdrawal event (the criteria is--the emitting contract is one of the ZRC20s in the foreign coin list).

brewmaster012 (changes requested):

x/crosschain/keeper/evm_hooks_test.go
```diff
@@ -0,0 +1,140 @@
...
...04a397460ee1dbe8c9503e098090567d6b7f4b3c02a8617d800b6d955\",\"0x00000000000000000000000033ead83db0d0c682b05ead61e8d8f481bb1b4933\"],\"data\":\"0x000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000030390000000000000000000000000000000000000000000000000000000000003d840000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002a626331717973643473703971386d793539756c3977736635727673397033383768663876667761747a7500000000000000000000000000000000000000000000\",\"blockNumber\":\"0x1a00f3\",\"transactionHash\":\"0x9aaefece38fd2bd87077038a63fffb7c84cc8dd1ed01de134a8504a1f9a410c3\",\"transactionIndex\":\"0x8\",\"blockHash\":\"0x9517356f0b3877990590421266f02a4ff349b7476010ee34dd5f0dfc85c2684f\",\"logIndex\":\"0x2b\",\"removed\":false}],\"transactionHash\":\"0x9aaefece38fd2bd87077038a63fffb7c84cc8dd1ed01de134a8504a1f9a410c3\",\"contractAddress\":\"0x0000000000000000000000000000000000000000\",\"gasUsed\":\"0x12575\",\"blockHash\":\"0x9517356f0b3877990590421266f02a4ff349b7476010ee34dd5f0dfc85c2684f\",\"blockNumber\":\"0x1a00f3\",\"transactionIndex\":\"0x8\"}\n"
```
kevinssgh, line 139: Would it be easier to put these strings in a file?

ws4charlie (approved):

x/crosschain/keeper/evm_hooks_test.go
ws4charlie: test can be improved in future maybe

PR #1767, opened 2024-02-16
test: emissions unit test

kingpinXD:
# Description

- Add unit tests for emissions module
Closes: https://github.com/zeta-chain/node/issues/1659

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/emissions/abci_test.go
```diff
@@ -1,229 +1,180 @@
...
+	"github.com/zeta-chain/zetacore/cmd/zetacored/config"
+	"github.com/zeta-chain/zetacore/common"
+	keepertest "github.com/zeta-chain/zetacore/testutil/keeper"
+	"github.com/zeta-chain/zetacore/testutil/sample"
+	emissionsModule "github.com/zeta-chain/zetacore/x/emissions"
+	emissionstypes "github.com/zeta-chain/zetacore/x/emissions/types"
+	observerTypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func TestBeginBlocker(t *testing.T) {
```
lumtis, line 20: Can we add a positive case test?
The emissions pool is decreasing and withdrawable emissions/validator rewards are emitted.

x/emissions/abci_test.go
```diff
@@ -1,229 +1,180 @@
...
+	}
+	zk.ObserverKeeper.SetBallotList(ctx, &observerTypes.BallotListForHeight{
+		Height:           0,
+		BallotsIndexList: ballotIdentifiers,
+	})
+
+	// Total block rewards is the fixed amount of rewards that are distributed
+	totalBlockRewards, err := common.GetAzetaDecFromAmountInZeta(emissionstypes.BlockRewardsInZeta)
+	totalRewardCoins := sdk.NewCoins(sdk.NewCoin(config.BaseDenom, totalBlockRewards.TruncateInt()))
+	assert.NoError(t, err)
```
lumtis, line 95: Any reason we need assert here?

x/emissions/abci_test.go
```diff
@@ -1,229 +1,180 @@
...
+		for i := 0; i < 100; i++ {
+			// produce a block
+			emissionsModule.BeginBlocker(ctx, *k)
+			ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 1)
+		}
+		require.True(t, sk.BankKeeper.GetBalance(ctx, feeCollectorAddress, config.BaseDenom).Amount.IsZero())
+		require.True(t, sk.BankKeeper.GetBalance(ctx, emissionstypes.EmissionsModuleAddress, config.BaseDenom).Amount.Equal(sdk.NewInt(1000000000000)))
+	})
+}
+func TestObserverRewards(t *testing.T) {
```
lumtis, line 74: I think it would be better to have `TestDistributeObserverRewards` that directly test with the `DistributeObserverRewards` method to test rewards distribution.

`BeginBlocker` should test that `DistributeObserverRewards` is correctly called but the details and edge cases for observer distribution should be tested with `DistributeObserverRewards` directly

x/emissions/abci_test.go
```diff
@@ -1,229 +1,180 @@
...
+		// Assert we are always under the max limit of block rewards
+		assert.True(t, totalRewardCoins.AmountOf(config.BaseDenom).
+			Sub(totalDistributedTillCurrentBlock).GTE(sdk.ZeroInt()))
+
+		ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 1)
+	}
+
+	// We can simplify the calculation as the rewards are distributed equally among all the observers
+	rewardPerUnit := observerRewardsForABlock.Quo(sdk.NewInt(int64(len(ballotList) * len(observerSet.ObserverList))))
+	emissionAmount := rewardPerUnit.Mul(sdk.NewInt(int64(len(ballotList))))
```
lumtis, line 135: The problem is we replicate logic of the function to test in the test which assume that this logic is correct.

It would be beter to directly define and check [input/expected output] with simple values we know are correct and check with actual output.

I think the best structure for tests for a distribution logic would be to use the Table Driven Test paradigm:
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

kingpinXD: This test tries to ensure that the observers receive the balance they should have . The distribution
logic itself already has some table-driven tests for the `BuildRewardsDistribution` function.

With that said, I see some value in adding tests to calculate the reward amount ,from the distribution map . I would not want to replace this test , and add that in separately

ws4charlie (approved):

x/emissions/abci_test.go
lumtis: >The distribution
logic itself already has some table-driven tests for the BuildRewardsDistribution function.

What do you mean by this? Any link?

kingpinXD: I was referring to this test 
https://github.com/zeta-chain/zeta-node/blob/0c9ca3df5d30a6ca4b661eeac5fcd000e45d6e9c/x/observer/types/ballot_test.go#L347-L347

lumtis (approved):

Issue #1704, opened 2024-02-05
Add code coverage diff as part of the CI


lumtis:
Add a code coverage diff action giving the difference between commits.
Make the CI fails for a certain threshold

lumtis:
Codecov looks like a good solution for this

PR #1781, opened 2024-02-19
ci: add codecov coverage upload as part of the CI

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/1704


ws4charlie:
Does codecov give combined coverage of both e2e tests + unit tests or just unit tests? zetaclient don't have enough unit tests to run at the moment.

ws4charlie (approved):

lumtis:
> lgtm, just question how is 60% selected to threshold?

Just a value taken for now based on our OKR.

> not relevant to this PR, but maybe we can consider using -mode=count heat map flag to get more insight, especially when developing tests locally, which should give how many is each statement covered, i think it should mostly be visible in html reports, but might not be needed

Good suggestion, we can think about it in the future

> Does codecov give combined coverage of both e2e tests + unit tests or just unit tests? zetaclient don't have enough unit tests to run (don't know the current coverage yet) at the moment and can lead to CI failure.

Coverage is for unit test only. For the threshold, it will not make the PR failing if it is not reached.

Issue #1770, opened 2024-02-16
evm signer - simplify TryProcessOutTx, long function


kevinssgh:
Implementation of TryProcessOutTx in the evm signer is quite long and can be broken into logical steps. Needs to be separated into several smaller functions.

PR #1809, opened 2024-02-26
refactor: evm signer - tryprocessout

kevinssgh:
# Description

Refactored the tryprocessout interface int the evm signer. The function was broken into 3 parts:
* Input setup
* Transaction signing
* Broadcasting

A new data structure was introduced to store the required inputs for all transaction types. Signatures were simplified using the new struct. 

Some input validation and setup was moved to another function as well.  

Closes: https://github.com/zeta-chain/node/issues/1770

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/evm/evm_signer.go
```diff
@@ -324,246 +332,233 @@ func (signer *Signer) SignCommandTx(
...
-	logger.Info().Msgf("EVM Chain TryProcessOutTx: %s, value %d to %s", cctx.Index, cctx.GetCurrentOutTxParam().Amount.BigInt(), cctx.GetCurrentOutTxParam().Receiver)
-
-	defer func() {
-		outTxMan.EndTryProcess(outTxID)
-	}()
-	myID := zetaBridge.GetKeys().GetOperatorAddress()
-
-	var to ethcommon.Address
-	var toChain *common.Chain
+func SetChainAndSender(cctx *types.CrossChainTx, logger zerolog.Logger, txData *TransactionData) bool {
```
lumtis, line 338: I think we can use `txData` as a receiver for the method to make things clearer
```
func (txData *TransactionData) SetChainAndSender(cctx *types.CrossChainTx, logger zerolog.Logger)
```
It is also unclear what the bool represent

zetaclient/evm/evm_signer.go
```diff
@@ -324,246 +332,233 @@ func (signer *Signer) SignCommandTx(
...
+	txData.gasLimit = cctx.GetCurrentOutTxParam().OutboundTxGasLimit
+	if txData.gasLimit < 100_000 {
+		txData.gasLimit = 100_000
+		logger.Warn().Msgf("gasLimit %d is too low; set to %d", cctx.GetCurrentOutTxParam().OutboundTxGasLimit, txData.gasLimit)
 	}
-	if gasLimit > 1_000_000 {
-		gasLimit = 1_000_000
-		logger.Warn().Msgf("gasLimit %d is too high; set to %d", cctx.GetCurrentOutTxParam().OutboundTxGasLimit, gasLimit)
+	if txData.gasLimit > 1_000_000 {
+		txData.gasLimit = 1_000_000
```
lumtis, line 366: We should define constants for these

zetaclient/evm/evm_signer.go
```diff
@@ -324,246 +332,233 @@ func (signer *Signer) SignCommandTx(
...
+			return fmt.Errorf("cannot convert gas price  %s ", cctx.GetCurrentOutTxParam().OutboundTxGasPrice)
 		}
 	} else {
-		gasprice = specified
+		txData.gasPrice = specified
 	}
+	return nil
+}
+
+func SetTransactionData(
```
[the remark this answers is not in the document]
kevinssgh: The function is doing some validation as well as setting up the data structure. NewTransationData would imply a simple constructor.

zetaclient/evm/evm_signer.go
```diff
@@ -324,246 +332,233 @@ func (signer *Signer) SignCommandTx(
...
+	if skipTx {
+		return true, nil
+	}
+
+	toChain := common.GetChainFromChainID(txData.toChainID.Int64())
+	if toChain == nil {
+		return true, fmt.Errorf("unknown chain: %d", txData.toChainID.Int64())
+	}
+
+	// Get nonce, Early return if the cctx is already processed
```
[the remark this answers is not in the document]
kevinssgh: not sure why this would happen, but the check was here before. If the cctx was processed already it shouldn't be in the pending list but there may be some timing issues where the cctx wasn't updated but the tx was confirmed. I'm going to leave it for now.

zetaclient/evm/evm_signer.go
```diff
@@ -324,246 +332,233 @@ func (signer *Signer) SignCommandTx(
...
-			sendhash,
-			cctx.GetCurrentOutTxParam().OutboundTxTssNonce,
-			gasprice,
-			height,
-		)
+		logger.Info().Msgf("SignOutboundTx: %d => %s, nonce %d, gasPrice %d", cctx.InboundTxParams.SenderChainId, toChain, cctx.GetCurrentOutTxParam().OutboundTxTssNonce, txData.gasPrice)
+		tx, err = signer.SignOutboundTx(&txData)
 	}
 
 	if err != nil {
```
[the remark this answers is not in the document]
kevinssgh: The else if ensures we only execute one code block, if we remove the else if's it will proceed to check all the other conditions and possibly sign the transaction again. The assumption above works only if the signature fails.

I can add the return err in each if block but like you said it will make the function longer. Either way I think there is a better solution, we can have a router of some kind to replace the else if's but it will take some time to think about the implementation.

zetaclient/evm/evm_signer.go
```diff
@@ -32,6 +37,26 @@ import (
...
+	gasLimit   uint64
+	message    []byte
+	sendHash   [32]byte
+	nonce      uint64
+	height     uint64
+
+	cmd            string
+	params         string
+	outboundParams *types.OutboundTxParams
+	flags          observertypes.CrosschainFlags
```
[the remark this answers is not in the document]
kevinssgh: Yes, I noticed this as well as I was reviewing again. It doesn't need to be passed to any of the signers. I removed it and queried it from zetabridge directly. We can create another issue for this because it might be out of the scope of this PR.

lumtis:
@ws4charlie can you please have a look in this PR sometime today?

zetaclient/evm/evm_signer.go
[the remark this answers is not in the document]
lumtis: You can have preprocessing or validation in a construction. Referring to the code it seems to be the actually entrypoint to initialize the structure:
```
	// Setup Transaction input
	txData := BaseTransactionData{}
	txData.height = height
	skipTx, err := txData.SetTransactionData(cctx, evmClient, signer.client, logger)
```
It would be better to make clear it's the entrypoint to initialize the structure with a new method that return the object.
This makes code maintainability easier and also mutable function are harder to debug

zetaclient/evm/evm_signer.go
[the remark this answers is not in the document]
lumtis: Not handing the error right after it is returned makes code maintainability harder. Having a longer function is less a concern. We can create new low-level methods to process the different cases.

zetaclient/evm/evm_signer.go
```diff
@@ -347,236 +332,109 @@ func (signer *Signer) TryProcessOutTx(
...
-		} else {
-			logger.Info().Msgf("please wait for pending outTx %s nonce %d to be included", pendingTx.Hash().Hex(), nonce)
-			return
-		}
-	}
+	// Get destination chain for logging
+	toChain := common.GetChainFromChainID(txData.toChainID.Int64())
 
-	flags, err := zetaBridge.GetCrosschainFlags()
+	// Get cross-chain flags
```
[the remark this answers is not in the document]
kevinssgh: created issue: https://github.com/zeta-chain/node/issues/1832

zetaclient/evm/evm_signer.go
```diff
@@ -275,19 +263,16 @@ func (signer *Signer) SignWithdrawTx(
...
-) (*ethtypes.Transaction, error) {
-	if cmd == common.CmdWhitelistERC20 {
-		erc20 := ethcommon.HexToAddress(params)
+// SignCommandTx signs a transaction based on the given command includes:
+//
+//	cmd_whitelist_erc20
+//	cmd_migrate_tss_funds
+func (signer *Signer) SignCommandTx(txData *BaseTransactionData) (*ethtypes.Transaction, error) {
+	outboundParams := txData.outboundParams
+	if txData.cmd == common.CmdWhitelistERC20 {
```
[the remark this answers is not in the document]
kevinssgh: I added the switch case, and I'm assuming you mean to move the logic in separate functions.

zetaclient/evm/evm_signer.go
```diff
@@ -297,16 +282,16 @@ func (signer *Signer) SignCommandTx(
...
 		if err != nil {
 			return nil, fmt.Errorf("sign error: %w", err)
 		}
 		return tx, nil
 	}
-	if cmd == common.CmdMigrateTssFunds {
-		tx := ethtypes.NewTransaction(outboundParams.OutboundTxTssNonce, to, outboundParams.Amount.BigInt(), outboundParams.OutboundTxGasLimit, gasPrice, nil)
+	if txData.cmd == common.CmdMigrateTssFunds {
+		tx := ethtypes.NewTransaction(outboundParams.OutboundTxTssNonce, txData.to, outboundParams.Amount.BigInt(), outboundParams.OutboundTxGasLimit, txData.gasPrice, nil)
 		hashBytes := signer.ethSigner.Hash(tx).Bytes()
```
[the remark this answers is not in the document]
kevinssgh: It is a bit strange, the Migrate TSS funds is creating and signing the tx directly in the logic instead of calling the sign function. I think that the Migrate command can also just call the higher level sign function instead of signing directly with the tssSigner.

zetaclient/evm/evm_signer.go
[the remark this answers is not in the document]
kevinssgh: I'm going to just return the error for each condition since this PR is getting a bit long. I will open another PR to clean this part up some more.

zetaclient/evm/base_transaction_data.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return false
+}
+
+// SetupGas sets the gas limit and price
+func (txData *BaseTransactionData) SetupGas(
+	cctx *types.CrossChainTx,
+	logger zerolog.Logger,
+	client interfaces.EVMRPCClient,
+	chain *common.Chain,
```
ws4charlie, line 69: I think we can just use a non-pointer parameter here?

zetaclient/evm/base_transaction_data.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const (
+	MinGasLimit = 100_000
+	MaxGasLimit = 1_000_000
+)
+
+// BaseTransactionData is a data structure containing input fields used to construct each type of transaction.
+// This is populated using cctx and other input parameters passed to TryProcessOutTx
+type BaseTransactionData struct {
```
ws4charlie, line 25: Instead of using `BaseTransactionData`, should we consider a more meaningful name, e.g. `OutboundKeysignData` or `OutboundTransactionData` to indicate the purpose of this data structure?

zetaclient/evm/base_transaction_data_test.go
```diff
@@ -0,0 +1,85 @@
...
+		chain := corecommon.BscMainnetChain()
+		err := txData.SetupGas(cctx, logger, evmSigner.EvmClient(), &chain)
+		require.NoError(t, err)
+	})
+
+	t.Run("SetupGas_error", func(t *testing.T) {
+		cctx.GetCurrentOutTxParam().OutboundTxGasPrice = "invalidGasPrice"
+		chain := corecommon.BtcMainnetChain()
+		err := txData.SetupGas(cctx, logger, evmSigner.EvmClient(), &chain)
+		require.Error(t, err)
```
ws4charlie, line 67: I have been curious whether the dynamic gas price special handling for Ethereum chain is still valid or not. We can confirm with Tanmay or Bertrand in future if we need to retain that logic. But you could use `require.ErrorContains(t, err, "cannot convert gas price")` to more accurately hit that error path. Also, using a EVM chain, e.g. `BscMainnetChain()`, will probably help us to better understand this test case.

zetaclient/evm/evm_signer_test.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func getNewOutTxProcessor() *outtxprocessor.Processor {
+	logger := zerolog.Logger{}
+	return outtxprocessor.NewOutTxProcessorManager(logger)
+}
+
+func getCCTX() (*types.CrossChainTx, error) {
+	var cctx crosschaintypes.CrossChainTx
+	err := testutils.LoadObjectFromJSONFile(&cctx, path.Join("../", testutils.TestDataPathCctx, "cctx_56_68270.json"))
```
ws4charlie, line 69: no blocker. Just a FYI, I created a `LoadCctxByNonce` in testutils for future unit tests.

zetaclient/evm/evm_signer_test.go
```diff
@@ -0,0 +1,296 @@
...
+	mockChainClient, err := getNewEvmChainClient()
+	require.NoError(t, err)
+
+	evmSigner.TryProcessOutTx(cctx, processorManager, "123", mockChainClient, stub.NewZetaCoreBridge(), 123)
+
+	//Check if cctx was signed and broadcasted
+	list := evmSigner.GetReportedTxList()
+	found := false
+	for range *list {
+		found = true
```
ws4charlie, line 88: Is it a good idea to compare the signed tx's hash against the archived cctx's outbound tx hash? otherwise, you would just simply need a `require.Len(t, *list, 1)`

zetaclient/evm/evm_signer_test.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func TestSigner_SignRevertTx(t *testing.T) {
+	// Setup evm signer
+	evmSigner, err := getNewEvmSigner()
+	require.NoError(t, err)
+
+	// Setup txData struct
+	txData := BaseTransactionData{}
+	cctx, err := getCCTX()
```
ws4charlie, line 131: Is this CCTX real in BSC testnet?

zetaclient/evm/evm_signer_test.go
```diff
@@ -0,0 +1,296 @@
...
+		tx, err := evmSigner.SignERC20WithdrawTx(&txData)
+		require.NoError(t, err)
+
+		evmSigner.BroadcastOutTx(tx, cctx, zerolog.Logger{}, sdktypes.AccAddress{}, stub.NewZetaCoreBridge(), &txData)
+
+		//Check if cctx was signed and broadcasted
+		list := evmSigner.GetReportedTxList()
+		found := false
+		for range *list {
+			found = true
```
ws4charlie, line 292: same as above

zetaclient/evm/evm_signer_test.go
ws4charlie, line 131: Is this CCTX a real revert outtx in BSC?

zetaclient/testdata/cctx/cctx_56_68270.json
```diff
@@ -0,0 +1,43 @@
...
+  "outbound_tx_params": [
+    {
+      "receiver": "0xb0C04e07A301927672A8A7a874DB6930576C90B8",
+      "receiver_chainId": 56,
+      "coin_type": 1,
+      "amount": "657177295293237048",
+      "outbound_tx_tss_nonce": 68270,
+      "outbound_tx_gas_limit": 21000,
+      "outbound_tx_gas_price": "6000000000",
+      "outbound_tx_hash": "",
```
ws4charlie, line 33: I think if we could download the full info the the cctx of `/zeta-chain/crosschain/cctx/56/68270`, then we could compare the outbound parameters, e.g. the `outbound_tx_hash` field. Please see the downloaded cctx.
[cctx_56_68270.json](https://github.com/zeta-chain/node/files/14500516/cctx_56_68270.json)

zetaclient/evm/evm_signer_test.go
ws4charlie: If we could download the full info the the cctx of `/zeta-chain/crosschain/cctx/56/68270`, then we would be able to compare the outbound parameters, e.g. the `outbound_tx_hash` field. Please see the downloaded cctx.
[cctx_56_68270.json](https://github.com/zeta-chain/node/files/14500516/cctx_56_68270.json)

ws4charlie (commented):
left a few minor comments

zetaclient/evm/base_transaction_data.go
kevinssgh: This function returns a pointer common.GetChainFromChainID, so I decided to maintain the pointer to avoid conversions or dereferencing.

zetaclient/evm/base_transaction_data_test.go
kevinssgh: I think that may have been a typo, should be BscMainnetChain()

zetaclient/evm/evm_signer_test.go
kevinssgh: Its not a revert, just a sample cctx.

zetaclient/evm/outbound_transaction_data_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	t.Run("SetupGas_error", func(t *testing.T) {
+		cctx.GetCurrentOutTxParam().OutboundTxGasPrice = "invalidGasPrice"
+		chain := corecommon.BscMainnetChain()
+		err := txData.SetupGas(cctx, logger, evmSigner.EvmClient(), &chain)
+		require.ErrorContains(t, err, "cannot convert gas price")
+	})
+}
+
+func TestSigner_SetTransactionData(t *testing.T) {
```
lumtis, line 71: Seems like we're not covering most error cases for this function

zetaclient/evm/evm_signer.go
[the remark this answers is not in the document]
lumtis: I don't get it. It's a trivial change. Why not adding it here?

lumtis (commented):
Looks good, setTransactionData needs more tests, and I think we should discuss https://github.com/zeta-chain/node/pull/1809#discussion_r1511427310

zetaclient/evm/evm_signer.go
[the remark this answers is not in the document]
kevinssgh: The change has been made already, check the code. If you read my comment I've added the error check in each case and I've opened another issue to fix some other things in this file.

ws4charlie (approved):

lumtis (approved):
👍

Issue #1653, opened 2024-01-28
`emissions` : enable withdraw 


kingpinXD:
- Enable observers to withdraw collected emissions

PR #1811, opened 2024-02-27
feat: add message to withdraw emission rewards

kingpinXD:
# Description

- Add a message to enable observers to collect emission rewards 

Closes: https://github.com/zeta-chain/node/issues/1653

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/emissions/abci.go
```diff
@@ -146,3 +146,40 @@ func DistributeTssRewards(ctx sdk.Context, amount sdk.Int, bankKeeper types.Bank
...
+
+func EndBlocker(ctx sdk.Context, keeper keeper.Keeper) {
+	allWithdrawEmissions := keeper.GetAllWithdrawEmissions(ctx)
+	// delete all withdraw emissions irrespective of weather they are processed or not; this is to avoid processing the same withdraw emission again if in case it fails
+	// if a witdraw emission fails, it will need to be created again by the observer
+	defer func() {
+		for _, withdrawEmission := range allWithdrawEmissions {
+			keeper.DeleteWithdrawEmissions(ctx, withdrawEmission.Address)
+		}
+	}()
```
lumtis, line 158: Why using a defer here? It seems we never return an error in the function body

lumtis (commented):
I'm not sure to understand.
Why creating a withdraw object in the message and processing it in the endblocker, instead of directly processing it in the message?
What does the `WithdrawEmission` object brings on top of `WithdrawableEmission`?

kingpinXD:
> I'm not sure to understand. Why creating a withdraw object in the message and processing it in the endblocker, instead of directly processing it in the message? What does the `WithdrawEmission` object brings on top of `WithdrawableEmission`?

This is to prevent more than one emission withdraw in the same block. Although I can't really think of a condition which,would allow observers to withdraw more rewards than they own , Using the endblock and processing all withdraws in the block at once makes the logic a bit safer 
I don't think it adds much of an overhead on top of a regular tx .

x/emissions/abci.go
kingpinXD: I used it just for code readability.

lumtis:
> > I'm not sure to understand. Why creating a withdraw object in the message and processing it in the endblocker, instead of directly processing it in the message? What does the `WithdrawEmission` object brings on top of `WithdrawableEmission`?
> 
> This is to prevent more than one emission withdraw in the same block. Although I can't really think of a condition which,would allow observers to withdraw more rewards than they own , Using the endblock and processing all withdraws in the block at once makes the logic a bit safer I don't think it adds much of an overhead on top of a regular tx .

I don't get it. 

We have a `WithdrawableEmissions` object, already in the store, containing the amount an address can withdraw.
We're implementing a `MsgWithdrawEmission` that creates a `WithdrawEmissions` containing the exact same fields as above.
The end blocker iterates these `WithdrawEmissions`, send funds to user, delete `WithdrawEmissions` and subtract `WithdrawableEmissions` from value in `WithdrawEmissions`

Why not directly having `MsgWithdrawEmission` that decrease `WithdrawableEmissions` and send funds to user?

> This is to prevent more than one emission withdraw in the same block.

Why is it a problem if we verify that the withdrawn amount is below withdrawable amount?

> Although I can't really think of a condition which,would allow observers to withdraw more rewards than they own

How does using a intermediate end blocker add more protection for this?

kingpinXD:
Closing pr for now to make changes mentioned above cc @lumtis

Issue #1651, opened 2024-01-27
Allows to provide arguments to E2E tests


lumtis:
Current type that define what a E2E test is:
```
type SmokeTestFunc func(*SmokeTestRunner)
```
(smoke test to be renamed into e2e test)

For more flexibility we need to add the ability to provide arguments to a test.

Example: provide the amount to be deposit of ETh in the `eth_deposit` test

To be generic it can be a array of string:

```
type TestFunc func(*Runner, []string) error
```

PR #1816, opened 2024-02-28
test: add args to e2e tests

skosito:
# Description

- add functionality to add args to e2e tests
- implemented only on `eth_deposit` as illustration
- change `list-tests` command to also print arguments description
- modify `run` command to accept `testName:args` cli arguments
- refactor runner methods a bit to support this addition

Closes: #1651 

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [x] Go integration tests

# Checklist:

### DIFF
--- a/cmd/zetae2e/run.go
+++ b/cmd/zetae2e/run.go
@@ -54,7 +54,7 @@ For example: zetae2e run deposit:1000 withdraw: --config config.yml`,
 
 func runE2ETest(cmd *cobra.Command, args []string) error {
 	// read the config file
-	configPath, err := cmd.Flags().GetString("config")
+	configPath, err := cmd.Flags().GetString(flagConfig)
 	if err != nil {
 		return err
 	}


e2e/e2etests/test_eth_deposit.go
```diff
@@ -20,16 +20,29 @@ import (
 )
 
 // TestEtherDeposit tests deposit of ethers
-func TestEtherDeposit(r *runner.E2ERunner) {
-	hash := r.DepositEtherWithAmount(false, big.NewInt(10000000000000000)) // in wei (0.01 eth)
+func TestEtherDeposit(r *runner.E2ERunner, args []string) {
+	defaultAmount := big.NewInt(10000000000000000) // default value of 0.01 ETH in wei
```
lumtis, line 24: Maybe we can group all the default argument as part of the `E2ETest` object in `e2etests.go`?

e2e/e2etests/test_eth_deposit.go
```diff
@@ -20,16 +20,29 @@ import (
...
+	amount := new(big.Int).Set(defaultAmount)
+
+	if len(args) > 1 {
+		r.Logger.Error("TestEtherDeposit accepts only one argument for the amount, using default value")
```
lumtis, line 28: I think we can instead throw an error in the test if argument not valid (which is currently done by panicking)

cmd/zetae2e/run.go
```diff
@@ -67,6 +85,18 @@ func runE2ETest(cmd *cobra.Command, args []string) error {
 		return errors.New("invalid EVM address")
 	}
 
+	// parse test names and arguments from cmd args
+	testArgsMap := make(map[string][]string)
```
lumtis, line 89: I think it would be better to use a slice. Iterating the map is not deterministic and the user might want to run the list of test in a specific order (example: depositing before withdrawing)

cmd/zetae2e/run.go
```diff
@@ -67,6 +85,18 @@ func runE2ETest(cmd *cobra.Command, args []string) error {
...
+	for _, arg := range args {
+		parts := strings.SplitN(arg, ":", 2)
+		testName := parts[0]
+		testArgs := []string{}
+		if len(parts) > 1 && parts[1] != "" {
+			testArgs = strings.Split(parts[1], ",")
+		}
+		testArgsMap[testName] = testArgs
```
lumtis, line 97: I think we can create a separate function for this logic

lumtis (commented):
Looks a good direction. Left a few comments.

lumtis:
> @lumtis i addressed all the comments and also pushed default args for most of existing e2e tests, where adding arg was simple (eg amount)
> 
> some of the tests that seems like they dont need args (pausing, updating bytecode) or have more than 1-2 args (swap) i left without for now, but let me know if i need to add these, and if all tests need args

Sounds good. Currently we haven't made the advanced tests like the swap one functional for live networks. So we can add the arguments later for those.

e2e/e2etests/test_eth_deposit.go
lumtis: We can add the argument as we test the different commands on live network. Though we can already do it for the basic tests:
- eth withdraw
- eth deposit
- erc20 withdraw
- erc20 deposit
- btc withdraw
- btc deposit

lumtis (commented):
Looks good, left a few comments. I'm gonna try the CLI on live networks

lumtis:
FYI @skosito just added some instruction to run the command in local env: c4a9e02d
Seems to work pretty well

lumtis (commented):
Need to update the readme: https://github.com/zeta-chain/node/blob/develop/cmd/zetae2e/README.md
to reflect the changes

cmd/zetae2e/run.go
```diff
@@ -40,7 +54,11 @@ func NewRunCmd() *cobra.Command {
 
 func runE2ETest(cmd *cobra.Command, args []string) error {
 	// read the config file
-	conf, err := config.ReadConfig(args[0])
+	configPath, err := cmd.Flags().GetString("config")
```
kevinssgh, line 57: Maybe use a const for the "config" string

e2e/e2etests/e2etests.go
```diff
@@ -48,169 +48,262 @@ const (
...
+			runner.ArgDefinition{Description: "amount in azeta", DefaultValue: "10000000000000000000"},
+		},
 		TestMessagePassing,
-	},
-	{
+	),
+	runner.NewE2ETest(
 		TestZRC20SwapName,
 		"swap ZRC20 USDT for ZRC20 ETH",
+		[]runner.ArgDefinition{},
```
kevinssgh, line 129: Should we have default args for amounts to swap?

e2e/e2etests/test_crosschain_swap.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/zeta-chain/zetacore/x/crosschain/types"
 )
 
-func TestCrosschainSwap(r *runner.E2ERunner) {
+func TestCrosschainSwap(r *runner.E2ERunner, _ []string) {
```
kevinssgh, line 15: I see that there are values that can be passed as arguments for the swap. Having them hardcoded is fine but it is something to consider.

e2e/e2etests/test_erc20_multiple_withdraws.go
```diff
@@ -10,15 +10,36 @@ import (
...
+		panic("TestMultipleWithdraws requires exactly two arguments: the withdrawal amount and the number of withdrawals.")
+	}
+
+	withdrawalAmount, ok := big.NewInt(0).SetString(args[0], 10)
+	if !ok || withdrawalAmount.Cmp(approvedAmount) >= 0 {
+		panic("Invalid withdrawal amount specified for TestMultipleWithdraws.")
+	}
+
+	numberOfWithdrawals, ok := big.NewInt(0).SetString(args[1], 10)
+	if !ok || numberOfWithdrawals.Int64() < 1 {
```
kevinssgh, line 25: We could check for a max number of withdrawals to avoid an excessive number.

kevinssgh (commented):
Looks good, minor comments.

e2e/e2etests/test_erc20_multiple_withdraws.go
kevinssgh: Yes, I saw the amount limit. Not a big problem but there could be many small withdrawals pointed at some production network. I will approve anyway but we can address this later if it is an issue.

kevinssgh (approved):
Looks good

PR #1817, opened 2024-02-28
feat: v14 migration

kingpinXD:
# Description
Hotfix: Resetes the pending nonces for testnet only. This would fix the issue caused by running the wrong migration script in v13 . 
The new migration script still sets `ZetaACccounting` for mainnet , but does not reset the nonces .

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/app/setup_handlers.go
+++ b/app/setup_handlers.go
@@ -8,7 +8,8 @@ import (
 	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
 )
 
-const releaseVersion = "v13"
+// "v13-testnet" is the release version for the v13 testnet upgrade
+const releaseVersion = "v14"
 
 func SetupHandlers(app *App) {
 	app.UpgradeKeeper.SetUpgradeHandler(releaseVersion, func(ctx sdk.Context, plan types.Plan, vm module.VersionMap) (module.VersionMap, error) {
@@ -17,7 +18,9 @@ func SetupHandlers(app *App) {
 		for m, mb := range app.mm.Modules {
 			vm[m] = mb.ConsensusVersion()
 		}
-		VersionMigrator{v: vm}.TriggerMigration(crosschaintypes.ModuleName)
+		// Set the consensus version for the crosschain module to 4.
+		// This would trigger the migration script for Version 4 to 5
+		vm[crosschaintypes.ModuleName] = 4
 
 		return app.mm.RunMigrations(ctx, app.configurator, vm)
 	})
@@ -36,13 +39,4 @@ func SetupHandlers(app *App) {
 		// instead the default which is the latest version that store last committed i.e 0 for new stores.
 		app.SetStoreLoader(types.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
-}
-
-type VersionMigrator struct {
-	v module.VersionMap
-}
-
-func (v VersionMigrator) TriggerMigration(moduleName string) module.VersionMap {
-	v.v[moduleName] = v.v[moduleName] - 1
-	return v.v
 }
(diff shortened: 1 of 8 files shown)


app/setup_handlers.go
```diff
@@ -8,7 +8,8 @@ import (
 	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
 )
 
-const releaseVersion = "v13"
+// "v13-testnet" is the release version for the v13 testnet upgrade
```
lumtis, line 11:
```suggestion
```
Is it a comment that didn't get removed?

x/crosschain/migrations/v5/migrate_test.go
```diff
@@ -61,6 +63,76 @@ func TestMigrateStore(t *testing.T) {
...
+
+		for _, chain := range testnetChains {
+			pn, found := zk.ObserverKeeper.GetPendingNonces(ctx, tss.TssPubkey, chain.ChainId)
+			require.True(t, found)
+			require.Equal(t, assertValues[chain], pn.NonceHigh)
+			require.Equal(t, assertValues[chain], pn.NonceLow)
+			cn, found := zk.ObserverKeeper.GetChainNonces(ctx, chain.ChainName.String())
+			require.True(t, found)
+			require.Equal(t, uint64(assertValues[chain]), cn.Nonce)
+		}
```
lumtis, line 132: To make sure we can set values for mainnet chains and check these are not modified?

x/crosschain/migrations/v5/migrate.go
```diff
@@ -25,12 +25,60 @@ type crosschainKeeper interface {
...
+	observerKeeper types.ObserverKeeper,
+) {
+	tss, found := observerKeeper.GetTSS(ctx)
+	if !found {
+		return
+	}
+	for chain, nonce := range CurrentTestnetChains() {
+		cn, found := observerKeeper.GetChainNonces(ctx, chain.ChainName.String())
+		if !found {
+			continue
```
ws4charlie, line 49: To confirm, Is there any chance the `GetChainNonces` returns a `found==false` but below `GetPendingNonces` returns a `found==true`, in case we are not reseting both.

kingpinXD: Do you mean the other way around? As in what would happen if we set chain nonces but are unable to set pending nonces for a particular chain ?  The outbounds for that chain would not get unblocked , but that is still the same as now . However the chances of that happening are almost impossible since the values exist on testnet right now . We can query to verify .
We can use tmpctx to make it atomic but I don't think its needed for this , wdyt ?

x/crosschain/migrations/v5/migrate.go
```diff
@@ -25,12 +25,60 @@ type crosschainKeeper interface {
...
+	}
+}
+
+type Nonce struct {
+	nonceHigh uint64
+	nonceLow  uint64
+}
+
+func CurrentTestnetChains() map[common.Chain]Nonce {
+	return map[common.Chain]Nonce{
```
lumtis, line 69: We should use a slice instead of a map to avoid determinism issue. Don't think these is issues but to be sure.

lumtis:
Added some logs for testing

x/crosschain/migrations/v5/migrate.go
kingpinXD: a4bf35f7ae8b41e36b395f0647297d6fdce7e24e

ws4charlie (approved):

akimov99900:
x/crosschain/migrations/v5/migrate.go

lumtis:
Closing as released

Issue #1807, opened 2024-02-26
Follow up on config/params separation in zetaclient


skosito:
**Describe the Issue**
There are 2 things that are leftovers from previous solution:
- check if we can reduce usage of pointers in config/core context
- add missing tests for core context

For more details check these 2 comments:
https://github.com/zeta-chain/node/pull/1774#discussion_r1500313189
https://github.com/zeta-chain/node/pull/1774#discussion_r1501071990

PR #1774, opened 2024-02-19
refactor: remove params from config and introduce app context

skosito:
# Description

Main purpose of linked issue is to remove runtime params from config. 
Simple split of these params and config resulted in one more argument being passed around from top level (called `ZetaCoreContext`). However, since config and these params are still related, it makes sense to have them together in top level struct (called `AppContext`).

I also added `masterLogger` to `AppContext` as illustration and proposal that these common application context specific things can be grouped together and passed around and accessible as needed in convenient way.

This is opened as draft PR to get some feedback on this solution and then will be converted in proper PR and tested.

Closes: #1608 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

Still in draft phase.

- [ ] Tested CCTX in localnet
- [ ] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

PR #1831, opened 2024-03-04
refactor: removing unnecessary pointers in context structure

skosito:
# Description

2 parts leftovers from discussions in PR: https://github.com/zeta-chain/node/pull/1774

- add missing unit tests for zeta_core_context

- reduce pointers in config/coreContext/appContext

i think pointers should be removed when we want immutability as config, but also considered when there is large struct and mutability (as with core context) so i tried to remove pointers from config struct and also pass config around without pointers (this maybe can be revisited since config might be too big for this?)

Closes: #1807 

## Type of change

Refactor

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

# Checklist:


zetaclient/config/types.go
```diff
@@ -146,6 +138,7 @@ func (c *Config) GetKeyringBackend() KeyringBackend {
 	return c.KeyringBackend
 }
 
+// TODO: is this function duplicate of one from observertypes?
```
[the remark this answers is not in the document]
lumtis: We should only keep the one in observer module.
I'm pretty sure we had already an issue created for this, can't find it anymore. We might want to create a new one and link it here

zetaclient/core_context/zeta_core_context_test.go
```diff
@@ -0,0 +1,261 @@
+package corecontext
```
lumtis, line 1:
```suggestion
package corecontext_test
```
Ideally it is better to use test package unless there are methods we need to keep private

lumtis:
It's a good direction, I think it simplify the code not using pointer when unnecessary

changelog.md
```diff
@@ -11,6 +11,7 @@
 * [1511](https://github.com/zeta-chain/node/pull/1511) - move ballot voting logic from `crosschain` to `observer`
 * [1783](https://github.com/zeta-chain/node/pull/1783) - refactor zetaclient metrics naming and structure
 * [1774](https://github.com/zeta-chain/node/pull/1774) - split params and config in zetaclient
+* [1831](https://github.com/zeta-chain/node/pull/1831) - split params and config in zetaclient follow up
```
lumtis, line 14: Can we make it more descriptive? "removing unnecessary pointers in context structure"

lumtis, line 14: It also needs to be put under `#unreleased`

lumtis (approved):
Looks good to me, just some comments on the changelogs

changelog.md
lumtis: Never mind, we just need to move v14: https://github.com/zeta-chain/node/pull/1839

Issue #1795, opened 2024-02-22
@zetachain/node-types improvements


GMaiolo:
**Is your feature request related to a problem? Please describe.**
1. It's hard to know where to look for a specific type due to a main `index.d.ts` not being present; ideally this index in the root of the package folder would export all types for easy imports.
2. Not being able to find matching types for APIs
For example when hitting `/zeta-chain/crosschain/cctx/{CCTX_INDEX}`, I see the `node/typescript/crosschain/cross_chain_tx_pb.d.ts` file, which has all types in camel case, but the API responds with snake case
Also, enums are number-based, but the API responds with string literals, for example:
![image](https://github.com/zeta-chain/node/assets/9776436/c04b5bb5-1add-4fbf-9a4a-0550d1325c7c)

**Describe the solution you'd like**
1. A generated `index.d.ts` in the root exporting all types
2. Update types to match exactly what the APIs return so these types can be used properly.

fadeev:
We might need to switch to a different generator as this one supports only camel case seems like:

https://github.com/bufbuild/protobuf-es/blob/17327dbfd44fbff2f9c2f6c2fc0981b67153c2fe/docs/generated_code.md#field-names

fadeev:
@GMaiolo @GMaiolo we need to switch to a different TS generator. I've been experimenting with Telescope and it seems to be working much better than generic proto-to-ts buf plugins.

Here's the generated client:

https://github.com/zeta-chain/node/tree/110646249ac485c85e9feaa3eea79aca2b247533/zetachaints/src/codegen/zetachain

For example, for foreign coins:

https://zetachain-athens.blockpi.network/lcd/v1/public/zeta-chain/fungible/foreign_coins

It generates proper types that match what the API is returning:

https://github.com/zeta-chain/node/blob/110646249ac485c85e9feaa3eea79aca2b247533/zetachaints/src/codegen/zetachain/fungible/foreign_coins.ts#L37C30-L48

GMaiolo:
@fadeev nice!
What about cctxs for example? I see this
https://github.com/zeta-chain/node/blob/110646249ac485c85e9feaa3eea79aca2b247533/zetachaints/src/codegen/zetachain/crosschain/cross_chain_tx.ts

```
export enum CctxStatus {
  /** PendingInbound - some observer sees inbound tx */
  PendingInbound = 0,
  /** PendingOutbound - super majority observer see inbound tx */
  PendingOutbound = 1,
[...]

PR #1845, opened 2024-03-06
Generate TS client with Telescope

fadeev:


lumtis:
Are you still working on this one @fadeev ? Seems like to be multiple conflicts in the proto files

fadeev:
@lumtis we can close this PR, but the underlying problem is still there, maybe you can help. We can't generate TS clients because proto files use options unsupported by client generators:

https://github.com/zeta-chain/node/issues/1795#issuecomment-1982659719

Issue #1838, opened 2024-03-05
Remove duplicate ValidateChainParams in zetaclient


skosito:
**Describe the Issue**

This is a duplicated and maintained on 2 places (config types in zetaclient and observer types). The one from observer types should be used.

**Expected Outcome**
Remove duplicated func.

PR #1863, opened 2024-03-08
refactor: remove duplicate ValidateChainParams function

skosito:
# Description

Removing duplicate ValidateChainParams function in zetaclient and using one from observer types, so it is not maintained on 2 places.

Closes: #1838 

## Type of change

Cleanup

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

# Checklist:

### DIFF
--- a/zetaclient/zetabridge/zetacore_bridge.go
+++ b/zetaclient/zetabridge/zetacore_bridge.go
@@ -218,7 +218,7 @@ func (b *ZetaCoreBridge) UpdateZetaCoreContext(coreContext *corecontext.ZetaCore
 
 	// check and update chain params for each chain
 	for _, chainParam := range chainParams {
-		err := config.ValidateChainParams(chainParam)
+		err := observertypes.ValidateChainParams(chainParam)
 		if err != nil {
 			b.logger.Warn().Err(err).Msgf("Invalid chain params for chain %d", chainParam.ChainId)
 			continue
--- a/changelog.md
+++ b/changelog.md
@@ -33,6 +33,7 @@
 ### Chores
 
 * [1814](https://github.com/zeta-chain/node/pull/1814) - fix code coverage ignore for protobuf generated files
+* [1863](https://github.com/zeta-chain/node/pull/1863) - remove duplicate ValidateChainParams function
 
 ## Version: v14
 
--- a/zetaclient/config/types.go
+++ b/zetaclient/config/types.go
@@ -2,13 +2,10 @@ package config
 
 import (
 	"encoding/json"
-	"fmt"
 	"strings"
 	"sync"
 
-	ethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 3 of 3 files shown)


lumtis (approved):

zetaclient/zetabridge/zetacore_bridge.go
```diff
@@ -218,7 +218,7 @@ func (b *ZetaCoreBridge) UpdateZetaCoreContext(coreContext *corecontext.ZetaCore
 
 	// check and update chain params for each chain
 	for _, chainParam := range chainParams {
-		err := config.ValidateChainParams(chainParam)
+		err := observertypes.ValidateChainParams(chainParam)
```
kingpinXD, line 221: I think it would make sense to move the validation and the chain params structure to the common package. It would be better to keep zetaclient and zetacore binaries as independent of each other as possible

lumtis: I don't so for this one. I feel like we're putting too many things in the `common` package while it should be more used for general, generic data types.
Here `ValidateChainParams` is clearly part of the `observer` namespace since the type is defined and stored in this module. Putting the validation the validation logic outside is contrary to the idea of modularity of the SDK.
>It would be better to keep zetaclient and zetacore binaries as independent of each other as possible
ZetaClient must import `observertypes` in any case to interact with the client.

Another example is the header and proof objects that we should move from common to lightclient once the module is created.

kingpinXD: I feel like all types and related functions which are used by both the client and the core should live in the common package , if possible .

Just so that we have a clear separation between them.

lumtis: Yes, it theory it shouldn't be needed on ZetaClient as invalid chain params would be a broken invariant on the blockchain. But I think we can keep it as an extra measure.

>I feel like all types and related functions which are used by both the client and the core should live in the common package , if possible .

The ZetaClient depends on ZetaCore in any case since it must communicate through its API, I don't think there is an issue on this side. It's having ZetaCore having some dependencies on ZetaClient that would be an anti-pattern

kingpinXD: The validation is good to have on both. I am suggesting having the common as something that is imported by both the zetacore and zetaclient so that they can remain independent of each other. Which I believe suits the naming `common`. 

However, if that is not needed or desirable, we can import zetacore into the client .

kingpinXD: That seems fine to me; we can merge this for now and discuss later . 
Zetacore does import common packages already in some instances. For example, Chain is defined in common.  I cant think of any reason we would not be able to move other types to common as well

kingpinXD (approved):

Issue #1880, opened 2024-03-11
fix: Lower the gas price multiplier for EVM chains


ws4charlie:
**Describe the Issue**
We're using a hardcoded multiplier `2` for all chains. We can lower the gas price multiplier as the EVM gas stability pool issue was fixed during mainnet launch.

**Expected Outcome**
A lower gas multiplier `1.2` will be applied to EVM chains. BTC chain will be still using the `2.0` gas price multiplier.

PR #1881, opened 2024-03-11
fix: lower the gas price multiplier for evm chains

ws4charlie:
# Description

- Lower evm chain outtx gas price multiplier (`2.0 -> 1.2`).

Closes: [1880](https://github.com/zeta-chain/node/issues/1880)

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:


lumtis (commented):
Maybe we can create a map using the chain ID instead and using 1 by default? We should not need something different for BSC for example

common/constant.go
```diff
@@ -1,8 +1,11 @@
 package common
 
 const (
-	// DefaultGasPriceMultiplier is the default gas price multiplier for outbond txs
-	DefaultGasPriceMultiplier = 2
+	// EVMOuttxGasPriceMultiplier is the default gas price multiplier for EVM-chain outbond txs
+	EVMOuttxGasPriceMultiplier = 1.2
+
+	// BTCOuttxGasPriceMultiplier is the default gas price multiplier for BTC outbond txs
+	BTCOuttxGasPriceMultiplier = 2.0
```
lumtis, line 8: I think we should move this in `zetaclient` package. `common` should define types used through the project but here this is specifically a value for ZetaClient

common/utils.go
```diff
@@ -42,6 +42,16 @@ func StringToHash(chainID int64, hash string) ([]byte, error) {
 	return nil, fmt.Errorf("cannot convert hash to bytes for chain %d", chainID)
 }
 
+// GasPriceMultiplier returns the gas price multiplier for the given chain
+func GasPriceMultiplier(chainID int64) float64 {
```
lumtis, line 46: Same comment as above

common/constant.go
kingpinXD: Actually, I think this might be a good place for the multiplier; this same multiplier should also be used for creating the TSS migration. outbound  . Whitelist erc20
https://github.com/zeta-chain/zeta-node/blob/7e25cd015c24665e685f329fda371bac01b9ea94/x/crosschain/keeper/msg_server_migrate_tss_funds.go#L110-L110

https://github.com/zeta-chain/zeta-node/blob/7e25cd015c24665e685f329fda371bac01b9ea94/x/crosschain/keeper/msg_server_whitelist_erc20.go#L144-L144

common/utils.go
```diff
@@ -42,6 +42,16 @@ func StringToHash(chainID int64, hash string) ([]byte, error) {
...
+	if IsEVMChain(chainID) {
+		return EVMOuttxGasPriceMultiplier
+	} else if IsBitcoinChain(chainID) {
+		return BTCOuttxGasPriceMultiplier
+	}
+	return 1.0
```
kingpinXD, line 52: Should we also return an error from this function? I don't think we would ever be using the 1.0 value , would we ?

common/constant.go
lumtis: >this same multiplier should also be used to create the TSS migration. outbound and Whitelist erc20

What do you mean by this?
The multiplier is used to vote on the gas price, it is therefore already taken into consideration there.
So far the direct usage of the multiplier is only in ZetaClient so I suggest to move it here

ws4charlie: https://github.com/zeta-chain/node/pull/1881/commits/b396ccb2db2dfdf16cfa02092d66de12b2c15c20

common/utils.go
ws4charlie: https://github.com/zeta-chain/node/pull/1881/commits/b396ccb2db2dfdf16cfa02092d66de12b2c15c20

common/utils.go
ws4charlie: https://github.com/zeta-chain/node/pull/1881/commits/b396ccb2db2dfdf16cfa02092d66de12b2c15c20

ws4charlie:
> Maybe we can create a map using the chain ID instead and using 1 by default? We should not need something different for BSC for example

We only differentiate EVM chain from BTC chain ATM. In the future, we'll either completely get rid of this gasPriceMultiplier or formalize it as a part of ChainParam (if we still need it).

common/constant.go
kingpinXD: okay just saw your comment @ws4charlie 
https://github.com/zeta-chain/node/pull/1881#issuecomment-1999958168

I think that's exactly what I was referring too.Resolving this for now , as this seems out of scope for this pr

kingpinXD (approved):

zetaclient/common/utils.go
```diff
@@ -50,3 +53,13 @@ func PrintComplianceLog(
 	logWithFields1.Warn().Msg(logMsg)
 	logWithFields2.Warn().Msg(logMsg)
 }
+
+// GasPriceMultiplier returns the gas price multiplier for the given chain
+func GasPriceMultiplier(chainID int64) (float64, error) {
```
lumtis, line 58: I didn't see this file but common/utils is very generic. Both names are generally not advised for package name.
Can we create new packages?
Compliance in a compliance package. 
This GasPriceMultiplier maybe in zetabridge since this is for posting the gas price on zetachain

ws4charlie: https://github.com/zeta-chain/node/pull/1881/commits/b1460f6180f10962c0c7d075811f879a4bbfef50

lumtis (approved):

Issue #1846, opened 2024-03-06
Add script for missing observations


lumtis:
Adding a script to check for missing observations.

### Possible solution

- initialize a new CLI under `cmd` package:
```
zetatool
```

- add it as part of `zetacored debug` package

- add the script under a `tool` package and provide a makefile entry to run it

PR #1884, opened 2024-03-12
feat: added new cmd "zetatool" and sub cmd filterdeposits 

kevinssgh:
# Description

A new command was added to the node repository as a utility tool. The first tool added is a filter for missed inbound transactions on BTC and ETH networks. 

Closes: https://github.com/zeta-chain/node/issues/1846

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

# Checklist:


cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	TssAddressEVM             = "0x70e967acfcc17c3941e87562161406d41676fd83"
+	BtcExplorer               = "https://blockstream.info/api/address/bc1qm24wp577nk8aacckv8np465z3dvmu7ry45el6y/txs"
+	EthRPC                    = "https://rpc.ankr.com/eth/2da24e4a1fd28f2bec1569eceb2c38a5694b7f5c83fd24c69ae714a89a514f9b"
+	ConnectorAddress          = "0x000007Cf399229b2f5A4D043F20E90C9C98B7C6a"
+	CustodyAddress            = "0x0000030Ec64DF25301d8414eE5a29588C4B0dE10"
+	EvmStartBlock      uint64 = 19200110
+	EvmMaxRange        uint64 = 1000
+)
+
+type Config struct {
```
lumtis, line 25: Need comment on the structure. What config it is for.
Same for the field unless the name is clear.
What BtcExplorer represent? url? API key?

cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
+package config
```
lumtis, line 1: This seems related to the filtering feature, this should be in the deposit filter package

cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/json"
+
+	"github.com/spf13/afero"
+)
+
+var AppFs = afero.NewOsFs()
+
+const (
+	Flag                      = "config"
```
lumtis, line 12:
```suggestion
	FlagConfig                      = "config"
```

cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	defaultCfgFileName        = "InboundTxFilter_config.json"
+	ZetaURL                   = "http://46.4.15.110:1317" //http://100.71.167.102:26657
```
lumtis, line 14: This needs to be removed. We should either use a local IP by default or none

cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	TssAddressBTC             = "bc1qm24wp577nk8aacckv8np465z3dvmu7ry45el6y"
+	TssAddressEVM             = "0x70e967acfcc17c3941e87562161406d41676fd83"
```
lumtis, line 16: This what can be fetched dynamically from the chain API, I don't think we need to have those in the config

cmd/zetatool/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	BtcExplorer               = "https://blockstream.info/api/address/bc1qm24wp577nk8aacckv8np465z3dvmu7ry45el6y/txs"
+	EthRPC                    = "https://rpc.ankr.com/eth/2da24e4a1fd28f2bec1569eceb2c38a5694b7f5c83fd24c69ae714a89a514f9b"
```
lumtis, line 18: We shouldn't use a default value for an API key

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/zeta-chain/zetacore/cmd/zetatool/config"
+)
+
+var btcCmd = &cobra.Command{
+	Use:   "btc",
+	Short: "Filter inbound btc deposits",
+	Run:   FilterBTCTransactions,
```
lumtis, line 21: Instead of using `log.Fatal`
Can we use
```
RunE: FilterBTCTransactions
```
and return a error in the functions

Same comments for the other commands

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,159 @@
...
+			_, found := vout0["scriptpubkey"]
+			scriptpubkey := ""
+			if found {
+				scriptpubkey = vout0["scriptpubkey"].(string)
+			}
+			_, found = vout0["scriptpubkey_address"]
+			targetAddr := ""
+			if found {
+				targetAddr = vout0["scriptpubkey_address"].(string)
+			}
```
lumtis, line 107: need some comments on this logic

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// getHashList is called by FilterBTCTransactions to help query and filter inbound transactions on btc
+func getHashList(cfg *config.Config) []Deposit {
+	var list []Deposit
+	lastHash := ""
+
+	url := cfg.BtcExplorer
+
+	for {
```
lumtis, line 51: Can we explain in comment the purpose of this iteration

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,159 @@
...
+			fmt.Println("error unmarshalling: ", err.Error())
+		}
+
+		if len(txns) == 0 {
+			break
+		}
+
+		fmt.Println("Length of txns: ", len(txns))
+
+		for _, txn := range txns {
```
lumtis, line 87: Can we explain in comment the purpose of this iteration

cmd/zetatool/filterdeposit/evm.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/spf13/cobra"
+	"github.com/zeta-chain/protocol-contracts/pkg/contracts/evm/erc20custody.sol"
+	"github.com/zeta-chain/protocol-contracts/pkg/contracts/evm/zetaconnector.non-eth.sol"
+	"github.com/zeta-chain/zetacore/cmd/zetatool/config"
+)
+
+const (
+	TopicsDeposited = 2
+	TopicsZetaSent  = 3
+	DonationMessage = "I am rich!"
```
lumtis, line 24: We already have the constant defined in the `common` package, we should reuse it

cmd/zetatool/filterdeposit/evm.go
```diff
@@ -0,0 +1,208 @@
...
+		fmt.Println("error loading filter: ", err.Error())
+		return deposits
+	}
+
+	// ********************** Get ERC20 Custody Deposit events
+	for custodyIter.Next() {
+		// sanity check tx event
+		err := CheckEvmTxLog(&custodyIter.Event.Raw, erc20CustodyAddress, "", TopicsDeposited)
+		if err == nil {
+			//fmt.Println("adding deposits")
```
lumtis, line 124: Can we remove all commented lines if these are no longer used?

cmd/zetatool/filterdeposit/evm.go
```diff
@@ -0,0 +1,208 @@
...
+			deposits = append(deposits, Deposit{
+				TxID:   connectorIter.Event.Raw.TxHash.Hex(),
+				Amount: connectorIter.Event.ZetaValueAndGas.Uint64(),
+			})
+		}
+	}
+
+	//********************** Get Transactions sent directly to TSS address
```
lumtis, line 145:
```suggestion
	// Get ERC20 Custody Deposit events
	for custodyIter.Next() {
		// sanity check tx event
		err := CheckEvmTxLog(&custodyIter.Event.Raw, erc20CustodyAddress, "", TopicsDeposited)
		if err == nil {
			//fmt.Println("adding deposits")
			deposits = append(deposits, Deposit{
				TxID:   custodyIter.Event.Raw.TxHash.Hex(),
				Amount: custodyIter.Event.Amount.Uint64(),
			})
		}
	}

	// Get Connector ZetaSent events
	for connectorIter.Next() {
		// sanity check tx event
		err := CheckEvmTxLog(&connectorIter.Event.Raw, connectorAddress, "", TopicsZetaSent)
		if err == nil {
			//fmt.Println("adding deposits")
			deposits = append(deposits, Deposit{
				TxID:   connectorIter.Event.Raw.TxHash.Hex(),
				Amount: connectorIter.Event.ZetaValueAndGas.Uint64(),
			})
		}
	}

	// Get Transactions sent directly to TSS address
```
Why these prefix in comment?
If this represent a specific separate logic in the function, we shoudl create sub functions

cmd/zetatool/filterdeposit/evm.go
```diff
@@ -0,0 +1,208 @@
...
+		fmt.Printf("getTSSDeposits returned err: %s", err.Error())
+	}
+	deposits = append(deposits, tssDeposits...)
+
+	return deposits
+}
+
+// getTSSDeposits more specifically queries and filters deposits based on direct transfers the TSS address.
+func getTSSDeposits(tssAddress string, startBlock uint64, endBlock uint64) ([]Deposit, error) {
+	client := etherscan.New(etherscan.Mainnet, "S3AVTNXDJQZQQUVXJM4XVIPBRYECGK88VX")
```
lumtis, line 157: `S3AVTNXDJQZQQUVXJM4XVIPBRYECGK88VX`
What this represent?

cmd/zetatool/main.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "zetatool",
+	Short: "utility tool for zeta-chain",
+}
+
+func init() {
+	rootCmd.AddCommand(filterdeposit.Cmd)
+	rootCmd.PersistentFlags().String(config.Flag, "", "custom config file: --config filename.json")
```
lumtis, line 19: This is the config for filtering (it would be too unclean to have a general config for all commands we may have) we should us a local flag instead of persistent for the `filterdeposit` command

cmd/zetatool/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/zeta-chain/zetacore/cmd/zetatool/filterdeposit"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "zetatool",
+	Short: "utility tool for zeta-chain",
+}
+
+func init() {
+	rootCmd.AddCommand(filterdeposit.Cmd)
```
lumtis, line 18: Would be cleaner to use a `NewFilterDepositCmd` function

docs/cli/zetatool/filterdeposit.md
```diff
@@ -0,0 +1,24 @@
+# filterdeposit
+
+Filter missing inbound deposits
+
+### Synopsis
+
+Filters relevant inbound transactions for a given network and attempts to find an associated cctx from the core. If a
```
lumtis, line 7: "from the core"
This part is unclear, ZetaChain?

tool/filter_missed_deposits/InboundTxFilter_config.json
```diff
@@ -0,0 +1,11 @@
+{
+ "ZetaURL": "http://46.4.15.110:1317",
```
lumtis, line 2: We need to remove this URL

tool/filter_missed_deposits/filter_missed_btc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+zetatool filterdeposit btc --config ./tool/filter_missed_deposits/InboundTxFilter_config.json
```
lumtis, line 3: Why created sub scripts instead of directly putting the command in the makefile?

cmd/zetatool/config/config.go
kevinssgh: I would prefer to keep it more generic in case there are more tools added. Some of the more specific fields can be passed as arguments and removed from the config file.

cmd/zetatool/main.go
kevinssgh: The config will be more generic in the next update, I'd like to have one config for several tools under zetatool_config.json

cmd/zetatool/filterdeposit/evm.go
kevinssgh: That is the etherscan API key, I removed it from the repo. I must have forgotten it there when testing.

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,188 @@
...
+
+			//Check if txn is confirmed
+			status := txn["status"].(map[string]interface{})
+			confirmed := status["confirmed"].(bool)
+			if !confirmed {
+				continue
+			}
+
+			//Filter out deposits less than min base fee
+			if vout0["value"].(float64) < 1360 {
```
ws4charlie, line 147: This fee is calculated dynamically using `CalcDepositorFee` method.

cmd/zetatool/filterdeposit/btc.go
```diff
@@ -0,0 +1,188 @@
...
+				continue
+			}
+
+			//Check if Deposit is a donation
+			scriptpubkey1 := vout1["scriptpubkey"].(string)
+			if len(scriptpubkey1) >= 4 && scriptpubkey1[:2] == "6a" {
```
ws4charlie, line 153: It feels to me the 3rd-party JSON API incurs extra tool maintenance as you'll have a maintain a JSON version of BTC intx filter which has to exactly match existing filter done by queries directly to the Bitcoin node. For instance, different type of addresses (going to have Taproot), donation filter, restricted addresses, etc.

kevinssgh: This was done long before the address changes, so it will need to be updated later on. Using explorers like this will limit some things but it runs much faster than parsing the actual blocks. The only thing that we need to match here is the way we are filtering in zetaclient, I will have to maintain it based on that. The address query in the explorer isn't specific, it will be a string and isn't tied to any format. 

I will just have to match what is being filtered in zetaclient.

cmd/zetatool/filterdeposit/btc.go
kevinssgh: I'll need to get the actual block for each transaction in order to calculate that accurately. I will see how much it slows down.

kevinssgh: Unfortunately the api used here doesn't show the block height of the transaction, the block queries aren't showing lookups by tx id. For now this fee will have to be estimated until I can find a better API.

kevinssgh: we can adjust the base fee to something reasonable ... I will try to fix it later because this might require a completely different explorer.

ws4charlie (approved):

lumtis (approved):

Issue #1715, opened 2024-02-06
Add a new policy group for operational purpose


lumtis:
We currently have two policy groups:
- Group1 - emergency actions
- Group2 - operational and admin actions

Operational (for example updating the chain params in observer) should require less signature than an admin action (migrating the TSS funds). We should introduce another policy group for this purpose.

To make things explicit, we should remove the `group1` and `group2` naming and introduce names like:
- emergency
- operational
- admin

PR #1935, opened 2024-03-22
feat: add new operational authority group

lumtis:
# Description

- Add a operational authority group for operational purpose (non-sensitive network param changes)
- Move all the admin messages to operational messages except the following which are sensitive as discussed in the doc:
  - `MsgUpdateTssAddress`
  - `MsgUpdateContractBytecode`
  - `MsgUpdateSystemContract`
  - `MsgMigrateTssFunds`

Change the migration script to set the current group 2 address to group admin and operational at the same time. The operational address should be changed with a gov prop later (with less required signatures)

Closes: https://github.com/zeta-chain/node/issues/1715


proto/authority/policies.proto
```diff
@@ -8,8 +8,9 @@ option go_package = "github.com/zeta-chain/zetacore/x/authority/types";
 // PolicyType defines the type of policy
 enum PolicyType {
   option (gogoproto.goproto_enum_stringer) = true;
-  groupEmergency = 0;
-  groupAdmin = 1;
+  groupEmergency = 0; // Used for emergency situations that require immediate action
+  groupOperational = 1; // Used for operational tasks like changing non-sensitive protocol parameters
+  groupAdmin = 2; // Used for administrative tasks like changing sensitive protocol parameters or moving funds
```
kingpinXD, line 13: Maybe use `groupAdministrative` , it sounds more similar to the other two group names

x/fungible/keeper/msg_server_update_zrc20_paused_status.go
```diff
@@ -31,7 +31,7 @@ func (k msgServer) UpdateZRC20PausedStatus(
 	// unpausing requires group2 admin
```
kingpinXD, line 31:
```suggestion
	// unpausing requires operational policy
```

x/observer/keeper/msg_server_update_observer.go
```diff
@@ -81,7 +81,7 @@ func (k Keeper) CheckUpdateReason(ctx sdk.Context, msg *types.MsgUpdateObserver)
 	case types.ObserverUpdateReason_AdminUpdate:
 		{
 			// Group admin is authorized to update observer
```
kingpinXD, line 83:
```suggestion
			// Operational policy is required to update an observer for admin update 
```

kingpinXD: Alos I think this should be policyAdmin , this affects the observer set , and we would almost never need to do this . Observers if tombstoned have the option to select `ObserverUpdateReason_Tombstoned` reason

x/observer/migrations/v7/migrate.go
```diff
@@ -29,16 +29,24 @@ func MigratePolicies(ctx sdk.Context, observerKeeper observerKeeper) error {
 	for _, adminPolicy := range params.AdminPolicy {
 		if adminPolicy != nil {
 
-			// Convert group1 -> emergency and group2 -> admin
-			policyType := authoritytypes.PolicyType_groupAdmin
 			if adminPolicy.PolicyType == types.Policy_Type_group1 {
-				policyType = authoritytypes.PolicyType_groupEmergency
+				// for policy group 1, we set the policy type to emergency
```
kingpinXD, line 33: So even after this mightation observers.AdminPolicy would exist . Would that get removed in the next release?

lumtis: Yes, we are removing all params in any case for v47 upgrade

x/crosschain/keeper/msg_server_whitelist_erc20.go
```diff
@@ -28,7 +28,7 @@ func (k msgServer) WhitelistERC20(goCtx context.Context, msg *types.MsgWhitelist
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	// check if authorized
-	if !k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupAdmin) {
+	if !k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupOperational) {
```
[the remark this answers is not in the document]
lumtis: Yes [255fbf0](https://github.com/zeta-chain/node/pull/1935/commits/255fbf0f3973bae1231de8a071785d508913d8df)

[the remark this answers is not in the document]
lumtis: Keeping the name `IsAuthorized` in `observer` to check observer was also a bit unclear, renamed it to `IsNonTombstonedObserver`: [8ac6e7b](https://github.com/zeta-chain/node/pull/1935/commits/8ac6e7ba4bba405cd0da9995eb88e25b99a58a7f)

proto/authority/policies.proto
lumtis: I would prefer to keep it short personally

x/observer/keeper/msg_server_update_observer.go
lumtis: Sounds good to use group admin [ba9f135](https://github.com/zeta-chain/node/pull/1935/commits/ba9f13515af948497c882c29ea761be17bacecd9) if sensitive and not frequently used

app/ante/ante.go
```diff
@@ -102,7 +102,7 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 			// if tx is a system tx, and singer is authorized, use system tx handler
 
 			isAuthorized := func(creator string) bool {
```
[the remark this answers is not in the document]
lumtis: I think it's fine to keep it more general here

kingpinXD (approved):

Issue #1905, opened 2024-03-15
Replace `common` package with `pkg`


lumtis:
**Describe the Issue**
`common` package is a bit too general as a name and group too many different thing under the same namespace. Also it is generally not recommended to use `common` as a name

**Expected Outcome**
Idea of the issue is to group the different logic in common into reusable and scoped packages under `pkg`
```
pkg/chains
pkg/address
pkg/proofs
```
as conventional in GO https://pkg.go.dev/github.com/geektime007/mgmt/pkg#section-readme

PR #1936, opened 2024-03-22
refactor: replace common package with pkg

skosito:
# Description

Its a big PR, but the only place to review is `common` package renaming to `pkg` and split into subpackages (couple of files like commands, version and constant are still in `pkg` package). 
Because of this split, i also did the same for `common` proto files, now called `pkg` too.

Best way to review this is to check it out locally and check proto and pkg files, eg. if organization in subpackages in ok, if naming of subpackages is ok etc - the rest of the changes are just fixing imports (apart from proof test which i will flag in the comment).

Closes: #1905 

## Type of change

Refactor

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

# Checklist:

### DIFF
--- a/.github/workflows/execute_advanced_tests.yaml
+++ b/.github/workflows/execute_advanced_tests.yaml
@@ -19,60 +19,164 @@ on:
         type: boolean
         required: false
         default: false
+      debug:
+        type: boolean
+        required: false
+        default: false
 
 jobs:
   e2e-admin-tests:
+    if: ${{ github.event.inputs.e2e-admin-tests == 'true' }}
     runs-on: buildjet-4vcpu-ubuntu-2204
     timeout-minutes: 120
     steps:
       - name: "Checkout Code"
-        if: ${{ github.event.inputs.e2e-admin-tests == 'true' }}
         uses: actions/checkout@v3
 
       - name: Execute e2e-admin-tests
-        if: ${{ github.event.inputs.e2e-admin-tests == 'true' }}
         shell: bash
         run: |
           make start-e2e-admin-test
+          container_id=$(docker ps --filter "ancestor=orchestrator:latest" --format "{{.ID}}")
+          if [ -z "$container_id" ]; then
+              echo "Orchestrator container is not currently running. Exiting..."
+              exit 1
+          fi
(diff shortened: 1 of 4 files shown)


pkg/version.go
```diff
@@ -1,4 +1,4 @@
-package common
+package pkg
```
lumtis, line 1: Seems like `CommitHash` and `BuildTime` are not used and can be removed.

I would suggest for simmplicity also moving
```
	Name 
	Version  
```
In the single place these are used in api.go

pkg/chains/utils.go
```diff
@@ -1,4 +1,4 @@
-package common
+package chains
```
lumtis, line 1: I think we can rename this file `conversion.go`

pkg/commands.go
```diff
@@ -1,4 +1,4 @@
-package common
+package pkg
```
lumtis, line 1: I think this can be moved into the constant package.

Also the comment originally put is not very good.

These are used in cctxs of type cmd to give an instruction to execute on the external chain (sending a whitelist erc20 tx from tss for example)

```
// CmdWhitelistERC20 is used for CCTX of type cmd to give the instruction to the TSS to whitelist an ERC20 on an exeternal chain
// CmdMigrateTssFunds is used for CCTX of type cmd to give the instruction to the TSS to transfer its funds on a new address
```

lumtis (commented):
It seems most of the content in `cosmos.go` is unused and can be removed (you can see those with Goland for ex)
I might even think this package is not needed anymore but removing it might be higher than the scope of this PR

lumtis:
> only thing left in cosmos package are some legacybech32 methods, i thought to maybe leave them there instead of having legacy package imported on couple of places, and we can maybe open issue to replace those deprecated methods with proper ones?

This sounds good to me

pkg/version.go
lumtis: Ok we should keep those then.
I think it would be better to define a `constant` package to put this file and `constant.go` content instead of using pkg as a package name

lumtis (approved):

kingpinXD (approved):

Issue #1604, opened 2024-01-20
Track concurrent keysigns as a metric


CharlieMc0:
It would be helpful to track concurent keysigns via telemetry.

PR #1954, opened 2024-03-27
feat: Added metric for concurrent keysigns

kevinssgh:
# Description

1. Keysign manager was added to keep track of outbound signature attempts
2. A new metric was integrated to export the number of active signatures. 

Closes: https://github.com/zeta-chain/node/issues/1604

## Type of change

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

```
/usr/local/bin # curl localhost:8886/metrics | grep zetaclient_num_active_message_signs
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 18048    0 18048    0     0  7861k      0 --:--:-- --:--:-- --:--:-- 8812k
# HELP zetaclient_num_active_message_signs Number of concurrent key signs
# TYPE zetaclient_num_active_message_signs gauge
zetaclient_num_active_message_signs 4
```

# Checklist:


zetaclient/tss/tss_keysign_manager.go
```diff
@@ -0,0 +1,49 @@
...
+	mu                sync.Mutex
+	Logger            zerolog.Logger
+}
+
+// NewKeySignManager - constructor
+func NewKeySignManager(logger zerolog.Logger) *KeySignManager {
+	return &KeySignManager{
+		numActiveMsgSigns: 0,
+		mu:                sync.Mutex{},
+		Logger:            logger.With().Str("module", "KeySignManager").Logger(),
```
kingpinXD, line 22: Does this mean the logs would two module fields? 
wondering if we should limit it to one module and call this is a submodule or something similar ?

kingpinXD (commented):
lgtm , left a minor comment

zetaclient/tss/tss_keysign_manager.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"sync"
+
+	"github.com/rs/zerolog"
+	"github.com/zeta-chain/zetacore/zetaclient/metrics"
+)
+
+// KeySignManager is a manager to keep track of concurrent keysigns performed by go-tss
+type KeySignManager struct {
```
lumtis, line 11: We should avoid using `Manager` name for types as often too generic, unclear.

If it is solely used to track concurrent key signs, we may call it `ConcurrentKeysignsTracker`?

Also a small note:
using `KeySign` instead of `Keysign` is not consistent with `keysign` filename. If these are two separate word, the filename should contains underscore, but I would tend to keep it as a single word

zetaclient/tss/tss_signer.go
```diff
@@ -74,6 +74,7 @@ type TSS struct {
 	logger        zerolog.Logger
 	Signers       []string
 	CoreBridge    interfaces.ZetaCoreBridger
+	KeySignMan    *KeySignManager
```
lumtis, line 77:
```suggestion
	KeySignManager    *KeySignManager
```
I commented to rename this type in any case, but more generally, let's keep fully explicit names for field names

zetaclient/tss/tss_signer.go
```diff
@@ -203,7 +210,9 @@ func (tss *TSS) Sign(digest []byte, height uint64, nonce uint64, chain *chains.C
 	}
 	// #nosec G701 always in range
 	keysignReq := keysign.NewRequest(tssPubkey, []string{base64.StdEncoding.EncodeToString(H)}, int64(height), nil, "0.14.0")
+	tss.KeySignMan.StartMsgSign()
```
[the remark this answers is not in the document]
kevinssgh: This metric is really just to keep a count or gauge of the current number of transactions/messages being signed by the TSS library. The participants can vary based on the height and message hash. Each transaction isn't associated with a particular pub key but the shared TSS key.

zetaclient/tss/tss_keysign_manager.go
```diff
@@ -0,0 +1,49 @@
+package tss
+
+import (
+	"sync"
+
+	"github.com/rs/zerolog"
+	"github.com/zeta-chain/zetacore/zetaclient/metrics"
+)
+
+// KeySignManager is a manager to keep track of concurrent keysigns performed by go-tss
```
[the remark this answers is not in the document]
kevinssgh: I can rename the struct, it is too generic for what it is actually doing. Having a separate structure for this does make it a bit cleaner because it requires a mutex and some logic. It will also allow us to expand on it later on if we want to keep track of other metrics like average keysign time etc.

kevinssgh:
> @kevinssgh can we target the branch to `v16.0.0`? to include it in the next release

https://github.com/zeta-chain/node/pull/1960

lumtis (approved):

Issue #1743, opened 2024-02-08
High-priority code coverage for `observer`


lumtis:
Tests to implement in priority for the `observer` module.
This concerns the `keeper` package specifically but `types` should also be checked for the different components:

- Abci.go
- Crosschain_flags.go
- Hooks.go
- Msg_add_blame_vote.go
- Msg_add_block_header.go
- msg_add_observer.go
- Msg_update_keygen
- Pending_nonces.go
- utils.go

Issue #1875, opened 2024-03-11
Reenable missing AddTracker tests and mock header/proof functionality for AddTracker and AddBlockHeader


lumtis:
https://github.com/zeta-chain/node/pull/1872 currently comment out some of the tracker tests as those doesn't work when Goerli RPC is no longer supported

We should add test back using static data instead of using RPCs. However, since block/header verification shouldn't be a concern for adding tracker, this is a functionality performed by the proof verification system itself, this functionality should be mocked for these tests.

This is the same for `MsgAddBlockHeader` in `observer`. The actual logic of adding the block header should be mocked in the observer module.

Depending on https://github.com/zeta-chain/node/issues/1491
once the new module is implemented we can define mocks for add header and proof verification

PR #1961, opened 2024-03-28
test: improve observer module coverage

skosito:
# Description

Improves coverage for observer module from 63.9% to 92.47%. Lot of tests were missing, but hopefully this is decent coverage now. Some notes:

- some header and proofs tests are not added because of this issue https://github.com/zeta-chain/node/issues/1875
- events.go and migrator.go added to ignore in codecov
- i didnt add tests for past migrations
- couple of todos in comment, please check before merge

Closes: #1743 

## Type of change

Tests

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/x/observer/types/ballot.go
+++ b/x/observer/types/ballot.go
@@ -3,7 +3,6 @@ package types
 import (
 	"fmt"
 
-	"cosmossdk.io/errors"
 	cosmoserrors "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
@@ -16,7 +15,7 @@ func (m Ballot) AddVote(address string, vote VoteType) (Ballot, error) {
 	// `index` is used to set the vote in the `Votes` array
 	index := m.GetVoterIndex(address)
 	if index == -1 {
-		return m, errors.Wrap(ErrUnableToAddVote, fmt.Sprintf("Voter %s not in voter list", address))
+		return m, cosmoserrors.Wrap(ErrUnableToAddVote, fmt.Sprintf("Voter %s not in voter list", address))
 	}
 	m.Votes[index] = vote
 	return m, nil


x/observer/types/chain_params_test.go
```diff
@@ -196,4 +197,23 @@ func (s *UpdateChainParamsSuite) Validate(params *types.ChainParams) {
 	copy.OutboundTxScheduleLookahead = 501
 	err = types.ValidateChainParams(&copy)
 	require.NotNil(s.T(), err)
+
```
lumtis, line 200: @ws4charlie I may think from git history that you worked on the tests in this file?
Why did we define a test suite for this?

lumtis (approved):
Looks very good.
Left one comment for @ws4charlie

x/observer/types/ballot.go
```diff
@@ -14,12 +14,20 @@ func (m Ballot) AddVote(address string, vote VoteType) (Ballot, error) {
 	// `index` is the index of the `address` in the `VoterList`
 	// `index` is used to set the vote in the `Votes` array
 	index := m.GetVoterIndex(address)
+	// TODO: is this valid possibility?
```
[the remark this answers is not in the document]
lumtis: Sounds good. We can also create an issue to track it

x/observer/keeper/ballot_test.go
```diff
@@ -4,20 +4,92 @@ import (
...
+
+	k.SetBallot(ctx, b)
+
+	ballot, found := k.GetBallot(ctx, identifier)
+	require.True(t, found)
+	require.Equal(t, *b, ballot)
+}
+
+func TestKeeper_GetBallotList(t *testing.T) {
+	k, ctx := SetupKeeper(t)
```
kingpinXD, line 34: It might make sense to remove this SetupKeeper function altogether and just use the one defined under the testutil directory, what do you think?

kingpinXD, line 34: NIT , but maybe we can add a test so the same index overwrites the ballot values

x/observer/keeper/ballot_test.go
```diff
@@ -4,20 +4,92 @@ import (
...
+	identifier := "0x9ea007f0f60e32d58577a8cf25678942d2b10791c2a34f48e237b76a7e998e4d"
```
kingpinXD, line 35: We can also use `sample.ZetaIndex(t)` the ballot identifier is the same as the cctx index for the inbound ballots, 
or you could create a sample function to generate a random index as well

x/observer/keeper/ballot_test.go
```diff
@@ -4,20 +4,92 @@ import (
...
+	require.Empty(t, ballots)
+
+	k.SetBallot(ctx, b)
+	ballots = k.GetAllBallots(ctx)
+	require.Equal(t, 1, len(ballots))
+	require.Equal(t, b, ballots[0])
+}
+
+func TestKeeper_GetMaturedBallotList(t *testing.T) {
+	k, ctx := SetupKeeper(t)
```
kingpinXD, line 77: Can we add a test case with super high BallotMaturityBlocks to make sure it doesn't break anything ?

x/observer/keeper/blame_test.go
```diff
@@ -0,0 +1,110 @@
...
+	blameRecords, found := k.GetBlame(ctx, index)
+	require.True(t, found)
+	require.Equal(t, index, blameRecords.Index)
+}
+
+func TestKeeper_GetBlameByChainAndNonce(t *testing.T) {
+	k, ctx, _, _ := keepertest.ObserverKeeper(t)
+	var chainId int64 = 97
+	var nonce uint64 = 101
+	digest := "85f5e10431f69bc2a14046a13aabaefc660103b6de7a84f75c4b96181d03f0b5"
```
kingpinXD, line 37: same comment as above , maybe we can define a function in sample for this

x/observer/keeper/chain_params_test.go
```diff
@@ -40,6 +40,37 @@ func TestKeeper_GetSupportedChainFromChainID(t *testing.T) {
...
 
+func TestKeeper_GetChainParamsByChainID(t *testing.T) {
+	t.Run("return false if chain params not found", func(t *testing.T) {
+		k, ctx, _, _ := keepertest.ObserverKeeper(t)
+
+		_, found := k.GetChainParamsByChainID(ctx, getValidEthChainIDWithIndex(t, 0))
+		require.False(t, found)
+	})
+
+	t.Run("return true if not found", func(t *testing.T) {
```
kingpinXD, line 51:
```suggestion
	t.Run("return true if found", func(t *testing.T) {
```

x/observer/keeper/chain_params_test.go
```diff
@@ -40,6 +40,37 @@ func TestKeeper_GetSupportedChainFromChainID(t *testing.T) {
...
+		k, ctx, _, _ := keepertest.ObserverKeeper(t)
+		chainParams := sample.ChainParams(getValidEthChainIDWithIndex(t, 0))
+		chainParams.IsSupported = false
```
kingpinXD, line 54:
```suggestion
		
```

x/observer/keeper/chain_params_test.go
```diff
@@ -40,6 +40,37 @@ func TestKeeper_GetSupportedChainFromChainID(t *testing.T) {
...
+		})
+		res, found := k.GetChainParamsByChainID(ctx, getValidEthChainIDWithIndex(t, 0))
+		require.True(t, found)
+		require.Equal(t, chainParams, res)
+	})
+
+	t.Run("return false if chain not found in params", func(t *testing.T) {
+		k, ctx, _, _ := keepertest.ObserverKeeper(t)
+		chainParams := sample.ChainParams(getValidEthChainIDWithIndex(t, 0))
+		chainParams.IsSupported = false
```
kingpinXD, line 66:
```suggestion
		
```

x/observer/keeper/grpc_query_ballot_test.go
```diff
@@ -0,0 +1,136 @@
...
+	})
+
+	t.Run("should return ballot if exists", func(t *testing.T) {
+		k, ctx, _, _ := keepertest.ObserverKeeper(t)
+		wctx := sdk.WrapSDKContext(ctx)
+
+		voter := sample.AccAddress()
+		ballot := types.Ballot{
+			Index:            "index",
+			BallotIdentifier: "index",
```
kingpinXD, line 113: NIT : we should remove the index field, does not have to be in this PR though

x/observer/keeper/hooks_test.go
```diff
@@ -0,0 +1,211 @@
...
+	require.NoError(t, err)
+	os := types.ObserverSet{
+		ObserverList: []string{accAddress.String()},
+	}
+	k.SetObserverSet(ctx, os)
+
+	hooks := k.Hooks()
+	hooks.AfterValidatorRemoved(ctx, nil, valAddr)
+
+	os, found := k.GetObserverSet(ctx)
```
kingpinXD, line 43: maybe we can add some comments here

x/observer/keeper/hooks_test.go
```diff
@@ -0,0 +1,211 @@
...
+	valAddr := sample.ValAddress(r)
+	accAddress, err := types.GetAccAddressFromOperatorAddress(valAddr.String())
+	require.NoError(t, err)
+	os := types.ObserverSet{
+		ObserverList: []string{accAddress.String()},
+	}
+	k.SetObserverSet(ctx, os)
+
+	hooks := k.Hooks()
+	hooks.AfterValidatorRemoved(ctx, nil, valAddr)
```
kingpinXD, line 41: NIT : maybe we can check the error here ?

x/observer/keeper/hooks_test.go
```diff
@@ -0,0 +1,211 @@
...
+		ValidatorAddress: validator.GetOperator().String(),
+		Shares:           sdk.NewDec(10),
+	})
+
+	k.SetObserverSet(ctx, types.ObserverSet{
+		ObserverList: []string{accAddressOfValidator.String()},
+	})
+
+	hooks := k.Hooks()
+	hooks.AfterValidatorBeginUnbonding(ctx, nil, validator.GetOperator())
```
kingpinXD, line 69: same as above , maybe check the error ?

kingpinXD (commented):
Looks great , left some comments please have a look. 

Not part of the PR but some thoughts I had while reviewing
- The validation logic for maturity blocks should probably check that the number is positive
- When we remove the observer for the observer set we should probably pause the outbound, or at least have a way to notify that the observer set is modified. When it gets modified it would mean that the observer can no longer sign inbounds but still need to be part of the TSS sign for outbounds, which is weird?

x/observer/keeper/ballot_test.go
kingpinXD: Wondering if we should still add this , I meant setting a high value for 
`maturityBlocks := k.GetParamsIfExists(ctx).BallotMaturityBlocks`

kingpinXD:
> > Looks great , left some comments please have a look.
> > Not part of the PR but some thoughts I had while reviewing
> > 
> > * The validation logic for maturity blocks should probably check that the number is positive
> > * When we remove the observer for the observer set we should probably pause the outbound, or at least have a way to notify that the observer set is modified. When it gets modified it would mean that the observer can no longer sign inbounds but still need to be part of the TSS sign for outbounds, which is weird?
> 
> @kingpinXD thanks for review, i addressed your comments, please recheck when you get a chance
> 
> regarding your suggestions, including this one [#1961 (comment)](https://github.com/zeta-chain/node/pull/1961#discussion_r1548175398) do you mind opening issues for it, since you have more context?

Added the issues here dc1a85e6ca6b9521b0520ede8a6668fe8e53cf01

I don't think the second point I mentioned needs to be fixed , this part of the logic should handle that 
https://github.com/zeta-chain/zeta-node/blob/c7bf3d81ba6af3ed051d5158d06a55230374b0eb/x/observer/abci.go#L34-L35

kingpinXD (approved):
Looks good , unresolved one of the comments , please have a look before merging

Issue #1926, opened 2024-03-21
Rename `CreateTSSVoter` to `VoteTSSCreation`, move to `observer` and add tests


lumtis:
Every TSS related logic should be in the `observer module`

PR #1966, opened 2024-04-01
refactor: move TSS vote message from `crosschain` to `observer`

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/1926


x/observer/client/cli/tx_vote_tss.go
```diff
@@ -0,0 +1,53 @@
...
+				return err
+			}
+			var status chains.ReceiveStatus
+			if args[2] == "0" {
+				status = chains.ReceiveStatus_Success
+			} else if args[2] == "1" {
+				status = chains.ReceiveStatus_Failed
+			} else {
+				return fmt.Errorf("wrong status")
+			}
```
kingpinXD, line 38: maybe we can move this to a separate function now that we are already doing some refactor

lumtis: 263638bb

PR #1972, opened 2024-04-02
ci: trigger changelog check on adding/removing a label

fadeev:


fadeev:
Forgot we're not using `main`.

PR #1973, opened 2024-04-02
ci: trigger changelog check on adding/removing a label

fadeev:
I think it's useful, because you can skip changelog check by adding a label.


lumtis (approved):
Thanks for fixing this, this is what we wanted originally

CharlieMc0 (approved):

Issue #1417, opened 2023-11-20
zetaclientd support for multiple RPCs per chain w/ failover


CharlieMc0:
Each zetaclientd node needs a reliable RPC for each external chain and if that RPC fails it needs to connect a new one. If zetaclient can support multiple RPCs per chain and support automatic failover operators don't need to build their own highly available centralized solutions. 

Since the external chain connectivity is critical for ongoing operations it would be good to offer failover capabilities in an easy and decentralized way.

ws4charlie:
To support multiple RPCs per chain for failover, zetaclient needs to be able to measure the health status of each endpoint and switch to a healthy one when the current one is unhealthy.



**Why health status measurement is needed?**

- Current RPC endpoint can be unhealthy (e.g., down, ran out of entitled requests, etc.).
- `zetaclient` needs a number/threshold of RPC health index to determine if RPC rotation is necessary. RPC rotation should not be triggered frequently due to a few RPC failures.
- When rotation is needed, `zetaclient` needs to decide which one of other RPC endpoints is good to switch to. RPC rotation should not be done blindly




**How to measure the health of each RPC endpoint?**

1. For each external chain, zetaclient creates a dedicated `RPCHealthCheck` goroutine.
2. `RPCHealthCheck` goroutine does basic health checks (get block number, block header, gas price etc.) against each RPC endpoint candidate on tickers. The `ticker` is configurable, e.g., every 15 seconds.
3. `RPCHealthCheck` goroutine maintains a history of health check results for each RPC endpoint candidate. The `history length` is configurable, e.g., the last 5 minutes (20 tickers).
4. `RPCHealthCheck` goroutine defines a `RPCRotationHealthCheckFailRate` to indicate the minimum percentage of failed health checks in the history to trigger a RPC rotation. The `RPCRotationHealthCheckFailRate` is configurable, e.g., 40%.
5. When `RPCRotationHealthCheckFailRate` is reached, `zetaclient` finds the healthiest RPC endpoint candidate `R` (other than the current one) and switches to endpoint `R` only if `R` is healthier than the current one.



**How to feed `zetaclient` with multiple RPC endpoints?**
`zetaclient` runner needs to fill in the list of RPC endpoints in config file and then restarts `zetaclientd`.



**Which endpoint should be used when `zetaclientd` starts?**
[...]

PR #1998, opened 2024-04-08
feat: zetaclient - multiple rpc endpoints 

kevinssgh:
# Description

* Added multiple endpoint capability to zetaclient through config
* Implemented failover logic for each interface to external chain client
* Implemented simple client queue to manage multiple clients.

Closes: https://github.com/zeta-chain/node/issues/1417

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/zetaclient/common/client_queue.go
+++ b/zetaclient/common/client_queue.go
@@ -1,0 +1,34 @@
+package common
+
+type ClientQueue struct {
+	clientList []any
+}
+
+func NewClientQueue() *ClientQueue {
+	return &ClientQueue{
+		clientList: []any{},
+	}
+}
+
+func (c *ClientQueue) Append(item any) {
+	c.clientList = append(c.clientList, item)
+}
+
+func (c *ClientQueue) Length() int {
+	return len(c.clientList)
+}
+
+func (c *ClientQueue) Next() {
+	if len(c.clientList) < 1 {
+		return
+	}
+	// Rotate client order by 1
+	c.clientList = append(c.clientList[1:], c.clientList[:1]...)
+}
+
+func (c *ClientQueue) First() any {
+	if len(c.clientList) < 1 {
+		return nil
+	}
+	return c.clientList[0]
+}
--- a/zetaclient/evm/evm_rpc_fallback.go
+++ b/zetaclient/evm/evm_rpc_fallback.go
@@ -1,0 +1,387 @@
+package evm
+
+import (
+	"context"
+	"errors"
+	"github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/onrik/ethrpc"
+	"github.com/rs/zerolog"
+	"github.com/zeta-chain/zetacore/zetaclient/common"
+	"github.com/zeta-chain/zetacore/zetaclient/config"
+	"github.com/zeta-chain/zetacore/zetaclient/interfaces"
+	"math/big"
+)
+
+// EthClientFallbackInterface consolidates interfaces to external chain clients
+type EthClientFallbackInterface interface {
+	// EVMRPCClient EVMJSONRPCClient - Need to implement both interfaces to support newer data types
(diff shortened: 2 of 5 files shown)


zetaclient/common/client_queue.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Length - returns total length of queue or number of clients
+func (c *ClientQueue) Length() int {
+	return len(c.clientList)
+}
+
+// Next - rotates list of clients by moving the first to the last
+func (c *ClientQueue) Next() {
+	if len(c.clientList) < 1 {
```
ws4charlie, line 27: Is it a good idea move the check to `NewClientQueue` and throw error if empty list is provided on queue creation?

zetaclient/evm/evm_rpc_fallback.go
```diff
@@ -0,0 +1,398 @@
...
+	ethClientFallback := EthClientFallback{}
+	ethClientFallback.ethClients = common.NewClientQueue()
+	ethClientFallback.jsonRPCClients = common.NewClientQueue()
+
+	// Initialize clients
+	for _, endpoint := range evmCfg.Endpoints {
+		//Initialize go-ethereum clients
+		client, err := ethclient.Dial(endpoint)
+		if err != nil {
+			logger.Error().Err(err).Msg("eth Client Dial")
```
ws4charlie, line 44: I'm think a scenario when an observer runner provides 2 endpoints provided in the EVM config and 1/2 endpoints is not working due to some reason. In order to run/restart `zetaclient`, the operator has to manually identify and remove the glitchy endpoint from EVM config to avoid the error returned by `NewEthClientFallback`. Would it be a good idea to let `NewEthClientFallback` 

- check each endpoint in the list
- use those good ones to initialize `EthClientFallback` instance
- print an error message for identified bad endpoints
- if all endpoints are bad, return error.

zetaclient/evm/evm_rpc_fallback.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+func (e *EthClientFallback) TransactionByHash(ctx context.Context, hash ethcommon.Hash) (tx *ethtypes.Transaction, isPending bool, err error) {
+	for i := 0; i < e.ethClients.Length(); i++ {
+		if client := e.ethClients.First(); client != nil {
+			tx, isPending, err = client.(interfaces.EVMRPCClient).TransactionByHash(ctx, hash)
+		}
+		if err != nil {
+			e.logger.Debug().Err(err).Msg("client endpoint failed attempting fallback client")
+			e.ethClients.Next()
```
ws4charlie, line 301: Seems the rotation happens on every error of the `First` endpoint. It gives me a feeling that the underlying endpoint rotation can happen often and sometimes unexpectly across invocations of this `Next` method. The reason is that some errors from RPC call are normal and expected, for example:
- `TransactionByHash` can return 'Not Found' when non-existing `hash` is provided (can be a outbound tx in mempool waiting for inclusion).
- `TransactionReceipt` can return 'Not Found' when tx is waiting for inclusion.
- `TransactionSender` usually return errors on Goerli testnet.
- `SendTransaction` can return error on nonce mismatch, sending duplicate tx, etc.
It's hard to enumerate all RPC errors across RPC providers and to tell which ones should trigger rotation. Is it a good idea to reuse and repurpose the RPC status monitoring thread `WatchRPCStatus` for RPC rotation goal. 

There can be different options regarding impl. A quick impl in my mind looks like:
- `WatchRPCStatus` does the basic check as it is doing now in the thread.
-  Reduce the ticker of RPC status check a bit to a shorter period, e.g., `60 secs` --> `15 secs`
-  Define a `RPCRotationFailRateThrehold` to indicate the max percentage of failed RPC status checks in a health check time window. e.g., 30%. Also, define a `RPCHealthCheckWindow` to, e.g., 10 tickers.
-  If the status check fail rate exceeds the threhold (g.g., > 30%) in the latest time window, zetaclient then trigger RPC rotation.

ws4charlie:
I'll close this PR according to the discussions under: https://github.com/zeta-chain/node/issues/1417
The infrastructure team said that it would be a better solution to do it outside the zetaclient using balancer.

Issue #848, opened 2023-07-26
Removing the `params` module


lumtis:
Since Cosmos SDK `0.46`, the `params` module has been deprecated in favor of having each module managing their params in their own store: https://docs.cosmos.network/main/migrations/upgrading#xparams-1

This issue is about removing the params module import from the app codebase and moving params for each module in their own store under the `/params` prefix

lumtis:
Completed https://github.com/zeta-chain/node/pull/2014

Issue #1898, opened 2024-03-14
Enable `ObserverSlashAmount` again in the param


lumtis:
**Describe the Issue**
`ObserverSlashAmount` has been disabled following v14 bug. The parameter should be enabled back

PR #2001, opened 2024-04-09
refactor: replace broadcast mode block with sync and remove fungible params

skosito:
# Description

Part of cosmos sdk update (block mode is deprecated in 0.47) , but also generally better to use sync mode.
Also, removing unused fungible params to make upgrade easier.

Closes: #1930 

## Type of change

Tests

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

- [ ] Tested CCTX in localnet
- [x] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [x] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

PR #2014, opened 2024-04-11
refactor: deprecate x/params

skosito:
# Description

Based on this: https://github.com/cosmos/cosmos-sdk/pull/12363/files

Done for emissions module, observer params are removed and BallotMaturityBlocks param is moved from observer to emissions module. Also, introducing ObserverSlashAmount as param in this PR in migration.
This is still not upgrading cosmos.

I tested this upgrade locally using develop as old version and `make start-upgrade-test`.

Closes: #848  https://github.com/zeta-chain/node/issues/1898

## Type of change

Refactor

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

# Checklist:

### DIFF
--- a/x/emissions/keeper/msg_server_update_params.go
+++ b/x/emissions/keeper/msg_server_update_params.go
@@ -9,6 +9,8 @@ import (
 	"github.com/zeta-chain/zetacore/x/emissions/types"
 )
 
+// UpdateParams defines a governance operation for updating the x/emissions module parameters.
+// The authority is hard-coded to the x/gov module account.
 func (k msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
 	if msg.Authority != k.authority {
 		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", k.authority, msg.Authority)
--- a/x/emissions/keeper/params.go
+++ b/x/emissions/keeper/params.go
@@ -12,7 +12,11 @@ func (k Keeper) GetParams(ctx sdk.Context) (params types.Params, found bool) {
 	if bz == nil {
 		return types.Params{}, false
 	}
-	k.cdc.MustUnmarshal(bz, &params)
+	err := k.cdc.Unmarshal(bz, &params)
+	if err != nil {
+		return types.Params{}, false
+	}
+
 	return params, true
 }
 
@@ -23,7 +27,11 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) error {
 	}
 
 	store := ctx.KVStore(k.storeKey)
-	bz := k.cdc.MustMarshal(&params)
+	bz, err := k.cdc.Marshal(&params)
+	if err != nil {
+		return err
+	}
+
 	store.Set(types.KeyPrefix(types.ParamsKey), bz)
 	return nil
 }
--- a/docs/spec/emissions/messages.md
+++ b/docs/spec/emissions/messages.md
@@ -2,6 +2,9 @@
 
 ## MsgUpdateParams
 
+UpdateParams defines a governance operation for updating the x/emissions module parameters.
+The authority is hard-coded to the x/gov module account.
+
 ```proto
 message MsgUpdateParams {
 	string authority = 1;


lumtis:
@skosito I think we can try to merge https://github.com/zeta-chain/node/pull/2001 that doesn't contains sensitive change and create a sdk47 feature branch for this one

changelog.md
```diff
@@ -30,6 +30,7 @@
 * [1966](https://github.com/zeta-chain/node/pull/1966) - move TSS vote message from crosschain to observer
 * [1853](https://github.com/zeta-chain/node/pull/1853) - refactor vote inbound tx and vote outbound tx
 * [2001](https://github.com/zeta-chain/node/pull/2001) - replace broadcast mode block with sync and remove fungible params
+* [2014](https://github.com/zeta-chain/node/pull/2014) - remove params module
```
lumtis, line 33: Will be after the next release
Maybe we can already add another unrelease section in the top to not forget about putting this one here

proto/observer/tx.proto
```diff
@@ -113,3 +114,10 @@ message MsgVoteTSSResponse {
   bool vote_finalized = 2;
   bool keygen_success = 3;
 }
+
+message MsgUpdateParams {
```
lumtis, line 118: Only `BallotMaturityBlocks` is still used in `observer` which can be moved to `emissions`

What about already moving it in the emissions module in this PR?

x/emissions/keeper/params.go
```diff
@@ -5,14 +5,25 @@ import (
...
-// non existent parameters will return zero values
-func (k Keeper) GetParamsIfExists(ctx sdk.Context) (params types.Params) {
-	k.paramStore.GetParamSetIfExists(ctx, &params)
-	return
+// GetParams get all parameters
+func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.KeyPrefix(types.ParamsKey))
+	if bz == nil {
+		return params
```
lumtis, line 13: Shouldn't we instead have a found boolean returned in this case?

x/emissions/abci.go
```diff
@@ -75,16 +71,24 @@ func DistributeObserverRewards(
 	ctx sdk.Context,
 	amount sdkmath.Int,
 	keeper keeper.Keeper,
-	slashAmount sdkmath.Int,
 ) error {
-
+	params, found := keeper.GetParams(ctx)
+	if !found {
+		return errors.New("params not found")
```
lumtis, line 77: Maybe we can provide the param object in argument since it is used in the calling function

Also we should add test case here if we keep it this way

x/emissions/abci.go
```diff
@@ -75,16 +71,24 @@ func DistributeObserverRewards(
...
 	if err != nil {
 		return err
 	}
-	ballotIdentifiers := keeper.GetObserverKeeper().GetMaturedBallotList(ctx)
+	maturityBlocks := params.BallotMaturityBlocks
+	list, found := keeper.GetObserverKeeper().GetBallotList(ctx, ctx.BlockHeight()-maturityBlocks)
+	ballotIdentifiers := []string{}
+	if found {
+		ballotIdentifiers = list.BallotsIndexList
+	}
```
lumtis, line 91:
```suggestion
	}
	
```

x/emissions/abci.go
```diff
@@ -75,16 +71,24 @@ func DistributeObserverRewards(
...
+	if !found {
+		return errors.New("params not found")
+	}
+	slashAmount := params.ObserverSlashAmount
 	rewardsDistributer := map[string]int64{}
 	totalRewardsUnits := int64(0)
 	err := keeper.GetBankKeeper().SendCoinsFromModuleToModule(ctx, types.ModuleName, types.UndistributedObserverRewardsPool, sdk.NewCoins(sdk.NewCoin(config.BaseDenom, amount)))
 	if err != nil {
 		return err
 	}
```
lumtis, line 85:
```suggestion
	}
	
```

x/emissions/abci.go
```diff
@@ -75,16 +71,24 @@ func DistributeObserverRewards(
 	ctx sdk.Context,
 	amount sdkmath.Int,
 	keeper keeper.Keeper,
-	slashAmount sdkmath.Int,
 ) error {
-
+	params, found := keeper.GetParams(ctx)
+	if !found {
+		return errors.New("params not found")
+	}
```
lumtis, line 78:
```suggestion
	}
	
```

x/emissions/genesis.go
```diff
@@ -19,7 +23,7 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 // ExportGenesis returns the emissions module's exported genesis.
 func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	var genesis types.GenesisState
-	genesis.Params = k.GetParamsIfExists(ctx)
+	genesis.Params, _ = k.GetParams(ctx)
```
lumtis, line 26: I would suggest to manually set empty params if not found
We should assume to not use the return params value if found is false

x/emissions/keeper/block_rewards_components.go
```diff
@@ -18,9 +18,14 @@ func (k Keeper) GetBlockRewardComponents(ctx sdk.Context) (sdk.Dec, sdk.Dec, sdk
 }
 
 func (k Keeper) GetBondFactor(ctx sdk.Context, stakingKeeper types.StakingKeeper) sdk.Dec {
-	targetBondRatio := sdk.MustNewDecFromStr(k.GetParamsIfExists(ctx).TargetBondRatio)
-	maxBondFactor := sdk.MustNewDecFromStr(k.GetParamsIfExists(ctx).MaxBondFactor)
-	minBondFactor := sdk.MustNewDecFromStr(k.GetParamsIfExists(ctx).MinBondFactor)
+	params, found := k.GetParams(ctx)
+	if !found {
+		// TODO: what to return here?
```
lumtis, line 23: This is same as above comment (and for other situation where this function is used)

I suggest giving the params as argument, so we just handle params not existing at the top level

Also since reading the state would no longer be necessary, we could move this logic in the `types` packages and makes the code more functional paradigm

x/emissions/keeper/msg_server_update_params.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (k msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
+	if msg.Authority != k.authority {
+		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", k.authority, msg.Authority)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	err := k.SetParams(ctx, msg.Params)
+	if err != nil {
+		return nil, err
```
lumtis, line 20: I think a good practice is to use wrapped error for all errors at the top level of the message for better assertion in test

x/emissions/keeper/msg_server_update_params_test.go
```diff
@@ -0,0 +1,56 @@
...
+		k, ctx, _, _ := keepertest.EmissionsKeeper(t)
+		msgServer := keeper.NewMsgServerImpl(*k)
+		params := types.DefaultParams()
+		params.ValidatorEmissionPercentage = "-1.5"
+		res, err := msgServer.UpdateParams(ctx, &types.MsgUpdateParams{
+			Authority: k.GetAuthority(),
+			Params:    params,
+		})
+
+		require.Error(t, err)
```
lumtis, line 53: I suggest using `ErrorIs` for better assertion

x/emissions/keeper/msg_server_update_params_test.go
```diff
@@ -0,0 +1,56 @@
...
+		require.Nil(t, res)
```
lumtis, line 54: NIT: I don't think we need to assert this value, it should be assume res would not be used in any case if an error is returned

x/emissions/keeper/params.go
```diff
@@ -5,14 +5,25 @@ import (
...
-// non existent parameters will return zero values
-func (k Keeper) GetParamsIfExists(ctx sdk.Context) (params types.Params) {
-	k.paramStore.GetParamSetIfExists(ctx, &params)
-	return
+// GetParams get all parameters
+func (k Keeper) GetParams(ctx sdk.Context) (params types.Params, found bool) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.KeyPrefix(types.ParamsKey))
+	if bz == nil {
+		return types.Params{}, false
```
lumtis, line 13: Can we have a test case for this one?

x/emissions/keeper/params_test.go
```diff
@@ -207,32 +221,23 @@ func TestKeeper_GetParams(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			k, ctx, _, _ := keepertest.EmissionsKeeper(t)
-			assertPanic(t, func() {
-				k.SetParams(ctx, tt.params)
-			}, tt.isPanic)
-
-			if tt.isPanic != "" {
-				require.Equal(t, emissionstypes.DefaultParams(), k.GetParamsIfExists(ctx))
+			err := k.SetParams(ctx, tt.params)
+			if tt.isErr != "" {
```
lumtis, line 233: IsErr has a content but not used. Can we use `ErrorContains`?

x/emissions/migrations/v3/migrate.go
```diff
@@ -0,0 +1,35 @@
...
+// module state.
+func MigrateStore(
+	ctx sdk.Context,
+	emissionsKeeper EmissionsKeeper,
+	legacySubspace exported.Subspace,
+) error {
+	var currParams types.Params
+	legacySubspace.GetParamSet(ctx, &currParams)
+
+	currParams.ObserverSlashAmount = sdkmath.NewInt(100000000000000000)
```
lumtis, line 27: I think we can define a variable above and explain why we use this value

x/emissions/types/params.go
```diff
@@ -41,27 +30,45 @@ func DefaultParams() Params {
...
 // Validate validates the set of params
 func (p Params) Validate() error {
-	return nil
+	err := validateMaxBondFactor(p.MaxBondFactor)
+	if err != nil {
+		return err
+	}
+	err = validateMinBondFactor(p.MinBondFactor)
+	if err != nil {
+		return err
```
lumtis, line 41: Not covered by tests

x/emissions/types/params.go
```diff
@@ -41,27 +30,45 @@ func DefaultParams() Params {
...
+	}
+	err = validateTssEmissionPercentage(p.TssSignerEmissionPercentage)
+	if err != nil {
+		return err
+	}
+	err = validateDurationFactorConstant(p.DurationFactorConstant)
+	if err != nil {
+		return err
```
lumtis, line 65: Not covered by tests

x/emissions/abci.go
```diff
@@ -75,16 +71,23 @@ func DistributeObserverRewards(
...
 	rewardsDistributer := map[string]int64{}
 	totalRewardsUnits := int64(0)
 	err := keeper.GetBankKeeper().SendCoinsFromModuleToModule(ctx, types.ModuleName, types.UndistributedObserverRewardsPool, sdk.NewCoins(sdk.NewCoin(config.BaseDenom, amount)))
 	if err != nil {
 		return err
 	}
-	ballotIdentifiers := keeper.GetObserverKeeper().GetMaturedBallotList(ctx)
+
+	maturityBlocks := params.BallotMaturityBlocks
+	list, found := keeper.GetObserverKeeper().GetBallotList(ctx, ctx.BlockHeight()-maturityBlocks)
```
lumtis, line 85: I think it would make beter encapsulation to use a method
```
ObserverKeeper.GetMaturedBallots(ctx, maturityBlocks)
```

lumtis (approved):
Looks good to me, just a minor refactoring that would be better I think

x/emissions/keeper/params.go
```diff
@@ -5,14 +5,25 @@ import (
...
 // SetParams set the params
-func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
-	k.paramStore.SetParamSet(ctx, &params)
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) error {
+	if err := params.Validate(); err != nil {
+		return err
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	bz := k.cdc.MustMarshal(&params)
```
kingpinXD, line 26: Can we use `Marshal` and return the error here?

x/emissions/keeper/params.go
```diff
@@ -5,14 +5,25 @@ import (
...
-	k.paramStore.GetParamSetIfExists(ctx, &params)
-	return
+// GetParams get all parameters
+func (k Keeper) GetParams(ctx sdk.Context) (params types.Params, found bool) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.KeyPrefix(types.ParamsKey))
+	if bz == nil {
+		return types.Params{}, false
+	}
+	k.cdc.MustUnmarshal(bz, &params)
```
kingpinXD, line 15: Same for this , we can use `Unmarshal` and return error

x/emissions/types/params.go
```diff
@@ -41,27 +30,80 @@ func DefaultParams() Params {
...
+	}
+	err = validateTssEmissionPercentage(p.TssSignerEmissionPercentage)
+	if err != nil {
+		return err
+	}
+	err = validateBallotMaturityBlocks(p.BallotMaturityBlocks)
+	if err != nil {
+		return err
+	}
+	return validateObserverSlashAmount(p.ObserverSlashAmount)
```
kingpinXD, line 67: Is there any value in keeping these interface-based methods? or can we just simply it using just the conditions for the values

x/emissions/migrations/v3/migrate_test.go
```diff
@@ -0,0 +1,72 @@
...
+			TargetBondRatio:             "00.50",
+			ValidatorEmissionPercentage: "00.50",
+			ObserverEmissionPercentage:  "00.35",
+			TssSignerEmissionPercentage: "00.15",
+			DurationFactorConstant:      "0.001877876953694702",
+			ObserverSlashAmount:         sdk.ZeroInt(),
+		}
+		legacySubspace := newMockSubspace(legacyParams)
+
+		require.Error(t, v3.MigrateStore(ctx, k, legacySubspace))
```
kingpinXD, line 70: Maybe we can use `ErrorIs`  and check for the error from the validate function ?

x/emissions/migrations/v3/migrate_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+func newMockSubspace(ps types.Params) mockSubspace {
+	return mockSubspace{ps: ps}
+}
+
+func (ms mockSubspace) GetParamSet(ctx sdk.Context, ps exported.ParamSet) {
+	*ps.(*types.Params) = ms.ps
+}
+
+func TestMigrate(t *testing.T) {
```
kingpinXD, line 28: NIT : Can we also add a test in which some values are missing in `legacyParams` .

x/emissions/migrations/v3/migrate.go
```diff
@@ -0,0 +1,43 @@
...
+	SetParams(ctx sdk.Context, params types.Params) error
+}
+
+// ObserverSlashAmount is the amount of tokens to be slashed from observer in case of incorrect vote
+// by default it is set to 0.1 ZETA
+var observerSlashAmountDefaultValue = sdkmath.NewInt(100000000000000000)
+
+// BallotMaturityBlocks is amount of blocks needed for ballot to mature
+// by default is set to 100
+var ballotMaturityBlocksDefaultValue = 100
```
kingpinXD, line 21: Can we put this in the types package, in keys.go

x/emissions/keeper/msg_server_update_params.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	"github.com/zeta-chain/zetacore/x/emissions/types"
+)
+
+func (k msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
```
kingpinXD, line 12: It might be a good idea to add some comments, specially around who the expectected authority is

x/emissions/keeper/msg_server_update_params.go
```diff
@@ -0,0 +1,24 @@
...
+	if msg.Authority != k.authority {
+		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", k.authority, msg.Authority)
+	}
```
kingpinXD, line 15: I am wondering should make this to be 
either authority or emergency group policy ? 

Would be good to get thoughts from @lumtis  as well.

kingpinXD (commented):
minor changes requested , and one point of discussion
https://github.com/zeta-chain/node/pull/2014#discussion_r1575242927

x/emissions/keeper/msg_server_update_params.go
kingpinXD: yeah , I think the authority approach is correct one , but I was wondering if it might be a good idea to add emergency policy to change this as well

x/emissions/keeper/params.go
kingpinXD: I think for set params, we are returning err already, 
for get, can we try returning false on error, I guess handling the error is slightly safer than panicking, wdyt?

kingpinXD (approved):

Issue #1928, opened 2024-03-21
Upgrade Cosmos SDK to `v0.47.10`


lumtis:
This epic lists the set of tasks to complete in order to upgrade the blockchain to using [v0.47.10](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.47.10) (latest 47.x version)

It follows these steps: https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md
And some other various tasks

```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/848
- [ ] https://github.com/zeta-chain/node/issues/1929
- [ ] https://github.com/zeta-chain/node/issues/1930
- [ ] https://github.com/zeta-chain/node/issues/1931
- [ ] https://github.com/zeta-chain/node/issues/1932
- [ ] https://github.com/zeta-chain/node/issues/2065
- [ ] https://github.com/zeta-chain/node/issues/2091
- [ ] https://github.com/zeta-chain/node/issues/2120
- [ ] https://github.com/zeta-chain/node/issues/2121
```

gartnera:
Reopening due to upgrade issues

gartnera:
Upgrade issues appear to be resolved

PR #2039, opened 2024-04-17
feat: cosmos v0.47 upgrade

skosito:
# Description

NOTE: a lot of files changed in the diff, but most are cli docs md files changes from auto generation, and things like moving files and changing imports, so PR is not that big.

Ethermint commit used here is in this PR: https://github.com/zeta-chain/ethermint/pull/2

Closes all issues from epic: https://github.com/zeta-chain/node/issues/1928

Going through this doc: https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#migration-to-cometbft-part-1

Migration to comet bft is done, but tendermint dependency is still there because go-tss is using it in zetaclient, but upgrade doc suggest to remove it everywhere. Can we migrate in go-tss too?

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#simulation

RandomizedParams is cleaned up and empty implementation of ProposalMsgs is added (even though not needed).

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#grpc

This is added in app.go RegisterAPIRoutes.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#appmodule-interface

These methods are removed.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#simapp
- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#encoding

Updated, also network tests are a bit updated with changes from cosmos 0.47 codebase.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#app-wiring

Not sure app wiring is used, so this is not relevant.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#constructor

Constructor is simplified, and also base app opts method is used, overriding mempool like this:
```
baseappOptions := server.DefaultBaseappOptions(appOpts)
	baseappOptions = append(baseappOptions, func(app *baseapp.BaseApp) {
		app.SetMempool(mempool.NoOpMempool{})
	})
```

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#export

Added modulesToExport argument as suggested.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#replaces

This was already done.

Proto updates: 
- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#replaces

This is updated, and also using docker as suggested in makefile for proto gen.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#gogoproto-import-paths

Added `zetachain/zetacore` subdirectories to proto folder to be inline with this.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#acceptsimplements_interface-proto-annotations

Nothing needed here.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#broadcast-mode

Done in separate PR: https://github.com/zeta-chain/node/pull/2001

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#all

This is not changed, as suggested when module name is not second in message route.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#xparams

Done in separate PR: https://github.com/zeta-chain/node/pull/2014 , but here relevant handler upgrade is added.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#xgov

No code changes needed here it seems.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#new-proposalproposer-field

Optional migration for existing proposals, skipped at the moment.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#xconsensus

Added to upgrade handler.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#appgo-changes

This is updated.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#xnft

Not relevant.

- https://github.com/cosmos/cosmos-sdk/blob/v0.47.10/UPGRADING.md#ledger

Looks not relevant.
Closes: <PD-XXXX>

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Important: tested also with upgrade e2e tests, by using current develop as `OLD_VERSION`.

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -357,4 +357,4 @@ replace (
 
 replace github.com/cometbft/cometbft-db => github.com/notional-labs/cometbft-db v0.0.0-20230321185329-6dc7c0ca6345
 
-replace github.com/evmos/ethermint => github.com/zeta-chain/ethermint v0.0.0-20240426125617-308e6f08a922
+replace github.com/evmos/ethermint => github.com/zeta-chain/ethermint v0.0.0-20240429123701-35f3f79bf83f
--- a/go.sum
+++ b/go.sum
@@ -1769,8 +1769,8 @@ github.com/yuin/goldmark v1.1.32/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9de
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
 github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=
-github.com/zeta-chain/ethermint v0.0.0-20240426125617-308e6f08a922 h1:U4+71OFfTYcOLL4OvZF8VDHJwopOnG7Z/8nD2VSKyMw=
-github.com/zeta-chain/ethermint v0.0.0-20240426125617-308e6f08a922/go.mod h1:s1zA6OpXv3Tb5I0M6M6j5fo/AssaZL/pgkc7G0W2kN8=
+github.com/zeta-chain/ethermint v0.0.0-20240429123701-35f3f79bf83f h1:joafCsPgohPEn93VCbNXi9IAl6kNvKy8u+kv5amEvUk=
+github.com/zeta-chain/ethermint v0.0.0-20240429123701-35f3f79bf83f/go.mod h1:s1zA6OpXv3Tb5I0M6M6j5fo/AssaZL/pgkc7G0W2kN8=
 github.com/zeta-chain/go-tss v0.1.1-0.20240208222330-f3be0d4a0d98 h1:GCSRgszQbAR7h/qK0YKjlm1mcnZOaGMbztRLaAfoOx0=
 github.com/zeta-chain/go-tss v0.1.1-0.20240208222330-f3be0d4a0d98/go.mod h1:+lJfk/qqt+oxXeVuJV+PzpUoxftUfoTRf2eF3qlbyFI=
 github.com/zeta-chain/keystone/keys v0.0.0-20231105174229-903bc9405da2 h1:gd2uE0X+ZbdFJ8DubxNqLbOVlCB12EgWdzSNRAR82tM=


app/export.go
```diff
@@ -30,7 +30,7 @@ func (app *App) ExportAppStateAndValidators(
 		app.prepForZeroHeightGenesis(ctx, jailAllowedAddrs)
 	}
 
-	genState := app.mm.ExportGenesis(ctx, app.appCodec)
+	genState := app.mm.ExportGenesisForModules(ctx, app.appCodec, modulesToExport)
```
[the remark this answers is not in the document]
kingpinXD: Nice that they added this change, this was exactly the functionality I wanted to implement for our export process , this would allow us to not export state data which we don't need to for our test . 
I think for a basic test  you could compare the exported genesis to an existing one , they should contain similar data , a older genesis data would be available at sample.GenDoc , if you rebase with develop. Not sure if we need to be writing tests for this though the ExportCMD is part of the SDK

app/setup_handlers.go
```diff
@@ -40,7 +82,7 @@ func SetupHandlers(app *App) {
 	}
 	if upgradeInfo.Name == releaseVersion && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
 		storeUpgrades := storetypes.StoreUpgrades{
-			Added: []string{authoritytypes.ModuleName, lightclienttypes.ModuleName},
+			Added: []string{authoritytypes.ModuleName, lightclienttypes.ModuleName, consensustypes.ModuleName, crisistypes.ModuleName},
```
kingpinXD, line 85: Should crisis be added here ?

app/setup_handlers.go
```diff
@@ -15,20 +26,51 @@ const releaseVersion = "v15"
...
+			}
 		}
+
+		VersionMigrator{v: vm}.TriggerMigration(authtypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(banktypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(stakingtypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(distrtypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(slashingtypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(govtypes.ModuleName)
+		VersionMigrator{v: vm}.TriggerMigration(crisistypes.ModuleName)
```
kingpinXD, line 72: are we only triggering migration scripts for modules with params ?

cmd/zetacored/root.go
```diff
@@ -231,59 +230,30 @@ func (ac appCreator) newApp(
...
 ) servertypes.Application {
-	var cache sdk.MultiStorePersistentCache
-
-	if cast.ToBool(appOpts.Get(server.FlagInterBlockCache)) {
-		cache = store.NewCommitKVStoreCacheManager()
-	}
-
+	baseappOptions := server.DefaultBaseappOptions(appOpts)
+	baseappOptions = append(baseappOptions, func(app *baseapp.BaseApp) {
+		app.SetMempool(mempool.NoOpMempool{})
```
kingpinXD, line 235: Would be great if you could add some comments around why we are selecting the NoOp mempool

app/setup_handlers.go
kingpinXD: Hmm , checked on develop the crisis module is already present . 
https://github.com/zeta-chain/zeta-node/blob/022ea57e25b4e0cb338b99bbd1c9e235f41ae954/app/app.go#L514-L514

Adding it again might result in a consensus version mismatch because the initial version has already been set . I am not sure though if there is a special reason for adding crisis again , the upgrade test might help when we run it

go.mod
```diff
@@ -58,41 +56,53 @@ require (
...
 require (
+	cosmossdk.io/simapp v0.0.0-20230608160436-666c345ad23d
+	cosmossdk.io/tools/rosetta v0.2.1
 	github.com/binance-chain/tss-lib v0.0.0-20201118045712-70b2cb4bf916
+	github.com/btcsuite/btcd/btcutil v1.1.3
+	github.com/cometbft/cometbft v0.37.4
+	github.com/cometbft/cometbft-db v0.8.0
 	github.com/nanmu42/etherscan-api v1.10.0
 	github.com/onrik/ethrpc v1.2.0
+	github.com/tendermint/tendermint v0.34.14
```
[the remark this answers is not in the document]
lumtis: Yes, let's create a tag in go-tss?

x/emissions/types/message_update_params_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/stretchr/testify/require"
+	"github.com/zeta-chain/zetacore/testutil/sample"
+	"github.com/zeta-chain/zetacore/x/emissions/types"
+)
+
+func TestMsgUpdateParams_ValidateBasic(t *testing.T) {
```
lumtis, line 13: We can also add a test with invalid params

x/emissions/types/params.go
```diff
@@ -41,27 +30,80 @@ func DefaultParams() Params {
...
-		paramtypes.NewParamSetPair(KeyPrefix(ParamAvgBlockTime), &p.AvgBlockTime, validateAvgBlockTime),
-		paramtypes.NewParamSetPair(KeyPrefix(ParamTargetBondRatio), &p.TargetBondRatio, validateTargetBondRatio),
-		paramtypes.NewParamSetPair(KeyPrefix(ParamValidatorEmissionPercentage), &p.ValidatorEmissionPercentage, validateValidatorEmissionPercentage),
-		paramtypes.NewParamSetPair(KeyPrefix(ParamObserverEmissionPercentage), &p.ObserverEmissionPercentage, validateObserverEmissionPercentage),
-		paramtypes.NewParamSetPair(KeyPrefix(ParamTssSignerEmissionPercentage), &p.TssSignerEmissionPercentage, validateTssEmissionPercentage),
-		paramtypes.NewParamSetPair(KeyPrefix(ParamDurationFactorConstant), &p.DurationFactorConstant, validateDurationFactorConstant),
+// Validate validates the set of params
+func (p Params) Validate() error {
+	err := validateMaxBondFactor(p.MaxBondFactor)
+	if err != nil {
```
lumtis, line 36:
```suggestion	
	if err := validateMaxBondFactor(p.MaxBondFactor); err != nil {
```
Not a strong opinion but I find it more visual doing it this way

x/emissions/types/params.go
```diff
@@ -179,3 +221,27 @@ func validateTssEmissionPercentage(i interface{}) error {
 	}
 	return nil
 }
+
+func validateObserverSlashAmount(i interface{}) error {
```
lumtis, line 225: I think we can use the true type in the argument. We used to have interface{} since generic data is retrieved from the param store but we now have type safety

x/emissions/types/params_legacy.go
```diff
@@ -0,0 +1,28 @@
+package types
```
lumtis, line 1: Can we give a note why we still need this file?

x/emissions/types/params_legacy.go
```diff
@@ -0,0 +1,28 @@
...
+		paramtypes.NewParamSetPair(KeyPrefix(ParamAvgBlockTime), &p.AvgBlockTime, validateAvgBlockTime),
+		paramtypes.NewParamSetPair(KeyPrefix(ParamTargetBondRatio), &p.TargetBondRatio, validateTargetBondRatio),
+		paramtypes.NewParamSetPair(KeyPrefix(ParamValidatorEmissionPercentage), &p.ValidatorEmissionPercentage, validateValidatorEmissionPercentage),
+		paramtypes.NewParamSetPair(KeyPrefix(ParamObserverEmissionPercentage), &p.ObserverEmissionPercentage, validateObserverEmissionPercentage),
+		paramtypes.NewParamSetPair(KeyPrefix(ParamTssSignerEmissionPercentage), &p.TssSignerEmissionPercentage, validateTssEmissionPercentage),
+		paramtypes.NewParamSetPair(KeyPrefix(ParamDurationFactorConstant), &p.DurationFactorConstant, validateDurationFactorConstant),
+
+		// TODO: enable this param
+		// https://github.com/zeta-chain/node/pull/1861
+		//paramtypes.NewParamSetPair(KeyPrefix(ParamObserverSlashAmount), &p.ObserverSlashAmount, validateObserverSlashAmount),
```
lumtis, line 26: I think this can be removed?

lumtis (commented):
Looks good, some small comments

lumtis (approved):

kingpinXD (approved):

Issue #2010, opened 2024-04-10
Advanced test CI pipeline seems to not fail if the underlying test fail


lumtis:
Case: https://github.com/zeta-chain/node/actions/runs/8632326588/job/23662742835

Is it an expected behavior?

cc @gzukel

PR #2068, opened 2024-04-23
test: ensure advanced tests fail if orchestrator fails

gartnera:
# Description

Both the admin tests and upgrade tests are currently failing on develop, but have a green check on github. Examples:
- https://github.com/zeta-chain/node/actions/runs/8740792255/job/23985274941
- https://github.com/zeta-chain/node/actions/runs/8789497964/job/24119390065

We need the github actions run to fail so we can properly track pass/fail rate and add slack notifications.

Ensure that we fail the github actions step by checking the docker container code via `docker wait`. Also:
- create an output group with the `make start-...` to reduce the noise
- stream the test log output
- remove `shell` option since we can tolerate `sh` again. By setting `shell` you clobber the `-e` option which means that errors will silently be ignored.

Example: https://github.com/zeta-chain/node/actions/runs/8804170791/job/24163953655

Will add slack alerting and node/client log dumps in a future PR.

# How Has This Been Tested?

### DIFF
--- a/.github/workflows/execute_advanced_tests.yaml
+++ b/.github/workflows/execute_advanced_tests.yaml
@@ -37,36 +37,17 @@ jobs:
       - name: "Checkout Code"
         uses: actions/checkout@v3
 
+      # use docker logs -f rather than docker attach to make sure we get the initial logs
       - name: Execute e2e-admin-tests
-        shell: bash
+        shell: bash -e
         run: |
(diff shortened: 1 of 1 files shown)


CharlieMc0 (approved):

lumtis (approved):
Thanks for addressing this

We actually had an issue for this problem: https://github.com/zeta-chain/node/issues/2010

PR #2081, opened 2024-04-25
docs: update release flow/docs

gartnera:
Update release docs to use `release/` branches. Please comment if you understand or have questions about this flow.

I haven't touched any of the release github workflows but they may need minor tweaks. The publish rc workflow reacts to a git tag push but the publish release workflow creates a tag during the workflow itself in reaction to the `workflow_dispatch`


readme.md
```diff
@@ -81,27 +81,54 @@ to [run the E2E test](./LOCAL_TESTING.md).
...
+
+```bash
+git fetch
+git checkout -b release/v15 origin/develop
+git push origin release/v15
+```
+
+</details>
+
+Most changes should first be merged into `develop` then backported to the release branch via a PR.
```
gartnera, line 99: It's generally good to backport the change to the release branch via a PR so you can make sure CI is still passing. 

I've previously used [this github action](https://github.com/marketplace/actions/backporting) to automatically generate a backport PR when the develop PR is merged. All you have to do is apply a `backport release/v15` label to your develop PR.

But we could also allow developers to just directly cherry pick and push their changes to the release branch and run the CI on push instead.

CharlieMc0 (approved):

readme.md
```diff
@@ -81,27 +81,54 @@ to [run the E2E test](./LOCAL_TESTING.md).
 [Discord](https://discord.com/invite/zetachain) |
 [Telegram](https://t.me/zetachainofficial) | [Website](https://zetachain.com)
 
+## Releases
 
-## Creating a Release Candidate
-Creating a release candidate for testing is a straightforward process. Here are the steps to follow:
+To start a new major release, begin by creating and pushing a `release/` branch based on `develop`.
```
lumtis, line 86: I think can also delete current `main` branch and rename `develop` into `main` since in our current flow we don't use main

lumtis (approved):

readme.md
```diff
@@ -81,27 +81,54 @@ to [run the E2E test](./LOCAL_TESTING.md).
...
+git cherry-pick <commit SHA from develop>
+git push origin my-backport-branch
+```
+
+</details>
+
+### Creating a Release Candidate
+Once a release branch has been created, you can use github actions to create a release candidate:
+1) Create the release candidate tag with the following format (e.g., vx.x.x-rc) ex. v11.0.0-rc.
+2) Push the tag and the automation will take care of the rest
```
kingpinXD, line 116: What do you think of creating multiple release candidates, such as rc1 and rc2, leading up to a release? 

From the workflow perspective, it would mean that, instead of backporting multiple changes onto RC1, we create RC2 instead.

We only create the `release/vXXX` branch once we are reasonably confident about the changes to minimize backporting changes into it. 

We can probably run the final RC on devnet or mock-mainnet for a few days, if needed, before cutting the release.

Overall, I want to avoid cherry-picking changes as much as possible and only do it if absolutely needed.

gartnera: @kingpinXD Creating RC tags directly off develop is fine so long as there is not content on develop that will not make it into the release branch. I will update the doc.

If we need to do RC tags for minor releases (`v16.0.1-rc1`), those would need to based on the release branch.

Issue #2083, opened 2024-04-26
Change `VerificationFlags` to hold a list ChainID->bool


lumtis:
**Describe the Issue**
Context: https://github.com/zeta-chain/node/pull/2071#discussion_r1579388951

The fact if a consensus is supported in the header proof system is a static info, either it is implemented or not.

The dynamic verification flag should map enabled chains with chain IDs

**Expected Outcome**
Change the `VerificationFlags` object:
```
VerificationFlags {
  EnabledChains {
    ChainID int64
    Enabled bool
  }
}
```

PR #2097, opened 2024-04-29
refactor: lightclient verification flags

kingpinXD:
# Description

- Refactor `MsgUpdateVerificationFlags` into `MsgEnableHeaderVerification` and `MsgDisableHeaderVerification` .
- Refactor Verification flags to use individual chain IDs instead of consensus mechanisms such as ETH /BTC 
- Rename Verifaction flags to `BlockHeaderVerification` which stores a list of chains

It does not add a migration script , as the flags can be enable once the upgrade rolls out 

Closes: https://github.com/zeta-chain/node/issues/2083

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


lumtis (commented):
> Refactor storage for Verification flags to use chain-id as index

Why adding this indexation instead of simply keep an array of chain IDs in a singleton?
```
VerificationFlags {
  ChainIDs []int64
}
```
It's easier to maintain, we can put most of the logic in `types` in pure function and easier to migrate
In general we should avoid adding indexation when necessary since this complexify the state. Here it would not be necessary, the rollout of chain header supported would be slow so the array would not continously grow unlike cctxs and it wouldn't be a performance issue to iterate the array to detect enabled chains

(the variable should also be renamed from flags since we don't have a a predifined list of values to set/unset but can be considered for another PR)

kingpinXD:
Having an index makes the update much easier, especially when we consider that the input is a list of chain IDs. 
From migration perspective I think both options are similar , we would either iterate before or after fetching the value.
In my opinion , maintaining the index is simpler and easier to manage in terms of get and set values. Let me know if you still strongly feel about keeping the singleton.

I agree regarding the names

lumtis:
> Having an index makes the update much easier, especially when we consider that the input is a list of chain IDs.
From migration perspective I think both options are similar , we would either iterate before or after fetching the value.
In my opinion , maintaining the index is simpler and easier to manage in terms of get and set values. Let me know if you still strongly feel about keeping the singleton.

Using a singleton simplifies the interaction and representation of the store with a single entrypoint. This structure is like parameters of a module, it's not a ever-growing array of data where indexing is critical, I see it as a simpler model to use a self-containing structure like we do for similar structures.
Updating might be more complex but is easy to encapsulate
```
list.Add(chainID)
list.Remove(chainID)
list.Exist(chainID)
```
The algo would be:
```
list := store.GetList()
for chainID := range chainIDs
  list.Add(chainID)
store.SaveList(list)
```

Finally the query would be simplified, the ZetaClient would directly get a list of chainIDs that are enabled instead of a list of VerificationFlag object that can contains false

pkg/chains/chain.go
```diff
@@ -101,8 +101,8 @@ func IsZetaChain(chainID int64) bool {
 	return ChainIDInChainList(chainID, ChainListByNetwork(Network_zeta))
 }
 
-// IsHeaderSupportedEvmChain returns true if the chain is an EVM chain supporting block header-based verification
-func IsHeaderSupportedEvmChain(chainID int64) bool {
+// IsHeaderSupportedChain returns true if the chain is supports block header-based verification
```
lumtis, line 104:
```suggestion
// IsHeaderSupportedChain returns true if the chain's consensus supports block header-based verification
```
Deciding if a chain support header is part of the verification flags. The static part here would be more about if the underlying consensus support headers

x/lightclient/keeper/grpc_query_header_enabled_chains.go
```diff
@@ -0,0 +1,38 @@
...
+	ctx := sdk.UnwrapSDKContext(c)
+
+	val, _ := k.GetBlockHeaderVerification(ctx)
+
+	return &types.QueryHeaderSupportedChainsResponse{HeaderSupportedChains: val.GetHeaderSupportedChainsList()}, nil
+}
+
+// HeaderEnabledChains implements the Query/HeaderEnabledChains gRPC method
+// It returns a list of chains that have block header verification enabled.
+func (k Keeper) HeaderEnabledChains(c context.Context, req *types.QueryHeaderEnabledChainsRequest) (*types.QueryHeaderEnabledChainsResponse, error) {
```
lumtis, line 28: Seems to not be tested

x/lightclient/keeper/grpc_query_header_enabled_chains.go
```diff
@@ -0,0 +1,38 @@
...
+// HeaderSupportedChains implements the Query/HeaderEnabledChains gRPC method
+// It returns a list for chains that support block header verification.
+// Some chains in this list might be disabled which is indicated by the value of the `enabled` field.
+func (k Keeper) HeaderSupportedChains(c context.Context, req *types.QueryHeaderSupportedChainsRequest) (*types.QueryHeaderSupportedChainsResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	ctx := sdk.UnwrapSDKContext(c)
+
+	val, _ := k.GetBlockHeaderVerification(ctx)
```
lumtis, line 21: Should return error if not found

x/lightclient/keeper/grpc_query_header_enabled_chains.go
```diff
@@ -0,0 +1,38 @@
...
+
+// HeaderEnabledChains implements the Query/HeaderEnabledChains gRPC method
+// It returns a list of chains that have block header verification enabled.
+func (k Keeper) HeaderEnabledChains(c context.Context, req *types.QueryHeaderEnabledChainsRequest) (*types.QueryHeaderEnabledChainsResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	ctx := sdk.UnwrapSDKContext(c)
+
+	val, _ := k.GetBlockHeaderVerification(ctx)
```
lumtis, line 34: Should return error if not found

x/lightclient/keeper/msg_server_disable_block_header_verification.go
```diff
@@ -0,0 +1,30 @@
...
+// Disabled chains do not allow the submissions of block headers or using it to verify the correctness of proofs
+func (k msgServer) DisableHeaderVerification(goCtx context.Context, msg *types.MsgDisableHeaderVerification) (*types.MsgDisableHeaderVerificationResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// check permission
+	if !k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupEmergency) {
+		return nil, authoritytypes.ErrUnauthorized
+	}
+
+	bhv, _ := k.GetBlockHeaderVerification(ctx)
```
lumtis, line 21: Can we explicitely set an empty value if not found?\
```
if !found{
  bhv = BlockHeaderVerification{}
}
```

x/lightclient/types/block_header_verification.go
```diff
@@ -0,0 +1,82 @@
+package types
```
lumtis, line 1: I think we can also add a validate method that check no duplicated chain IDs and all chain IDs are valid. And add this validate to genesis validate

We can also create an issue and add it to the block header epic since this module is not a priority for now

x/lightclient/keeper/grpc_query_header_enabled_chains.go
kingpinXD: 1e5d62c979b52915ce73e5ef819d5a479ab8cd77

lumtis (approved):

proto/lightclient/genesis.proto
```diff
@@ -12,5 +12,5 @@ option go_package = "github.com/zeta-chain/zetacore/x/lightclient/types";
 message GenesisState {
   repeated proofs.BlockHeader block_headers = 1 [(gogoproto.nullable) = false];
   repeated ChainState chain_states = 2 [(gogoproto.nullable) = false];
-  VerificationFlags verification_flags = 3 [(gogoproto.nullable) = false];
+  BlockHeaderVerification block_header_verification = 3 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
kingpinXD: do you mean would the export work after we add this in? Then yes, it should.

x/lightclient/keeper/msg_server_enable_block_header_verification.go
```diff
@@ -0,0 +1,35 @@
...
+// EnableHeaderVerification enables the verification flags for the given chain IDs
+// Enabled chains allow the submissions of block headers and using it to verify the correctness of proofs
+func (k msgServer) EnableHeaderVerification(goCtx context.Context, msg *types.MsgEnableHeaderVerification) (
+	*types.MsgEnableHeaderVerificationResponse,
+	error,
+) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// check permission
+	if !k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupOperational) {
```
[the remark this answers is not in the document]
kingpinXD: Not sure about the exact reasoning , but this pr tries to retain the existing functionality

x/lightclient/types/block_header_verification.go
```diff
@@ -0,0 +1,95 @@
...
+			ChainId: chainID,
+			Enabled: true,
+		})
+	}
+}
+
+// DisableChain disables block header verification for a specific chain
+// This function does not remove the chain from the list of enabled chains, it just disables it
+// This keeps track of the chains tha support block header verification and also the ones that currently disabled or enabled
+func (b *BlockHeaderVerification) DisableChain(chainID int64) {
```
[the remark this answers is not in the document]
kingpinXD: do you mean if disable chain is called for a chain is which is not part of `HeaderSupportedChains` ? I think its fine to set it as disabled

PR #2113, opened 2024-05-02
feat: add zetaclientd-supervisor

gartnera:
# Description

Add a `zetaclientd-supervisor` process which will watch the version of `zetacored` and automatically restart zetaclient if the version changes. When the version has changed, cosmovisor will have already downloaded and symlinked in the new `zetaclientd` binary so all we need to do is restart to pick up the new change.

We are careful to do a few things:
- forward signals from `zetaclientd-supervisor` to `zetaclientd`
- serialize the stdout so that log messages from the `zetaclientd` subprocess are not interleaved with the `zetaclientd-supervisor` logs messages
- cache the hotkey and tss password so that it only needs to be input once even if zetaclientd restarts/crashes

~~If we decide not to rely on cosmovisor autodownload for delivery of the `zetaclientd` binary, we would have to add autodownload/upgrade functionality to this process as well. That logic isn't terribly complex but great care must be taken to ensure the download is performed correctly. We would retrieve the binaries by grabbing the current upgrade plan from zetacore via rpc.~~

Automatic download and symlinking like cosmovisor has now been added. The new zetaclient binary will automatically be downloaded and loaded when the zetacored version changes.

This is functional in the localnet docker images and upgrade tests. Example with v16 -> v17 -> v18 (dummy release delivered via upgrade info binary urls):

```
➜  ~ docker logs zetaclient0 2>&1 | grep zetaclientd-supervisor | grep -v 'connection error'
2024-05-08T22:25:10Z WRN got new upgrade plan (v17) module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:25:10Z ERR downloadZetaclientd failed error="upgrade info empty" module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:27:02Z WRN core version change (v12.3.0-rc-96-g181ba7ff -> v12.3.0-rc-124-gcb66f60c) module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:27:02Z ERR error while waiting error="exit status 2" process=zetaclientd-supervisor
2024-05-08T22:30:43Z WRN got new upgrade plan (v18) module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:30:43Z INF downloading zetaclientd module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:31:35Z WRN core version change (v12.3.0-rc-124-gcb66f60c -> v12.3.0-rc-113-gb4605569) module=zetaclientdSupervisor process=zetaclientd-supervisor
2024-05-08T22:31:35Z ERR error while waiting error="exit status 2" process=zetaclientd-supervisor
```

```
zetaclient0:~/.zetaclientd/upgrades# find .
.
./genesis
./genesis/zetaclientd
./v18
./v18/zetaclientd
./current
./v17
./v17/zetaclientd
zetaclient0:~/.zetaclientd/upgrades# ls -lth
total 12K
lrwxrwxrwx    1 root     root          31 May  8 22:31 current -> /root/.zetaclientd/upgrades/v18
drwxr-x---    2 root     root        4.0K May  8 22:30 v18
drwxr-xr-x    2 root     root        4.0K May  8 22:22 v17
drwxr-xr-x    2 root     root        4.0K May  8 22:22 genesis
```

Relates to DEVOP-642

Degraded without #2135 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [x] Go integration tests

# Checklist:


cmd/zetaclientd-supervisor/main.go
```diff
@@ -0,0 +1,160 @@
...
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+
+	hotkeyPassword, tssPassword, err := promptPasswords()
+	if err != nil {
+		panic(fmt.Errorf("unable to get passwords: %w", err))
+	}
+
+	for {
+		cmd := exec.Command("zetaclientd", os.Args[1:]...)
```
lumtis, line 106: Maybe we can use a const for `zetaclientd`

cmd/zetaclientd-supervisor/main.go
```diff
@@ -0,0 +1,160 @@
...
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+
+	hotkeyPassword, tssPassword, err := promptPasswords()
+	if err != nil {
+		panic(fmt.Errorf("unable to get passwords: %w", err))
+	}
+
+	for {
```
lumtis, line 105: Unclear why we use a for loop here, this is when a new version is detected? maybe we can add a comment to explain?

Also maybe ew could put the content in a sub function

cmd/zetaclientd-supervisor/main.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client/grpc/tmservice"
+	"github.com/rs/zerolog"
+	"github.com/zeta-chain/zetacore/app"
+	"github.com/zeta-chain/zetacore/zetaclient/config"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+)
+
+type serializedWriter struct {
```
lumtis, line 25: Need comment for the purpose for this type

cmd/zetaclientd-supervisor/main.go
lumtis, line 106: Need to use nosec if this gosec deemed as false positive. Might need a review from sec to give a confirmation

contrib/localnet/scripts/start-zetaclientd.sh
```diff
@@ -43,7 +43,7 @@ then
      set_sepolia_endpoint
     fi
 
-    zetaclientd start < /root/password.file
+    zetaclientd-supervisor start < /root/password.file
```
lumtis, line 46: In theory with this change, we should be able to remove `restart-zetaclient` script used in local upgrade test, right?

gartnera: yes exactly

cmd/zetaclientd-supervisor/main.go
gartnera: yeah code structure still needs some though

gartnera: > Unclear why we use a for loop here

I use a loop because we need to to restart the zetaclientd process when we get the restart signal. The supervisor process generally won't restart because it needs to keep the hotkey/tss password stored in memory.

kingpinXD (approved):
LGTM , 
NIT : comments before every function might be useful, not required but definitely good to have

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,258 @@
...
+		// even without #2135, the version will still change and trigger the update
+		newVersionDir := s.dirForVersion(s.upgradePlanName)
+		currentLinkPath := s.dirForVersion("current")
+
+		err = atomicSymlink(newVersionDir, currentLinkPath)
+		if err != nil {
+			s.logger.Error().Err(err).Msgf("unable to update current symlink (%s -> %s)", newVersionDir, currentLinkPath)
+			return
+		}
+		s.reloadSignals <- true
```
kingpinXD, line 153: Would this trigger in the block after the upgrade happens on zetacore ? 

We can maybe consider saving the upgrade height as well during handleCoreUpgrade , so that the clients can be restart at the height directly , wdyt

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,258 @@
...
+	binURL, ok := config.Binaries[binKey]
+	if !ok {
+		return fmt.Errorf("no binary found for: %s", binKey)
+	}
+	upgradeDir := s.dirForVersion(plan.Name)
+	err = os.MkdirAll(upgradeDir, 0o750)
+	if err != nil {
+		return fmt.Errorf("mkdir %s: %w", upgradeDir, err)
+	}
+	upgradePath := path.Join(upgradeDir, "zetaclientd")
```
kingpinXD, line 219:
```suggestion
	upgradePath := path.Join(upgradeDir, zetaclientdBinaryName)
```

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,258 @@
...
+func (s *zetaclientdSupervisor) handleCoreUpgradePlan(ctx context.Context) {
+	client := upgradetypes.NewQueryClient(s.zetacoredConn)
+
+	prevPlanName := ""
+	for {
+		select {
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			return
+		}
```
kingpinXD, line 166: comments here would be useful

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,258 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("grpc dial: %w", err)
+	}
+
+	return &zetaclientdSupervisor{
+		zetacoredConn: conn,
+		logger:        logger,
+		reloadSignals: make(chan bool, 1),
+		upgradesDir:   os.ExpandEnv("$HOME/.zetaclientd/upgrades"),
```
kingpinXD, line 77: Maybe let's put this as a constant at the top of the file

cmd/zetaclientd-supervisor/lib.go
gartnera: This will not trigger until the zetacored upgrade has completed.

zetaclient itself should detect once upgrade height has been reached and pause itself:

https://github.com/zeta-chain/node/blob/e0287d7d784c2f20c637e8d0409704a3a768cdf3/zetaclient/zetabridge/zetacore_bridge.go#L196-L200

although if blocks are created faster than the the core context updater ticker then zetaclient wouldn't actually pause itself (I'm thinking about mainnet)

changelog.md
```diff
@@ -11,6 +11,7 @@
 
 * [2032](https://github.com/zeta-chain/node/pull/2032) - improve some general structure of the ZetaClient codebase
 * [2100](https://github.com/zeta-chain/node/pull/2100) - cosmos v0.47 upgrade
+* [2113](https://github.com/zeta-chain/node/pull/2113) - add zetaclientd-supervisor
```
lumtis, line 14:
```suggestion
* [2113](https://github.com/zeta-chain/node/pull/2113) - add zetaclientd-supervisor process
```

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	fmt.Print("TSS Password: ")
+	TSSKeyPass, err := reader.ReadString('\n')
+	if err != nil {
+		return "", "", err
+	}
+
+	//trim delimiters
+	hotKeyPass = strings.TrimSuffix(hotKeyPass, "\n")
+	TSSKeyPass = strings.TrimSuffix(TSSKeyPass, "\n")
```
lumtis, line 264:
```suggestion
	tssKeyPass = strings.TrimSuffix(TSSKeyPass, "\n")
```

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return hotKeyPass, TSSKeyPass, err
```
lumtis, line 266:
```suggestion
	return hotKeyPass, tssKeyPass, err
```

cmd/zetaclientd-supervisor/lib.go
```diff
@@ -0,0 +1,267 @@
...
+
+func promptPasswords() (string, string, error) {
+	reader := bufio.NewReader(os.Stdin)
+	fmt.Print("HotKey Password: ")
+	hotKeyPass, err := reader.ReadString('\n')
+	if err != nil {
+		return "", "", err
+	}
+	fmt.Print("TSS Password: ")
+	TSSKeyPass, err := reader.ReadString('\n')
```
lumtis, line 257:
```suggestion
	tssKeyPass, err := reader.ReadString('\n')
```

lumtis (commented):
Looks good to me and will be very useful.

Regarding
> cache the hotkey and tss password so that it only needs to be input once even if zetaclientd restarts/crashes
Caching here means it is kept as a local variable in the function and reused? Doesn't it bring some security concerns if someone can dump the memory?
ccing @CharlieMc0

gartnera:
> > cache the hotkey and tss password so that it only needs to be input once even if zetaclientd restarts/crashes
> 
> Caching here means it is kept as a local variable in the function and reused? Doesn't it bring some security concerns if someone can dump the memory? ccing @CharlieMc0

Today, you could just dump the zetaclient process memory. We make no efforts to zero the memory/variables we use to store the password in zetaclient.

CharlieMc0:
> > > cache the hotkey and tss password so that it only needs to be input once even if zetaclientd restarts/crashes
> > 
> > 
> > Caching here means it is kept as a local variable in the function and reused? Doesn't it bring some security concerns if someone can dump the memory? ccing @CharlieMc0
> 
> Today, you could just dump the zetaclient process memory. We make no efforts to zero the memory/variables we use to store the password in zetaclient.

Yeah the risk of someone dumping memory already exists because zetaclient has to hold the key in memory unencrypted. It's already cached in memory while zetaclient is running. The only difference here is the supervisor also caches it between restarts.

CharlieMc0 (approved):

Issue #2107, opened 2024-05-02
Make authorized actions more flexible to manage


lumtis:
This epic groups task to make the authorized actions from `authority` more flexible to manage:

- Add the ability with a table to dynamically change authorization requirements for messages
- Add the governance messages (UpdateParams in emissions currently) here, so this can be grouped in the same table? Some messages could become governance authorized over time

```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/1562
- [ ] https://github.com/zeta-chain/node/issues/1236
- [ ] https://github.com/zeta-chain/node/issues/2244
- [ ] https://github.com/zeta-chain/node/issues/2245
- [ ] https://github.com/zeta-chain/node/issues/2246
- [ ] https://github.com/zeta-chain/node/issues/2288
- [ ] https://github.com/zeta-chain/node/issues/2307
- [ ] https://github.com/zeta-chain/node/issues/2310
- [ ] https://github.com/zeta-chain/node/issues/2311
- [ ] https://github.com/zeta-chain/node/issues/2247
```

lumtis:
Putting this v18 for now, we will reevaluate priority after v17 upgrade

Issue #1236, opened 2023-10-04
Create a table for policy group authorizations


lumtis:
We currently manually check for each privilege messages if it the sender has a specific policy group for execution. Example:
https://github.com/zeta-chain/node/blob/dec6172627009f37de6fd872d1837e4dc75ebfd4/x/fungible/keeper/msg_server_udpate_zrc20_liquidity_cap.go#L18

It can be interesting to have in the codebase somewhere (in `observer` module) a place where all authorizations are centralized in a table.

This would allow having an overview on the policy group and simplifying upgrades.

## Possible solution:

Having a hardcoded map that associates actions with group:
```
Policies := map[Action]PolicyGroup{
	DisableInbound:  Group1,
	EnablingInbound: Group2,
	...
}
```

```
func CheckAuthorized(signer string, action Action) bool {
	requiredGroup := Policies[action]
	return signer == k.GetParams(ctx).GetAdminPolicyAccount(requiredGroup)
}
```

```
func (k msgServer) UpdateCoreParams(...) ... {
	if CheckAuthorized() {
		return types.ErrNotAuthorizedPolicy
	}
	...
}
```

kingpinXD:
Closing this issue , and working on the new sub tasks created

PR #2114, opened 2024-05-03
refactor: authorizations table

kingpinXD:
# Description

- Adds an authorization table
- Separate `MsgUpdateZRC20` into `MsgPauseZRC20` and `MsgUnPauseZRC20` 
- Separate `MsgUpdateCrosschainFlags` into `MsgEnableCCTXFlags` `MsgDisableCCTXFlags` and `MsgUpdateGasPriceIncreaseFlags` 
- Policy requirements for new messages 
 - `MsgPauseZRC20 :  PolicyType_groupEmergency`
 - `MsgUnPauseZRC20 PolicyType_groupOperational`
 - `MsgUpdateGasPriceIncreaseFlags : PolicyType_groupOperational`
 - `MsgEnableCCTXFlags : PolicyType_groupOperational`
 - `MsgDisableCCTXFlags : PolicyType_groupEmergency`

Closes:https://github.com/zeta-chain/node/issues/2107

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


pkg/authorizations/authorizations.go
```diff
@@ -0,0 +1,52 @@
+package authorizations
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+	crosschainTypes "github.com/zeta-chain/zetacore/x/crosschain/types"
```
lumtis, line 6:
```suggestion
	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
```

pkg/authorizations/authorizations.go
```diff
@@ -0,0 +1,52 @@
...
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func AuthorizationTable() map[string]authoritytypes.PolicyType {
```
lumtis, line 12: I realize the issue https://github.com/zeta-chain/node/issues/1236 is about having it hardcoded but in the future we could think about having this structure on-chain to update it dynamically.

While I think we don't need to implement it for this PR, I think we should still consider changing the map into a slice so we can persist it in the store in the future

lumtis, line 12: Why using a function instead of defining a global var directly?

testutil/keeper/authority.go
```diff
@@ -74,8 +75,13 @@ func AuthorityKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 }
 
 // MockIsAuthorized mocks the IsAuthorized method of an authority keeper mock
-func MockIsAuthorized(m *mock.Mock, address string, policyType types.PolicyType, isAuthorized bool) {
-	m.On("IsAuthorized", mock.Anything, address, policyType).Return(isAuthorized).Once()
+// TODO : https://github.com/zeta-chain/node/issues/2153
```
lumtis, line 78: We should add a sentence explaining todo in addition to the link

x/crosschain/keeper/msg_server_migrate_tss_funds.go
```diff
@@ -27,8 +27,9 @@ func (k msgServer) MigrateTssFunds(goCtx context.Context, msg *types.MsgMigrateT
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	// check if authorized
-	if !k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupAdmin) {
-		return nil, errorsmod.Wrap(authoritytypes.ErrUnauthorized, "Update can only be executed by the correct policy account")
+	ok, err := k.GetAuthorityKeeper().IsAuthorized(ctx, msg)
+	if !ok || err != nil {
+		return nil, errorsmod.Wrap(authoritytypes.ErrUnauthorized, err.Error())
```
lumtis, line 32: This would panic if ok == false, err == nil. Has it been tested?

x/crosschain/keeper/msg_server_add_to_outtx_tracker.go
```diff
@@ -48,7 +48,7 @@ func (k msgServer) AddToOutTxTracker(goCtx context.Context, msg *types.MsgAddToO
 		return &types.MsgAddToOutTxTrackerResponse{IsRemoved: true}, nil
 	}
 
-	isEmergencyGroup := k.GetAuthorityKeeper().IsAuthorized(ctx, msg.Creator, authoritytypes.PolicyType_groupEmergency)
+	isEmergencyGroup, _ := k.GetAuthorityKeeper().IsAuthorized(ctx, msg)
```
lumtis, line 51: Why skipping errors in these?

x/authority/keeper/policies.go
```diff
@@ -25,15 +29,25 @@ func (k Keeper) GetPolicies(ctx sdk.Context) (val types.Policies, found bool) {
 }
 
 // IsAuthorized checks if the address is authorized for the given policy type
-func (k Keeper) IsAuthorized(ctx sdk.Context, address string, policyType types.PolicyType) bool {
+func (k Keeper) IsAuthorized(ctx sdk.Context, msg sdk.Msg) (bool, error) {
```
lumtis, line 32: I don't think we need a error a any point
If policies is not found, no policies so not authorized
If more than 2 signers, not authorized

x/fungible/keeper/msg_server_unpause_zrc20.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	err = ctx.EventManager().EmitTypedEvent(
+		&types.EventZRC20PausedStatusUpdated{
+			MsgTypeUrl:     sdk.MsgTypeURL(&types.MsgUnpauseZRC20{}),
+			Action:         types.UpdatePausedStatusAction_UNPAUSE,
+			Zrc20Addresses: msg.Zrc20Addresses,
+			Signer:         msg.Creator,
+		},
+	)
```
lumtis, line 49: We no longer need to emit events for message as it is done automatically since sdk 0.47

lumtis (changes requested):
Not complete review but giving some comments already

x/crosschain/keeper/msg_server_migrate_tss_funds.go
kingpinXD: How would that happen? The function always returns an error message if the `IsAuthorized` Fails. The error message is used to add some more context to why the authorization failed.

x/crosschain/keeper/msg_server_add_to_outtx_tracker.go
kingpinXD: We should not return on error here as the message can be signed by an observer also .

x/authority/keeper/policies.go
kingpinXD: The error adds some context to why the message is not authorized.
I considered logging the message but found using an error to be better for the user who is trying to broadcast .

pkg/authorizations/authorizations.go
kingpinXD: There is no specific reason. I guess both options work.
The initial version of this pr , had some logic to deal with messages which need more than one policy type (before the refactor for light client pr was merged ) and thus needed a function . But it can be changed to a var now if needed.

lumtis: Var is easier to navigate with an IDE

x/crosschain/keeper/msg_server_migrate_tss_funds.go
lumtis: It's an antipattern to return other variables with the error, if the function returns an error it should be assumed that the other returned values should not be used, if custom values are returned, it should be assumed err is nil.

We should change the error into a string `nonAuthorizedReason` or remove the boolean and just return an error when not auithorized

kingpinXD: d3741a6e5e58747e2fa9cd6f39ece7d6ca0f0a37

testutil/keeper/authority.go
kingpinXD: d3741a6e5e58747e2fa9cd6f39ece7d6ca0f0a37

pkg/authorizations/authorizations.go
kingpinXD: 9485063a5dcc6ff1f35d38cf0c3f70a92507a9a5

kingpinXD: 9485063a5dcc6ff1f35d38cf0c3f70a92507a9a5

x/fungible/keeper/msg_server_unpause_zrc20.go
kingpinXD: Not sure what you mean , The Pause and Unpause use the same event type with only the Action being different

lumtis: Ok but I don't think we need this event if we already emit one for pause or unpause

kingpinXD: The unpause TX has 
`Action:         types.UpdatePausedStatusAction_UNPAUSE,`
The pause TX has 
`Action:         types.UpdatePausedStatusAction_PAUSE,`

Are you saying that we should emit only one?

x/observer/keeper/msg_server_disable_cctx_flags.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// check permission
+	err := k.GetAuthorityKeeper().IsAuthorized(ctx, msg)
+	if err != nil {
+		return nil, cosmoserrors.Wrap(authoritytypes.ErrUnauthorized, err.Error())
+	}
+
+	// check if the value exists,
+	// if not, set the default value for the Inbound and Outbound flags only
+	flags, isFound := k.GetCrosschainFlags(ctx)
```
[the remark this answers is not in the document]
kingpinXD: Thus, PR just retains the existing functionality. 
We can change it if needed but it should be a separate PR

x/observer/client/cli/tx_disable_cctx_flags.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func CmdDisableCCTXFlags() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "disable-cctx-flags [disable-inbound] [disable-outbound]",
+		Short: "Disable inbound and outbound cross-chain flags",
```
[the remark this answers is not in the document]
kingpinXD: Similar to another reply, this tries to retain the existing functionality so that we can disable only one if needed. 
The main reason for splitting the two messages is to have one authorized policy per message type

x/authority/keeper/policies.go
```diff
@@ -25,15 +29,25 @@ func (k Keeper) GetPolicies(ctx sdk.Context) (val types.Policies, found bool) {
...
-func (k Keeper) IsAuthorized(ctx sdk.Context, address string, policyType types.PolicyType) bool {
+func (k Keeper) IsAuthorized(ctx sdk.Context, msg sdk.Msg) error {
+	// Policy transactions must have only one signer
+	if len(msg.GetSigners()) != 1 {
+		return errors.Wrap(types.ErrSigners, fmt.Sprintf("msg: %v", sdk.MsgTypeURL(msg)))
+	}
+	signer := msg.GetSigners()[0].String()
+	policyRequired := authorizations.GetRequiredPolicy(sdk.MsgTypeURL(msg))
+	if policyRequired == types.PolicyType_emptyPolicyType {
+		return errors.Wrap(types.ErrMsgNotAuthorized, fmt.Sprintf("msg: %v", sdk.MsgTypeURL(msg)))
```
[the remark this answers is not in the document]
kingpinXD: Imho its fine , the function checks for authorization irrespective of what the reason is , i.e its the creator or the msg

testutil/sample/authority.go
```diff
@@ -20,3 +24,54 @@ func Policies() authoritytypes.Policies {
 		},
 	}
 }
+
+func PoliciesWithAdmin(address string) authoritytypes.Policies {
```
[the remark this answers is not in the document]
kingpinXD: imo its allright , I can use something like `PolicyListWithProvidedAdmin` but it sounds to descriptive and this function is used for tests only

pkg/authorizations/authorizations.go
```diff
@@ -0,0 +1,81 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func GetRequiredPolicy(msgURl string) authoritytypes.PolicyType {
+	if CheckPolicyList(msgURl, OperationalPolicyMessageList) {
```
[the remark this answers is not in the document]
kingpinXD: I think the improvement would negligible in this case . wdyt

x/observer/keeper/msg_server_update_gas_price_increase_flags.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+
+	cosmoserrors "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+	"github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func (k msgServer) UpdateGasPriceIncreaseFlags(goCtx context.Context, msg *types.MsgUpdateGasPriceIncreaseFlags) (*types.MsgUpdateGasPriceIncreaseFlagsResponse, error) {
```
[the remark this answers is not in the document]
kingpinXD: I considered that while working on this ,but Inbound and Outbound are single values where as GasPrice is a set of values , It might make sense to group the inbound and outbound flags under CCTX flags in the future but I don't think its necessary for this pr .

pkg/authorizations/authorizations.go
```diff
@@ -0,0 +1,81 @@
...
+func CheckPolicyList(msgURl string, msgList []string) bool {
+	for _, policy := range msgList {
+		if policy == msgURl {
+			return true
+		}
+	}
+	return false
+}
+
+var OperationalPolicyMessageList = []string{
```
lumtis, line 34: I'm thinking about defining a type `PolicyTable` containing a list [url,policy] tuples with crud methods, so it can be easily update, even if we add a new policy type and can be used once we persist the table in the state

pkg/authorizations/authorizations.go
```diff
@@ -0,0 +1,81 @@
...
+	// Crosschain admin messages
+	sdk.MsgTypeURL(&crosschaintypes.MsgRefundAbortedCCTX{}),
```
lumtis, line 36: Can be made less verbose defining an alias
```
msgURL := sdk.MsgTypeURL
```

pkg/authorizations/authorizations_test.go
```diff
@@ -0,0 +1,76 @@
+package authorizations_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zeta-chain/zetacore/pkg/authorizations"
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+)
+
```
lumtis, line 10: We can add assertion with all messages in the tests

x/fungible/types/message_unpause_zrc20.go
```diff
@@ -0,0 +1,59 @@
+package types
```
lumtis, line 1: The test for this file use the pause message

PR #2135, opened 2024-05-07
feat: add develop build version logic

gartnera:
# Description

As part of developnet, we need to stamp the version differently to ensure that we have a new unique version for every build. Release builds will be named based on their tag but develop and local builds will be named based on their timestamp.

This will allow us to deploy developnet changes via governance proposals. Each software update deployed via a governance proposal must have a matching UpgradeHandler configured. Use the version constant in `setup_handlers.go` so that the `zetacored version` always matches the upgrade handler version.

This also enables #2113 by ensuring that the version numbers in the upgrades/governance proposals match what zetacore says it's version actually is.

Relates to DEVOP-642

Needed for #2095 

## Type of change

# Checklist:

- [x] verify both develop and release upgrades work


app/setup_handlers.go
```diff
@@ -50,8 +49,8 @@ func SetupHandlers(app *App) {
 		}
 	}
 	baseAppLegacySS := app.ParamsKeeper.Subspace(baseapp.Paramspace).WithKeyTable(paramstypes.ConsensusParamsKeyTable())
-	app.UpgradeKeeper.SetUpgradeHandler(releaseVersion, func(ctx sdk.Context, plan types.Plan, vm module.VersionMap) (module.VersionMap, error) {
-		app.Logger().Info("Running upgrade handler for " + releaseVersion)
+	app.UpgradeKeeper.SetUpgradeHandler(constant.Version, func(ctx sdk.Context, plan types.Plan, vm module.VersionMap) (module.VersionMap, error) {
```
lumtis, line 52: Right now this value is empty

lumtis (commented):
Looks like `make install` fails for the E2E tests. It runs on my machine

app/setup_handlers.go
gartnera: It's empty in the code. It is set by the linker:

https://github.com/zeta-chain/node/blob/1ed1afcffc0299a3994db5359bcfb6e56b565d8e/Makefile#L16

gartnera:
> Looks like `make install` fails for the E2E tests. It runs on my machine

fixed

lumtis (approved):

kingpinXD (approved):

lumtis:
FYI Just added a changelog entry, I think the change is relevant enough for one

PR #2138, opened 2024-05-07
fix: adjust docker login conditional

gartnera:
Adjust the docker login conditional again to actually fix forks.

If we want the workflows to be run by external people, we should not use the buildjet runners. [github did recently upgrade the free runners for opensource projects btw](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/)

You cannot test the presence of secrets in if statements btw.


lumtis (approved):
LGTM

julianrubino (approved):

Issue #2089, opened 2024-04-29
Local E2E tests fail with `crosschain_swap failed: btc transaction not signed`


lumtis:
**Describe the Bug**
The Bitcoin local e2e Crosschain Swap sometimes fails with the error
```
bitcoin tests failed: crosschain_swap failed: btc transaction not signed, stack trace goroutine 119 [running]:
```
This used to happen sometimes but appears to occur more and more frequently, we should therefore consider fixing it once for all

**To Reproduce**
Happens sometime in E2E tests execution like: https://github.com/zeta-chain/node/actions/runs/8877450563/job/24371074194?pr=2060
It appears to never happens locallly on my side

**Screenshots**
![e2efail](https://github.com/zeta-chain/node/assets/7040350/44de38b0-14bf-4478-a7a1-7563d5efed7a)

gartnera:
https://github.com/zeta-chain/node/actions/runs/8972215322/job/24639719251?pr=2119

gartnera:
https://github.com/zeta-chain/node/runs/24807859621

gartnera:
https://github.com/zeta-chain/node/actions/runs/9169683804/job/25210551669

PR #2151, opened 2024-05-10
fix: add retry for Bitcoin signing

lumtis:
# Description

Tentative fix for https://github.com/zeta-chain/node/issues/2089
Add a retry for the signature, in case the error is related to the network.

Closes: https://github.com/zeta-chain/node/issues/2089


lumtis:
Not solving the issue: https://github.com/zeta-chain/node/actions/runs/9032989998/job/24822298819

Issue #2147, opened 2024-05-09
Use ubuntu/debian Base Docker Images


gartnera:
We build our release binaries for ubuntu/glibc. This means that we cannot use our release binaries in the alpine docker containers. We should just use ubuntu everywhere for consistency.

Issue #2146, opened 2024-05-09
Refactor upgrade tests to allow three way upgrades


gartnera:
We need the ability to easily test upgrading from version A -> B -> C. I have some [hacky tools](https://gist.github.com/gartnera/a2c82a46053034c594e06312bd855b3d) to do it locally but we should refactor how the upgrade tests work to enable this workflow.

TODO:
- [ ] always run localnet under cosmovisor and zetaclientd to enable upgrading any node image
- [ ] push these images to registry
- [ ] pull base image to start upgrade from docker registry rather than building it every time
- [ ] move upgrade proposal submitter to sidecar container. or just push down into e2e.
- [ ] serve new binaries via `docker run --rm -it --net localnet_mynetwork --ip 172.20.0.109 --name newzeta --entrypoint /usr/bin/python zetanode:v16.0.0 -m http.server` and deliver via autodownload

PR #2158, opened 2024-05-10
refactor: unify CI docker images

gartnera:
# Description

- unify the localnet and upgrade docker images by using a multistage build. you can stop a multistage build by using the `--target` option.
- use debian instead of alpine for glibc compatibility. this means that we can deploy our real releases into the localnet containers.
- always run `zetacored` under `cosmovisor`. all localnet builds can now accept and process upgrade proposals.
- serve upgrade binaries from a sidecar container `upgradehost`
- run the upgrade governance proposal from `upgradeorchestrator`
- simplify the docker compose file a bit (`docker-compose-upgrade-light.yml` now inherits from `docker-compose-upgrade.yml`)
- remove static ssh keys from docker build. ssh keys are now generated on first `zetacored0` boot and stored on a volume. this will enable caching docker images to speed up CI. update: this highlighted a few race conditions that I had to fix.

Relates to #2146  #2147

Relates to DEVOP-642 DEVOP-643 

# How Has This Been Tested?

- [x] Go integration tests
- [x] Tested via GitHub Actions ([upgrade tests pass](https://github.com/zeta-chain/node/actions/runs/9067779336/job/24913637184))

# Checklist:

- [x] light upgrade tests pass locally


Dockerfile-localnet
```diff
@@ -34,11 +50,31 @@ COPY contrib/localnet/zetacored /root/zetacored
...
+FROM base-runtime AS latest-runtime
+
+COPY --from=cosmovisor-build /go/bin/cosmovisor /usr/local/bin
+COPY --from=latest-build /go/bin/zetacored /go/bin/zetaclientd /go/bin/zetaclientd-supervisor /go/bin/zetae2e /usr/local/bin
+
+# optional old version build
+# use --target latest-runtime to skip
+# 
+# TODO: just download binaries from github release now that we're using glibc
+# we can't do this right now since we do not have a v16 release candidate
```
lumtis, line 66: I think it would be great if we can create issues for all TODOs and link those in comment

Dockerfile-localnet
```diff
@@ -34,11 +50,31 @@ COPY contrib/localnet/zetacored /root/zetacored
...
-ENV SHELL /bin/sh
 EXPOSE 22
 
-ENTRYPOINT ["/usr/sbin/sshd", "-D"]
+FROM base-runtime AS latest-runtime
+
+COPY --from=cosmovisor-build /go/bin/cosmovisor /usr/local/bin
+COPY --from=latest-build /go/bin/zetacored /go/bin/zetaclientd /go/bin/zetaclientd-supervisor /go/bin/zetae2e /usr/local/bin
+
+# optional old version build
```
lumtis, line 62: Since we're removing `upgrade` Dockerfile, it might be good to add some comments adding context for having the old version build here: would be used to test upgrade locally

contrib/localnet/scripts/start-upgradehost.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
lumtis, line 1: `start-upgradehost.sh` do we name it like this because the name of the service? I think it would be better to use dash case entirely `start-upgrade-host.sh`

contrib/localnet/scripts/start-upgradeorchestrator.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
```
lumtis, line 1: Same comment as above for dash case

lumtis (commented):
LGTM just a few minor comments

Dockerfile-localnet
gartnera: done

Dockerfile-localnet
gartnera: done

contrib/localnet/scripts/start-upgradehost.sh
gartnera: added dash to script name and docker service name

lumtis (approved):

gartnera:
upgrade tests are passing before final merge: https://github.com/zeta-chain/node/actions/runs/9082090862/job/24957575444

PR #2175, opened 2024-05-13
ci: upload zetacore/zetaclient binaries from build

gartnera:
Upload binaries build in CI so that we can download and use them to upgrade developnet


gartnera:
You cannot download the artifacts without authentication, so we cannot serve developnet upgrades from here.

julianrubino (approved):

lumtis (approved):
LGTM

Issue #2182, opened 2024-05-14
Include priority mempool checks in e2e/performance tests


skosito:
**Describe the Issue**
Check if system txs are included first in e2e/performance tests

PR #2184, opened 2024-05-14
test: add tx priority checks to e2e tests

skosito:
# Description

Add function to monitor blocks while e2e tests are executed, and at the end of execution, n some block system transactions are not above ethermint txs, it should fail. Initially wanted to do this only for performance tests, but this seems more complete and better because one monitoring is needed per execution, not per test - open to discussion.

Monitor function is not super clean, but it is based on information in txs results.

These txs are on top because they have highest priority in mempool. If this is executed on develop before priority mempool PR was merged, it would fail, because txs were included in FIFO order.

Closes: #2182 

## Type of change

Test addition.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/cmd/zetae2e/local/monitor_priority_txs.go
+++ b/cmd/zetae2e/local/monitor_priority_txs.go
@@ -71,6 +71,7 @@ func processTx(txResult *coretypes.ResultTx, nonSystemTxFound *bool, errCh chan 
 					// a non system tx has been found in the block before a system tx
 					if *nonSystemTxFound {
 						errCh <- errors.New("wrong tx priority, system tx not on top")
+						return
 					}
 				} else {
 					*nonSystemTxFound = true
@@ -92,6 +93,7 @@ func isMsgTypeURLSystemTx(attr types.EventAttribute) bool {
 		"\"/zetachain.zetacore.crosschain.MsgVoteOnObservedInboundTx\"",
 		"\"/zetachain.zetacore.crosschain.MsgVoteGasPrice\"",
 		"\"/zetachain.zetacore.crosschain.MsgAddToOutTxTracker\"",
+		"\"/zetachain.zetacore.crosschain.MsgAddToInTxTracker\"",
 		"\"/zetachain.zetacore.observer.MsgVoteBlockHeader\"",
 		"\"/zetachain.zetacore.observer.MsgVoteTSS\"",
 		"\"/zetachain.zetacore.observer.MsgAddBlameVote\"",
--- a/app/ante/ante.go
+++ b/app/ante/ante.go
@@ -172,6 +172,7 @@ func IsSystemTx(tx sdk.Tx, isAuthorizedSigner func(string) bool) bool {
 		*crosschaintypes.MsgVoteOnObservedInboundTx,
 		*crosschaintypes.MsgVoteOnObservedOutboundTx,
 		*crosschaintypes.MsgAddToOutTxTracker,
+		*crosschaintypes.MsgAddToInTxTracker,
 		*observertypes.MsgVoteBlockHeader,
 		*observertypes.MsgVoteTSS,
 		*observertypes.MsgAddBlameVote:
--- a/app/ante/ante_test.go
+++ b/app/ante/ante_test.go
@@ -36,6 +36,7 @@ func TestIsSystemTx(t *testing.T) {
 	//		*cctxtypes.MsgVoteOnObservedInboundTx,
 	//		*cctxtypes.MsgVoteOnObservedOutboundTx,
 	//		*cctxtypes.MsgAddToOutTxTracker,
+	//		*cctxtypes.MsgAddToInTxTracker,
 	//		*observertypes.MsgVoteBlockHeader,
 	//		*observertypes.MsgVoteTSS,
 	//		*observertypes.MsgAddBlameVote:
@@ -162,6 +163,24 @@ func TestIsSystemTx(t *testing.T) {
 		{
 			"MsgExec{MsgAddToOutTxTracker}",
 			buildAuthzTxFromMsg(&crosschaintypes.MsgAddToOutTxTracker{
+				Creator: sample.AccAddress(),
+			}),
+			isAuthorized,
+
+			true,
+		},
+		{
+			"MsgAddToInTxTracker",
+			buildTxFromMsg(&crosschaintypes.MsgAddToInTxTracker{
+				Creator: sample.AccAddress(),
+			}),
+			isAuthorized,
+
+			true,
+		},
+		{
+			"MsgExec{MsgAddToInTxTracker}",
+			buildAuthzTxFromMsg(&crosschaintypes.MsgAddToInTxTracker{
 				Creator: sample.AccAddress(),
 			}),
 			isAuthorized,


cmd/zetae2e/local/utils.go
```diff
@@ -100,3 +103,52 @@ func waitKeygenHeight(
...
+		case <-ticker.C:
+			block, err := rpc.Block(ctx, nil)
+			if err != nil {
+				errCh <- err
+			}
+
+			// iterate txs events and check if some MsgEthereumTx is included above crosschain and observer txs
+			nonSystemTxFound := false
+			for _, tx := range block.Block.Txs {
+				txRes, err := rpc.Tx(context.Background(), tx.Hash(), false)
```
lumtis, line 128: Why not using `ctx` object here?

cmd/zetae2e/local/utils.go
```diff
@@ -100,3 +103,52 @@ func waitKeygenHeight(
...
+
+func MonitorTxPriorityInBlocks(ctx context.Context, conf config.Config, errCh chan error) {
+	rpc, err := rpchttp.New(conf.RPCs.ZetaCoreRPC, "/websocket")
+	if err != nil {
+		errCh <- err
+	}
+
+	ticker := time.NewTicker(2 * time.Second)
+	defer ticker.Stop()
+	for {
```
lumtis, line 115: This is a first implementation that can be refactored in the future but I still think we could already create sub function for some of the logic in the loops. The many nested loops makes it hard to read through the code

cmd/zetae2e/local/utils.go
```diff
@@ -100,3 +103,52 @@ func waitKeygenHeight(
...
+				if err != nil {
+					continue
+				}
+
+				for _, ev := range txRes.TxResult.Events {
+					for _, attr := range ev.Attributes {
+						if attr.Key == "msg_type_url" {
+							if strings.Contains(attr.Value, "zetachain.zetacore.crosschain.MsgVote") ||
+								strings.Contains(attr.Value, "zetachain.zetacore.observer.MsgVote") ||
+								strings.Contains(attr.Value, "zetachain.zetacore.observer.MsgAddBlameVote") {
```
lumtis, line 138: I don't think we should rely on `strings.Contains(attr.Value, "zetachain.zetacore.crosschain.MsgVote")`
We should do a equal check with a complete message name.

We can create a method for checking the URL is a system tx directly, maybe we already have it for the logic in the application mempool

cmd/zetae2e/local/utils.go
```diff
@@ -2,10 +2,13 @@ package local
 
 import (
 	"context"
+	"errors"
```
lumtis, line 5: I think we can create a separate file `monitor_priority_txs.go`

We should remove `utils.go` btw

cmd/zetae2e/local/utils.go
```diff
@@ -100,3 +103,52 @@ func waitKeygenHeight(
...
+					continue
+				}
+
+				for _, ev := range txRes.TxResult.Events {
+					for _, attr := range ev.Attributes {
+						if attr.Key == "msg_type_url" {
+							if strings.Contains(attr.Value, "zetachain.zetacore.crosschain.MsgVote") ||
+								strings.Contains(attr.Value, "zetachain.zetacore.observer.MsgVote") ||
+								strings.Contains(attr.Value, "zetachain.zetacore.observer.MsgAddBlameVote") {
+								if nonSystemTxFound {
```
lumtis, line 139: We should add a comment `a non system tx has been found in the block before a system tx`

cmd/zetae2e/local/utils.go
```diff
@@ -100,3 +103,52 @@ func waitKeygenHeight(
...
+									errCh <- errors.New("wrong tx priority, system tx not on top")
+								}
+							}
+						}
+						if attr.Key == "action" {
+							if strings.Contains(attr.Value, "ethermint.evm.v1.MsgEthereumTx") {
+								nonSystemTxFound = true
+							}
+						}
```
lumtis, line 148: Why doing this check? A non system tx would be any messages in the blockchain outside of the system txs defined.

For example a end user could spam bank send messages

lumtis (commented):
If we didn't yet, would be great if we could (at least manually)
- Toggle off the application mempool
- Checking the test no longer pass

cmd/zetae2e/local/monitor_priority_txs.go
```diff
@@ -0,0 +1,111 @@
...
+	"time"
+
+	"github.com/cometbft/cometbft/abci/types"
+	rpchttp "github.com/cometbft/cometbft/rpc/client/http"
+	coretypes "github.com/cometbft/cometbft/rpc/core/types"
+	"github.com/zeta-chain/zetacore/e2e/config"
+)
+
+// MonitorTxPriorityInBlocks checks for transaction priorities in blocks and reports errors
+func MonitorTxPriorityInBlocks(ctx context.Context, conf config.Config, errCh chan error) {
```
lumtis, line 16: The function name can start with a lowercase isn't it?

cmd/zetae2e/local/monitor_priority_txs.go
```diff
@@ -0,0 +1,111 @@
...
+					*nonSystemTxFound = true
+				}
+			}
+		}
+	}
+}
+
+func isMsgTypeURLSystemTx(attr types.EventAttribute) bool {
+	// type urls in attr.Value are in double quotes, so it needs to be formatted like this
+	systemTxsMsgTypeUrls := []string{
```
lumtis, line 90: Any way we could already use this function instead? https://github.com/zeta-chain/node/blob/b63b5dc7354b2bc87a085aff4b73bc13482cecbc/app/ante/ante.go#L153

If this requires a refactoring, we can create an issue to centralize this list

lumtis (approved):
LGTM - minor comments

cmd/zetae2e/local/monitor_priority_txs.go
lumtis: https://github.com/zeta-chain/node/issues/2193

cmd/zetae2e/local/monitor_priority_txs.go
```diff
@@ -0,0 +1,111 @@
...
+		"\"/zetachain.zetacore.crosschain.MsgVoteOnObservedOutboundTx\"",
+		"\"/zetachain.zetacore.crosschain.MsgVoteOnObservedInboundTx\"",
+		"\"/zetachain.zetacore.crosschain.MsgVoteGasPrice\"",
+		"\"/zetachain.zetacore.crosschain.MsgAddToOutTxTracker\"",
```
ws4charlie, line 94: I think we probably missed the `MsgAddToInTxTracker`, not sure if there is more qualifying msg.

cmd/zetae2e/local/monitor_priority_txs.go
```diff
@@ -0,0 +1,111 @@
...
+			if strings.Contains(attr.Value, ".internal.") {
+				continue
+			}
+			switch attr.Key {
+			// if attr key is msg_type_url, check if it's one of system txs, otherwise mark it as non system tx
+			case "msg_type_url":
+				if isMsgTypeURLSystemTx(attr) {
+					// a non system tx has been found in the block before a system tx
+					if *nonSystemTxFound {
+						errCh <- errors.New("wrong tx priority, system tx not on top")
```
ws4charlie, line 73: It would be also good to just short cut the loop with a `return`?

ws4charlie (approved):

Issue #2183, opened 2024-05-14
Fork unit tests for priority mempool


skosito:
**Describe the Issue**
Priority mempool is forked and adapted a bit from cosmos-sdk codebase. Purpose of this issue is to fork and adapt unit tests too.

PR #2199, opened 2024-05-16
test: custom mempool unit tests

skosito:
# Description

Contains unit tests for priority mempool and custom proposal handler. These tests are forked from cosmos-sdk 0.47 codebase, but also extended on someplaces where we are using custom `sender_with_nonce.go` to handle ethermint txs, mostly on Insert method on priority mempool and selecting txs in handler. I marked those changes with comments, tried to keep same tests style.

Closes: #2183 

## Type of change

Unit tests

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -43,6 +43,8 @@
 * [1484](https://github.com/zeta-chain/node/issues/1484) - replaced hard-coded `MaxLookaheadNonce` with a default lookback factor
 * [2125](https://github.com/zeta-chain/node/pull/2125) - fix develop upgrade test
 * [2222](https://github.com/zeta-chain/node/pull/2222) - removed `maxHeightDiff` to let observer scan from Bitcoin height where it left off
+* [2233](https://github.com/zeta-chain/node/pull/2233) - fix `IsSupported` flag not properly updated in zetaclient's context
+* [2243](https://github.com/zeta-chain/node/pull/2243) - fix incorrect bitcoin outbound height in the CCTX outbound parameter
 
 ### CI
 
@@ -51,6 +53,7 @@
 * [2189](https://github.com/zeta-chain/node/pull/2189) - Updated the docker tag when a release trigger runs to be the github event for the release name which should be the version. Removed mac specific build as the arm build should handle that.
 * [2191](https://github.com/zeta-chain/node/pull/2191) - Fixed conditional logic for the docker build step for non release builds to not overwrite the github tag.
 * [2192](https://github.com/zeta-chain/node/pull/2192) - Added release status checker and updater pipeline that will update release statuses when they go live on network.
+
 ## v16.0.0
 
 ### Breaking Changes
@@ -113,8 +116,7 @@
 * [2045](https://github.com/zeta-chain/node/pull/2046) - add grpc query with outbound rate limit for zetaclient to use
 * [2046](https://github.com/zeta-chain/node/pull/2046) - add state variable in crosschain for rate limiter flags
 * [2034](https://github.com/zeta-chain/node/pull/2034) - add support for zEVM message passing
-
-
+* [1825](https://github.com/zeta-chain/node/pull/1825) - add a message to withdraw emission rewards
 
 ### Tests
 
@@ -141,6 +143,7 @@
 * [2008](https://github.com/zeta-chain/node/pull/2008) - add test for connector bytecode update
 * [2047](https://github.com/zeta-chain/node/pull/2047) - fix liquidity cap advanced test
 * [2076](https://github.com/zeta-chain/node/pull/2076) - automatically deposit native zeta to an address if it doesn't exist on ZEVM.
+
 ### Fixes
 
 * [1861](https://github.com/zeta-chain/node/pull/1861) - fix `ObserverSlashAmount` invalid read
@@ -185,7 +188,6 @@
 ### Features
 
 * [1698](https://github.com/zeta-chain/node/issues/1698) - bitcoin dynamic depositor fee
-* [1811](https://github.com/zeta-chain/node/pull/1811) - add a message to withdraw emission rewards
 
 ### Docs
 
--- a/zetaclient/chains/bitcoin/observer/live_test.go
+++ b/zetaclient/chains/bitcoin/observer/live_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math/big"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


app/mempool/priority_nonce_mempool_test.go
```diff
@@ -0,0 +1,970 @@
...
+
+func (s *MempoolTestSuite) TestPriorityNonceTxOrder() {
+	t := s.T()
+	ctx := sdk.NewContext(nil, cmtproto.Header{}, false, log.NewNopLogger())
+	accounts := simtypes.RandomAccounts(rand.New(rand.NewSource(0)), 5)
+	sa := accounts[0].Address
+	sb := accounts[1].Address
+	sc := accounts[2].Address
+
+	tests := []struct {
```
lumtis, line 74: I think it would be more visual to add a name field for the test case

app/mempool/priority_nonce_mempool_test.go
```diff
@@ -0,0 +1,970 @@
...
+		txs   []txSpec
+		order []int
+		fail  bool
```
lumtis, line 77: I don't see cases using fail==true. Is it part of the forked tests?

lumtis (commented):
I don't think we have this file in codecov scope. In general I don't think it is relevant to have the `app` package in codecov because it is generic logic but for this one this would make sense since we have a customized implementation.

We could also consider moving the mem pool priority logic in `pkg` as an independant package

lumtis (approved):
LGTM

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+	type testTx struct {
+		tx       sdk.Tx
+		priority int64
+		bz       []byte
+		size     int
+	}
+
+	testTxs := []testTx{
+		// test 1
```
kingpinXD, line 156: Can we add the test name in these comments instead of numbers.

kingpinXD: or alternatively add these numbers into the test names, which you feel is better, this is just to make the test cases a bit easier to read

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+		"skip multi-signers msg non-sequential sequence": {
+			ctx:      s.ctx,
+			txInputs: []testTx{testTxs[4], testTxs[5], testTxs[6], testTxs[7], testTxs[8]},
+			req: abci.RequestPrepareProposal{
+				MaxTxBytes: 195 + 196,
+			},
+			expectedTxs: []int{4, 8},
+		},
+		"only the first tx is added": {
+			// Because tx 10 is valid, tx 11 can't be valid as they have higher sequence numbers.
```
kingpinXD, line 266: Using index in the comments as well might be better .

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+	s.Require().Equal(testTxs[17].size, 248)
+	s.Require().Equal(testTxs[18].size, 248)
+
+	s.Require().Equal(testTxs[19].size, 259)
+	s.Require().Equal(testTxs[20].size, 248)
+	s.Require().Equal(testTxs[21].size, 248)
+
+	s.Require().Equal(testTxs[22].size, 248)
+
+	testCases := map[string]struct {
```
kingpinXD, line 226: just curious is there any reason for not using `[]struct` , and using `map[string]struct` 
Is the order of execution important for these tests? 

If not, we can consider using `[]struct` as we have been using that format for the rest of the table-driven tests . 
note : I don't see any issues with using the map

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+		ctx         sdk.Context
+		txInputs    []testTx
+		req         abci.RequestPrepareProposal
+		handler     sdk.PrepareProposalHandler
+		expectedTxs []int
+	}{
+		"skip same-sender non-sequential sequence and then add others txs": {
```
kingpinXD, line 233: What does nonsequential here refer to? 
Does it mean that all txs from the sender would be ignored if one of them exceeds fails to get added ?

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+	s.Require().Equal(testTxs[22].size, 248)
+
+	testCases := map[string]struct {
+		ctx         sdk.Context
+		txInputs    []testTx
+		req         abci.RequestPrepareProposal
+		handler     sdk.PrepareProposalHandler
+		expectedTxs []int
+	}{
```
kingpinXD, line 232: Can we add the two following test cases
-  Tx1,Tx2,Tx3  :  Where Tx2 and Tx3 have the same nonce , but Tx2 is larger than maxbytes . 
 The resultant response would be Tx1 and Tx3 
 
 - All valid transactions but only a few get added because of maxbytes limit

pkg/mempool/custom_proposal_handler_test.go
```diff
@@ -0,0 +1,393 @@
...
+		tx       sdk.Tx
+		priority int64
+		bz       []byte
+		size     int
+	}
+
+	testTxs := []testTx{
+		// test 1
+		{tx: buildMsg(s.T(), txConfig, []byte(`0`), [][]byte{secret1}, []uint64{1}), priority: 10},
+		{tx: buildMsg(s.T(), txConfig, []byte(`12345678910`), [][]byte{secret1}, []uint64{2}), priority: 10},
```
kingpinXD, line 158: NIT : maybe we can use a constant here and in the other cases , might make it easier to read 
`MoreThanMaxBytes = []byte(`12345678910`)`

pkg/mempool/custom_proposal_handler_test.go
kingpinXD: yes but also the comment in that case should be also be `Because tx 10 is invalid` ?

pkg/mempool/custom_proposal_handler_test.go
kingpinXD: okay , yeah maybe we can add more meaningfull comments in case we write our own functions in the future , it makes sense to keep the changes as minimal as possible to reduce the diffs

pkg/mempool/custom_proposal_handler_test.go
kingpinXD: - `skip same-sender non-sequential sequence and then add others txs` has a different sender for `testTxs[3]` 
- makes sense

pkg/mempool/custom_proposal_handler_test.go
kingpinXD: Hmmm , is there a reason we need the different sizes  ?
But based on the other comments, I think it would make sense to keep the test as long as it is able to test the functionality

kingpinXD (approved):

Issue #1882, opened 2024-03-12
Upgrade to mockery v3 for mocks generation


lumtis:
Following message when generating mocks:
```
--> Generating mocks
go: github.com/vektra/mockery/v2@latest (in github.com/vektra/mockery/v2@v2.42.0): go.mod:5: unknown directive: toolchain
12 Mar 24 10:20 CET INF couldn't read any config file version=v2.38.0
12 Mar 24 10:20 CET INF Starting mockery dry-run=false version=v2.38.0
12 Mar 24 10:20 CET INF Using config:  dry-run=false version=v2.38.0
12 Mar 24 10:20 CET WRN DEPRECATION: use of the packages config will be the only way to generate mocks in v3. Please migrate your config to use the packages feature. dry-run=false migration=https://vektra.github.io/mockery/v2.38/migrating_to_packages/ url=https://vektra.github.io/mockery/v2.38/features/#packages-configuration version=v2.38.0
```

PR #2207, opened 2024-05-17
chore: use fixed version for Mockery and add mocks generation to `make generate`

lumtis:
# Description

- Use `2.38.0` as a fixed version for mockery
- Add mock generation to make generate
- Regenerate mocks with the used version

Note: Mockery advise to migrate to "packages" for mock generation https://vektra.github.io/mockery/v2.39/migrating_to_packages/#separate-mocks-directory and it will be the only supported way from Mockery v3. However, I think use Go generate is a simpler solution and I would stick to it unless we determine we need to upgrade Mockery in the future

Closes: https://github.com/zeta-chain/node/issues/1882


ws4charlie (approved):
lgtm

Issue #2208, opened 2024-05-17
Create a new E2E tests runner for zevm message passing


lumtis:
**Describe the Issue**
ZEVM message passing makes the number of tests run by the zeta test runner long. The current test in the CI fail because of timeout. We should split this runner into two in order to run the tests concurrently. We could consider one runner for external chain message passing and one of zevm message passing

PR #2209, opened 2024-05-17
test(`e2e`): add new runner for ZEVM message passing

lumtis:
# Description

Everything in the issue

Note: I realize we have lot of duplicated code in runner definition, I think we can make some simplification on this in another PR

Closes: https://github.com/zeta-chain/node/issues/2208


cmd/zetae2e/local/local.go
```diff
@@ -265,14 +265,17 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 		}
 		zetaAdvancedTests := []string{
 			e2etests.TestZetaDepositRestrictedName,
+			e2etests.TestZetaDepositName,
```
ws4charlie, line 268: Is this a advanced test or because it's been already done at the very beginning?

ws4charlie (approved):
Looks good. I just left a question.

cmd/zetae2e/local/local.go
lumtis: It is done by default to fund the account

PR #2229, opened 2024-05-21
chore: updates for developnet

gartnera:
# Description

- build on ubuntu 20.04 for lower glibc version (related to #2175)
- parameterize upgrade orchestrator binary urls and allow overriding
- follow redirects when downloading binary to get version

Relates to DEVOP-642

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ env:
 
 jobs:
   build-and-test:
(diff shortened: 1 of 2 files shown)


CryptoFewka (approved):
Looks good. I'd like to switch off of 20.04 in favor of something more recent, but if it's blocking something we can use 20.04 for the time being.

PR #2230, opened 2024-05-21
Debug trace tx changes

skosito:
# Description

This is more for discussion purposes to get your thoughts about this approach.

Transaction tracing works as following:
- tx info is pulled out from cosmos events and parsed
- if tx is found in indexer, it is pulled from block and `MsgEthereumTx` is found in `sdk.Msg` of ethermint tx and passed to grpc query to trace
- also, all predecessors of this msg both from block and cosmos tx (can have multiple `MsgEthereumTx` msgs) are included

(NOTE: There is a weird behavior in ethermint in event parsing  `ParseTxResult` method, where it seems is assumed that there won't be any other msgs other than `MsgEthereumTx` in cosmos tx - i guess this is ok assumption, but technically there could be others? This would impact how `MsgIndex` is calculated there, but that is another topic.)

For zeta transactions that are applied directly (by calling `ApplyMessage` directly, tx type 88) there is no cosmos tx, no sdk msg, but events are emitted and it is indexed, so it can be traced, but code for tracing has to be changed:

- for every tx, use existing `ParseTxResult` method which will parse tx details from cosmos event (i extended event with data and nonce for these type of txs, this data will be useful for adapting other rpc methods as well) - both for predecessors in the block, and predecessors in tx sdk.Msgs
- if tx is type 88, it doesn't have `MsgIndex` because it doesn't have corresponding sdk.Msg so can not get `MsgEthereumTx` from sdk.Msg
- for these tx to get `MsgEthereumTx` i just created eth `LegacyTx` to recreate `MsgEthereumTx` using existing methods

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/x/fungible/keeper/evm.go
+++ b/x/fungible/keeper/evm.go
@@ -755,9 +755,6 @@ func (k Keeper) CallEVMWithData(
 		}
 
 		if !noEthereumTxEvent {
-			if err != nil {
-				return nil, fmt.Errorf("failed to convert tx msg, err=%w", err)
-			}
 			attrs = append(attrs, sdk.NewAttribute("TxData", hexutil.Encode(msg.Data()))) // adding txData for more info in rpc methods
 			attrs = append(attrs, sdk.NewAttribute("TxNonce", fmt.Sprint(nonce)))         // adding nonce for more info in rpc methods
 			ctx.EventManager().EmitEvents(sdk.Events{
--- a/rpc/types/events.go
+++ b/rpc/types/events.go
@@ -183,14 +183,14 @@ func ParseTxResult(result *abci.ResponseDeliverTx, tx sdk.Tx) (*ParsedTxs, error
 	}
 
 	// fix msg indexes, because some eth txs indexed here don't have corresponding sdk.Msg
-	currMsgIndex := -1
+	currMsgIndex := 0
 	for _, tx := range p.Txs {
 		if tx.Type == 88 {
 			tx.MsgIndex = math.MaxUint32
 			// todo: fix mapping as well
 		} else {
-			currMsgIndex++
 			tx.MsgIndex = currMsgIndex
+			currMsgIndex++
 		}
 	}
 	return p, nil


rpc/backend/tracing.go
```diff
@@ -131,6 +150,36 @@ func (b *Backend) TraceTransaction(hash common.Hash, config *evmtypes.TraceConfi
 	return decodedResult, nil
 }
 
+func getMsgEthereumTxFromParsedTx(parsedTx rpctypes.ParsedTx, tx sdk.Tx) *evmtypes.MsgEthereumTx {
+	if parsedTx.Type == 88 {
```
lumtis, line 154: We should precise in comment what does it represent

rpc/backend/tracing.go
```diff
@@ -131,6 +150,36 @@ func (b *Backend) TraceTransaction(hash common.Hash, config *evmtypes.TraceConfi
...
+		recipient := parsedTx.Recipient
+		// TODO: is this ok?
```
lumtis, line 156: I think so, we should explain in comment for signature not set

rpc/backend/tracing.go
```diff
@@ -131,6 +150,36 @@ func (b *Backend) TraceTransaction(hash common.Hash, config *evmtypes.TraceConfi
...
+			GasPrice: nil,
+			Value:    parsedTx.Amount,
+			V:        big.NewInt(0),
+			R:        big.NewInt(0),
+			S:        big.NewInt(0),
+		})
+		ethMsg := &evmtypes.MsgEthereumTx{}
+		err := ethMsg.FromEthereumTx(t)
+		if err != nil {
+			return nil
```
lumtis, line 171: We shouldn't return an error here?

x/fungible/keeper/evm.go
```diff
@@ -756,6 +755,8 @@ func (k Keeper) CallEVMWithData(
 		}
 
 		if !noEthereumTxEvent {
+			attrs = append(attrs, sdk.NewAttribute("TxData", hexutil.Encode(msg.Data()))) // adding txData for more info in rpc methods
+			attrs = append(attrs, sdk.NewAttribute("TxNonce", fmt.Sprint(nonce)))         // adding nonce for more info in rpc methods
```
lumtis, line 759: Comments should be above the line. And I think we can make these more descriptive

PR #2254, opened 2024-05-23
hack: test athens upgrade handler

gartnera:
Hack to test the v17-athens upgrade handler.

https://github.com/zeta-chain/node/actions/runs/9214733457/job/25351583884


gartnera:
Upgrade tests pass

PR #2231, opened 2024-05-21
feat: add no-op v17 athens upgrade handler

gartnera:
# Description

We need to upgrade athens to v17 but not run the migrations. You must define a upgrade handler for any upgrades run via upgrade proposals. Define a no-op `v17-athens` handler.

> SetUpgradeHandler sets an UpgradeHandler for the upgrade specified by name. This handler will be called when the upgrade with this name is applied. In order for an upgrade with the given name to proceed, a handler for this upgrade must be set even if it is a no-op function.

Upgrade proposal will be called `v17-athens`

Related to #2228

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

- [ ] Tested CCTX in localnet
- [ ] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

PR #2255, opened 2024-05-23
fix: set upgrade test new version to v17

gartnera:
Upgrade tests are failing on release/v17 branch: https://github.com/zeta-chain/node/actions/runs/9214972522/job/25352337296

I forgot to set this in #2231


lumtis (approved):
This looks good to me
I think we should also test on a branch with `OLD_VERSION=v16.0.0` for the testnet case?

kingpinXD (approved):

gartnera:
> This looks good to me I think we should also test on a branch with `OLD_VERSION=v16.0.0` for the testnet case?

@lumtis I did that here: https://github.com/zeta-chain/node/pull/2254

Issue #2002, opened 2024-04-09
Investigate and add new chain IDs info in `chains` package depending of potential future considerations


lumtis:
**Describe the Issue**
The chain IDs are currently hardcoded in the `chains` package. We should investigate and implement chain ID for potential new chain integration. This would allow not depending on doing upgrades for integration.

To be added:
- Polygon Testnet Amoy: 80002

PR #2258, opened 2024-05-24
feat: add Optimism and Base in static chain data and rename chain objects

lumtis:
# Description

- Add the following chains in static chain data
  - Optimism mainnet
  - Optimism sepolia
  - Base mainnet
  - Base sepolia

Add `op-stack` as a consensus type

Perform some renaming for chain objects: remove the chain suffix in the names. The objects are already in the `chains` package, namespace. Adding this suffix is unnecessary.

Closes: https://github.com/zeta-chain/node/issues/2002


e2e/e2etests/test_migrate_chain_support.go
```diff
@@ -31,7 +31,7 @@ const EVM2RPCURL = "http://eth2:8545"
 
 // EVM2ChainID is the chain ID for the additional EVM localnet
 // We set Sepolia testnet although the value is not important, only used to differentiate
-var EVM2ChainID = chains.SepoliaChain.ChainId
+var EVM2ChainID = chains.Sepolia.ChainId
```
kingpinXD, line 34: Consider renaming this var to `EVMSeploliaChainID`

proto/zetachain/zetacore/pkg/chains/chains.proto
```diff
@@ -63,31 +78,37 @@ enum Vm {
...
-
+// Chain represents static data about a blockchain network
+// it is identified by a unique chain ID
 message Chain {
-  ChainName chain_name = 1;
+  // ChainId is the unique identifier of the chain
   int64 chain_id = 2;
+
+  // ChainName is the name of the chain
+  ChainName chain_name = 1;
```
kingpinXD, line 98: Is the reason behind moving chain_id up to signify it is the unique identifier ?
Consider maintaining the numerical order , as the chainname is unique as well

pkg/chains/chains.go
```diff
@@ -29,7 +17,9 @@ var (
 		Consensus:   Consensus_tendermint,
 		IsExternal:  false,
 	}
-	EthChain = Chain{
+
+	// Ethereum is Ethereum mainnet
+	Ethereum = Chain{
```
kingpinXD, line 22: NIT: going by the naming, this should be EthereumMainnet . Same for Polygon .

proto/zetachain/zetacore/pkg/chains/chains.proto
lumtis: Yes, chain ID is the identifier, it should be first, I think it is ok using `1` for chain ID as the object is not saved in the chain store

pkg/chains/chains.go
lumtis: There reason for this one is that Ethereum and Polygon use names for their testnet, so `Ethereum` already reflect a mainnet since the testnet is Sepolia
Note: We could also use `Athens3` for ZetaChain but it is a bit ugly and then would become inconsistent with localnet

kingpinXD (approved):
lgtm , left a few NITS

e2e/e2etests/test_migrate_chain_support.go
lumtis: [b3ce9fd](https://github.com/zeta-chain/node/pull/2258/commits/b3ce9fda74f8206eb780436e272d23bd912ef7e5)

Issue #2270, opened 2024-05-28
Implement `ChainInfo` state variable


lumtis:
**Describe the Issue**
`ChainInfo` is used to overwrite or define statis chain data that have not been yet implemented in the `chains` package (as doing so require a chain upgrade to support a new chain)

This task is only about adding the state variable in the store

In the authority module since it would be used by any other module (authority doesn't depend on any module, all module have an authority dep)

Issue #2271, opened 2024-05-28
Implement `MsgUpdateChainInfo`


lumtis:
**Describe the Issue**
Depend on https://github.com/zeta-chain/node/issues/2270

This allow to update the list of static chain data

New values are added for new chain support, values are removed once the chain data are hardcoded after chain upgrade

PR #2275, opened 2024-05-28
Feat/dynamic chain info

lumtis:
# Description

Initialize a new `ChainInfo` state variable in the `authority` store to allow dynamically adding new chain info without upgrade

Closes: https://github.com/zeta-chain/node/issues/2270


lumtis:
E2E tests fail in the CI but pass locally on my machine 🤔

pkg/chains/chain.go
```diff
@@ -14,14 +14,47 @@ type SigninAlgo string
...
+
 // IsEqual compare two chain to see whether they represent the same chain
 func (chain Chain) IsEqual(c Chain) bool {
 	return chain.ChainId == c.ChainId
 }
 
+// IsZetaChain returns true if the chain is a ZetaChain chain
 func (chain Chain) IsZetaChain() bool {
-	return chain.Network == Network_zeta
+	return !chain.IsExternal
```
kingpinXD, line 54: it would be better to keep the older logic as it is more consistent with the function name. 

This check should be under a function called `IsNotExternalChain` or something similar

pkg/chains/chain_test.go
```diff
@@ -11,6 +11,133 @@ import (
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.errStr != "" {
+				require.ErrorContains(t, tt.chain.Validate(), tt.errStr)
+			} else {
+				require.NoError(t, tt.chain.Validate())
+			}
+		})
```
kingpinXD, line 131: consider using an `ErrorAssertionFunc`, to remove this logic and have everything in the tests.

x/authority/types/chain_info.go
```diff
@@ -0,0 +1,30 @@
...
+// DefaultChainInfo returns the structure with an empty list of chains
+func DefaultChainInfo() ChainInfo {
+	return ChainInfo{
+		Chains: []chains.Chain{},
+	}
+}
+
+// Validate performs basic validation of chain info
+// It checks all chains are valid and they're all of external type
+// The strucure is used to store external chain information
```
kingpinXD, line 18:
```suggestion
// The structure is used to store external chain information
```

x/authority/keeper/chain_info.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/zeta-chain/zetacore/x/authority/types"
+)
+
+// SetChainInfo sets the chain info to the store
+func (k Keeper) SetChainInfo(ctx sdk.Context, chainInfo types.ChainInfo) {
```
kingpinXD, line 11: How would this be set?

pkg/chains/chain.go
lumtis: I think we can simply remove `IsExternal` boolean and check everywhere if network is Zeta

x/authority/keeper/chain_info.go
lumtis: https://github.com/zeta-chain/node/issues/2271

pkg/chains/chain_test.go
lumtis: What do you mean having everything in the tests?

proto/zetachain/zetacore/authority/chain_info.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+package zetachain.zetacore.authority;
+
+import "zetachain/zetacore/pkg/chains/chains.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/zeta-chain/zetacore/x/authority/types";
+
+// ChainInfo contains static information about the chains
```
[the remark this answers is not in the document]
lumtis: It's information for chain, information should be plural but plural form of information is information

pkg/chains/chain.go
lumtis: To be addressed in a separate PR, I can revert this change

pkg/chains/chain_test.go
kingpinXD: Using an assertion function would allow you to have the executing loop for the table be something like , and the error assertion logic would be in the table itself,

```
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			tt.error(t, tt.chain.Validate())
		})
	}
```

pkg/chains/chain.go
kingpinXD: Yes that can be done , but might need some more changes for functions using the IsExternal flag
Maybe take that in a separate pr to make sure this doesn't get too big ?

pkg/chains/chain_test.go
lumtis: I saw this before, I don't like this pattern. We make the table more complex to leave the body more simple, which doesn't make sense to me since table driven tests are to make the table visual and easier to maintain

pkg/chains/chain.go
lumtis: >Maybe take that in a separate pr to make sure this doesn't get too big ?
This is what I said

kingpinXD: Okay, lol I think your comment might have been added in when I was reviewing, Resolving this now

kingpinXD (approved):

Issue #2129, opened 2024-05-07
Add a field in static chain data describing the way to transport messages with the external chains


lumtis:
**Describe the Issue**
Currently all chains use the observer/signer set to transport messages with the external chains.
As we will enable IBC for some chains, and there might be other different way to transport data, we should describe somewhere what transport method is used. 

The best candidate looks to be the chain static data since this information will mostly not chain, IBC chain -> use IBC otherwise use observer/signer.

The idea is to introduce a field like `TransportLayer` containing the information

Example:
```
type TransportLayer = OBSERVER | IBC

Ethereum {
  ...
  TransportLayer: OBSERVER
  ...
}
CosmosHub {
  ...
  TransportLayer: IBC
  ...
}
```

PR #2279, opened 2024-05-28
feat(`pkg/chains`): add a `CCTXGateway` field to chain static data

lumtis:
# Description

Closes: https://github.com/zeta-chain/node/issues/2129

NOTE: gonna fix the unit tests later today


proto/zetachain/zetacore/pkg/chains/chains.proto
```diff
@@ -88,6 +88,19 @@ enum Consensus {
   op_stack = 3;
 }
 
+// CCTXGateway describes for the chain the gateway used to handle CCTX outbounds
+enum CCTXGateway {
+  option (gogoproto.goproto_enum_stringer) = true;
+
+  // zevm is the internal CCTX gateway to process outbound on the ZEVM in
+  // ZetaChain only used for ZetaChain chains
```
kingpinXD, line 96: The comment seems inconsistent with the one for observers as they do not talk about the inbound; if we want to treat this as an outbound-only gateway, that should be mentioned in the comments or the naming.

lumtis: For the initial implementation, I think it is simpler to just have a `ValidateInbound` method for crosschain keeper that can be called from anywhere. In the future, I think this would make more sense to only allow calling `ValidateInbound` from a CCTXGateway (like passing a `SetOnInboundValidation(func ...)` method as part of the interface) this would make the design a bit more complex but more intuitive. This is why I wanted to keep the name more general `CCTXGateway` instead of `OutboundGateway`.
But since renaming the interface would be trivial, maybe we can call it `OutboundGateway` for now if this makes clearer

kingpinXD: Okay, that makes sense; maybe adding this to the comments might be enough.

lumtis: [fc6f66b](https://github.com/zeta-chain/node/pull/2279/commits/fc6f66b8b0c9c283ee9e5fe08c57af54e6aebbe3)

Issue #1959, opened 2024-03-28
Fix JSON RPC issues linked to Ethermint


skosito:
**Describe the Issue**
Some json rpc methods do not currently work or work partially for specific tx types (cosmos tx that calls evm, marked as tx with type 88):

- `debug_traceTransaction`
Fails with invalid transaction type *tx.wrapper.

- `eth_getBlockByNumber`
- `eth_getBlockByHash` 

Response should be extended (nonce is 0, input is empty), v,s,r values set to nil - should be set to zero.

- `eth_getTransactionByHash`

Response should be extended (nonce is 0, input is empty).

- `eth_getTransactionReceipt`

Response should be extended (nonce is 0, input is empty, contractAddress and effectiveGasPrice are empty).

- `eth_getTransactionByBlockHashAndIndex`
- `eth_getTransactionByBlockNumberAndIndex`

Fails with code -32000.

**Expected Outcome**
Json rpc methods should work as expected for all tx types.

PR #2282, opened 2024-05-28
feat: modify rpc methods to include synthetic txs

skosito:
# Description

This PR introduces changes in rpc methods to support synthetic ethermint txs (txs of type 88, where ethermint state is changed by directly calling ApplyMessage method from zetacore).

Some details:

EthMsgsFromTendermintBlock method is modified to include all txs, both regular ethermint and synthetic.

This enabled simple implementation of following methods:

- debug_traceTransaction
1. get all eth msgs from block
2. iterate msgs - all before requested tx are predecessors
3. once tx is found, use it for tracing

- debug_traceBlock
1. get all eth msgs from block and trace block

- eth_getTransactionByBlockHashAndIndex
- eth_getTransactionByBlockNumberAndIndex

1. get all eth msgs from block
2. use their order and requested index to find requested tx
3. because index of synthetic txs is always 8888, this is the only way to get both tx types by index

Regarding other tx info methods:

- eth_getTransactionByHash

- eth_getTransactionReceipt

If its regular ethermint tx, it can be decoded from tx. If it's not, recreate it from additional tx fields.

Regarding block methods it seems they can remain the same.

Closes: [1959](https://github.com/zeta-chain/node/issues/1959)

## Type of change

# How Has This Been Tested?

Smoke tests done for athens3 and local txs, more testing in next PR (https://github.com/zeta-chain/node/pull/2282) : move unit tests from ethermint fork here and extend them for these modifications.

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

### DIFF
--- a/rpc/backend/tx_info.go
+++ b/rpc/backend/tx_info.go
@@ -68,15 +68,15 @@ func (b *Backend) GetTransactionByHash(txHash common.Hash) (*rpctypes.RPCTransac
 		}
 		ethMsg = tx.GetMsgs()[res.MsgIndex].(*evmtypes.MsgEthereumTx)
 		if ethMsg == nil {
-			b.logger.Error("failed to get eth msg")
-			return nil, fmt.Errorf("failed to get eth msg")
+			b.logger.Error("failed to get eth msg from sdk.Msgs")
+			return nil, fmt.Errorf("failed to get eth msg from sdk.Msgs")
 		}
 	} else {
 		// if additional fields are not empty try to parse synthetic tx from them
 		ethMsg = b.parseSyntethicTxFromAdditionalFields(additional)
 		if ethMsg == nil {
-			b.logger.Error("failed to parse tx")
-			return nil, fmt.Errorf("failed to parse tx")
+			b.logger.Error("failed to get synthetic eth msg from additional fields")
+			return nil, fmt.Errorf("failed to get synthetic eth msg from additional fields")
 		}
 	}
 


rpc/types/utils.go
```diff
@@ -253,7 +254,7 @@ func NewRPCTransactionFromIncompleteMsg(
 		GasPrice: (*hexutil.Big)(baseFee),
 		Hash:     common.HexToHash(msg.Hash),
 		Input:    []byte{},
-		Nonce:    0, // TODO: get nonce for "from" from ethermint
+		Nonce:    hexutil.Uint64(txAdditional.Nonce), // TODO: get nonce for "from" from ethermint
```
lumtis, line 257: Is the TODO still up to date here as we set a nonce?

rpc/backend/blocks.go
```diff
@@ -296,28 +298,82 @@ func (b *Backend) EthMsgsFromTendermintBlock(
 		tx, err := b.clientCtx.TxConfig.TxDecoder()(tx)
 		if err != nil {
 			b.logger.Debug("failed to decode transaction in block", "height", block.Height, "error", err.Error())
+			// try to check if there is synthetic eth tx in tx result
+			ethMsg, additional := b.parseSyntheticTxFromBlockResults(txResults, i, tx, block)
+			if ethMsg != nil {
+				ethMsgs = append(ethMsgs, ethMsg)
+				txsAdditional = append(txsAdditional, additional)
+			}
```
kingpinXD, line 306: Wondering if it would make the code more readable if we abstract these four lines to a different function , as the same logic is also used in case error == nil , and msg != evmtypes.MsgEthereumTx)

rpc/backend/blocks.go
```diff
@@ -296,28 +298,82 @@ func (b *Backend) EthMsgsFromTendermintBlock(
 		tx, err := b.clientCtx.TxConfig.TxDecoder()(tx)
 		if err != nil {
 			b.logger.Debug("failed to decode transaction in block", "height", block.Height, "error", err.Error())
+			// try to check if there is synthetic eth tx in tx result
+			ethMsg, additional := b.parseSyntheticTxFromBlockResults(txResults, i, tx, block)
```
kingpinXD, line 302: Just for my understanding, is it okay to use tx here even if `err != nil` . Also maybe we should call this txDecoded to prevent shadowing?

rpc/backend/blocks.go
kingpinXD: Also would it possible to call `parseSyntheticTxFromBlockResults` right in the beginning once , and add use the `ethMsg != nil` check , to append values and continue. 
And then check for `evmtypes.MsgEthereumTx ` .

Assuming ofcourse that `tx` cannot be both `MsgEthereumTx` and `SyntheticTX`

rpc/backend/tx_info.go
```diff
@@ -170,26 +176,35 @@ func (b *Backend) GetTransactionReceipt(hash common.Hash) (map[string]interface{
...
 	}
-	tx, err := b.clientCtx.TxConfig.TxDecoder()(resBlock.Block.Txs[res.TxIndex])
-	if err != nil {
-		b.logger.Debug("decoding failed", "error", err.Error())
-		return nil, fmt.Errorf("failed to decode tx: %w", err)
-	}
 
 	var txData evmtypes.TxData
 	var ethMsg *evmtypes.MsgEthereumTx
 	if additional == nil {
```
kingpinXD, line 182:
```suggestion
// if additional fields are empty we can try to parse it as MsgEthereumTx
	if additional == nil {
```

rpc/backend/tx_info.go
```diff
@@ -170,26 +176,35 @@ func (b *Backend) GetTransactionReceipt(hash common.Hash) (map[string]interface{
...
 		txData, err = evmtypes.UnpackTxData(ethMsg.Data)
 		if err != nil {
 			b.logger.Error("failed to unpack tx data", "error", err.Error())
 			return nil, err
 		}
 	} else {
-		ethMsg = &evmtypes.MsgEthereumTx{
-			From: additional.Sender.Hex(),
-			Hash: additional.Hash.Hex(),
+		ethMsg = b.parseSyntethicTxFromAdditionalFields(additional)
```
kingpinXD, line 204:
```suggestion
// Additional fields are present , so we should try to parse this as a synthetic tx
		ethMsg = b.parseSyntethicTxFromAdditionalFields(additional)
```

rpc/types/events.go
```diff
@@ -177,6 +184,18 @@ func ParseTxResult(result *abci.ResponseDeliverTx, tx sdk.Tx) (*ParsedTxs, error
 			p.Txs[i].GasUsed = gasLimit
 		}
 	}
+
+	// fix msg indexes, because some eth txs indexed here don't have corresponding sdk.Msg
+	currMsgIndex := 0
+	for _, tx := range p.Txs {
+		if tx.Type == 88 {
```
kingpinXD, line 191: can we define a constant for 88 ?

x/fungible/keeper/evm.go
```diff
@@ -789,6 +788,10 @@ func (k Keeper) CallEVMWithData(
 		}
 
 		if !noEthereumTxEvent {
```
kingpinXD, line 790: Can we add some comments as to what this flag refers to 
As in when to pass valid and when false. 

Might make more sense to add it to the top of the function though .

kingpinXD (commented):
Looks good , wondering if there is a separate task any plans to add add unit tests for some of the functions added or modified ?

rpc/backend/blocks.go
```diff
@@ -296,30 +299,81 @@ func (b *Backend) EthMsgsFromTendermintBlock(
 		tx, err := b.clientCtx.TxConfig.TxDecoder()(tx)
 		if err != nil {
 			b.logger.Debug("failed to decode transaction in block", "height", block.Height, "error", err.Error())
+			// try to check if there is synthetic eth tx in tx result
```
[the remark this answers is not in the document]
lumtis: ok

rpc/types/utils.go
lumtis: Let's create an issue?

lumtis:
> Looks good , wondering if there is a separate task any plans to add add unit tests for some of the functions added or modified ?

https://github.com/zeta-chain/node/issues/2283

lumtis (approved):

kingpinXD (approved):

rpc/backend/tx_info.go
```diff
@@ -170,26 +172,31 @@ func (b *Backend) GetTransactionReceipt(hash common.Hash) (map[string]interface{
...
+		if ethMsg == nil {
+			b.logger.Error("failed to parse tx")
```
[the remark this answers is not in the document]
lumtis: I think we can make the error a bit more descriptive

Issue #2246, opened 2024-05-23
Add PolicyTable to authority module


kingpinXD:
- Add the `AuthorizationList` struct to the authority module.
```
type Authorization {
 msgUrl string 
 policy PolicyType
}

type AuthorizationList struct {
 policyList []AdminPolicy
}
```

- Define Get, Set, And Check Authorizations methods
- Set Default Values 
- Define Genesis Import and export

PR #2289, opened 2024-05-29
feat: add authorization list for policy transactions

kingpinXD:
# Description
- This pr adds a new authorization list 
- It adds the Get,Set and Genesis methods for the newly added state 
- It defines the type and adds functions to modify the list 

Closes:https://github.com/zeta-chain/node/issues/2246

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


x/authority/genesis.go
```diff
@@ -11,6 +11,10 @@ import (
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
 	k.SetPolicies(ctx, genState.Policies)
 	k.SetChainInfo(ctx, genState.ChainInfo)
+	err := k.SetAuthorizationList(ctx, genState.AuthorizationList)
+	if err != nil {
+		ctx.Logger().Error("Failed to set authorization list in InitGenesis", "error", err)
```
lumtis, line 16: We should panic if initGenesis fails

x/authority/keeper/authorization_list.go
```diff
@@ -0,0 +1,30 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/zeta-chain/zetacore/x/authority/types"
+)
+
+// SetAuthorizationList sets the authorization list to the store.It returns an error if the list is invalid.
+func (k Keeper) SetAuthorizationList(ctx sdk.Context, list types.AuthorizationList) error {
```
lumtis, line 10: It is easier to keep Set functions non-opiniated as a low level function. This is the caller that should ensure the type is valid

This is what we have with all other types

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,196 @@
...
+// it returns an error and the first value of the enum.
+func (a *AuthorizationList) GetAuthorizedPolicy(msgURL string) (PolicyType, error) {
+	for _, auth := range a.Authorizations {
+		if auth.MsgUrl == msgURL {
+			return auth.AuthorizedPolicy, nil
+		}
+	}
+	// Returning first value of enum, can consider adding a default value of `EmptyPolicy` in the enum.
+	return PolicyType(0), ErrAuthorizationNotFound
+}
```
lumtis, line 180: Why not just using `GetAuthorizedPolicy` and return `(policyType, found boolean)`, what `CheckAuthorizationExists` used for?

Even if we want to be future-proof, I don't see a case where there would be another error that can be returned from it

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	"github.com/zeta-chain/zetacore/testutil/sample"
+	"github.com/zeta-chain/zetacore/x/authority/types"
+	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+func TestAuthorizationList_SetAuthorizations(t *testing.T) {
```
lumtis, line 17: Would be better to change to Table-Driven tests. This would also simplify in the review checking the assertions are correct
{
  CurrentTable,
  NewValue,
  ExpectedTable
}

lumtis, line 17: Would be better to make the assertions clear to use values defined in the test, where urls and policyTypes can clearly be seen instead of using `DefaultAuthorizationsList` or `sample`

`sample` is good when need to fill types to be used in test but we should manually define the values we want to assert

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	t.Run("get authorized policy fails when msg not found in list", func(t *testing.T) {
+		authorizationsList := types.DefaultAuthorizationsList()
+		policy, err := authorizationsList.GetAuthorizedPolicy("ABC")
+		require.ErrorIs(t, err, types.ErrAuthorizationNotFound)
+		require.Equal(t, types.PolicyType(0), policy)
+	})
+}
+
+func TestAuthorizationList_CheckAuthorizationExists(t *testing.T) {
```
lumtis, line 65: We should check edge case: empty string

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,196 @@
...
+		authorizationsList := types.DefaultAuthorizationsList()
+		newAuthorization := sample.Authorization()
+		require.False(t, authorizationsList.CheckAuthorizationExists(newAuthorization))
+
+		authorizationsList.SetAuthorizations(newAuthorization)
+		require.True(t, authorizationsList.CheckAuthorizationExists(newAuthorization))
+	})
+}
+
+func TestAuthorizationList_Validate(t *testing.T) {
```
lumtis, line 76: Same, TDT would be more visual

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,196 @@
...
+				MsgUrl:           "ABC",
+				AuthorizedPolicy: types.PolicyType_groupOperational,
+			},
+		}}
+
+		require.ErrorIs(t, authorizationsList.Validate(), types.ErrInValidAuthorizationList)
+	})
+}
+
+func TestAuthorizationList_RemoveAuthorizations(t *testing.T) {
```
lumtis, line 112: Same, TDT would be more visual

x/authority/types/errors.go
```diff
@@ -2,4 +2,8 @@ package types
 
 import errorsmod "cosmossdk.io/errors"
 
-var ErrUnauthorized = errorsmod.Register(ModuleName, 1102, "sender not authorized")
+var (
+	ErrUnauthorized             = errorsmod.Register(ModuleName, 1102, "sender not authorized")
+	ErrInValidAuthorizationList = errorsmod.Register(ModuleName, 1103, "invalid authorization list")
```
lumtis, line 7:
```suggestion
	ErrInvalidAuthorizationList = errorsmod.Register(ModuleName, 1103, "invalid authorization list")
```

x/authority/types/authorizations.go
kingpinXD: I added the tho functions just to make the functions easier to read , IMO its easier to comment on the different uses of the same logic if we have two separate functions .

x/authority/types/authorizations_test.go
kingpinXD: I had originally considered a table , but discarded it as it looked more messy in my opinions specially considering , that we have about 2 test cases per Test 

Here is a sample of what that function would look like, I do prefer to keep this style of tests as in my opinion it looks easier to read , but happy to change if it makes for an easier review 

```
func TestAuthorizationList_SetAuthorizations(t *testing.T) {
	tt := []struct {
		name             string
		oldList          types.AuthorizationList
		addAuthorization types.Authorization
		expectedList     types.AuthorizationList
	}{
		{
			name:             "set new authorization successfully",
			oldList:          types.DefaultAuthorizationsList(),
			addAuthorization: sample.Authorization(),
			expectedList: types.AuthorizationList{
				Authorizations: append(types.DefaultAuthorizationsList().Authorizations, sample.Authorization()),
			},
		},
		{
			name:    "update existing authorization successfully",
			oldList: types.AuthorizationList{Authorizations: []types.Authorization{sample.Authorization()}},
			addAuthorization: types.Authorization{
				MsgUrl:           sample.Authorization().MsgUrl,
				AuthorizedPolicy: types.PolicyType_groupEmergency,
			},
			expectedList: types.AuthorizationList{
				Authorizations: []types.Authorization{
					{
						MsgUrl:           sample.Authorization().MsgUrl,
						AuthorizedPolicy: types.PolicyType_groupEmergency,
					},
				},
			},
		},
	}

	for _, tc := range tt {
		t.Run(tc.name, func(t *testing.T) {
			tc.oldList.SetAuthorizations(tc.addAuthorization)
			require.Equal(t, tc.expectedList, tc.oldList)
		})
	}
}
```

kingpinXD: In my opinion, the default and sample here make the number of lines of code shorter, so it's easier to read. 
Defining the same values again and again might be unnecessary. 

I am guessing you mean creating  list like this in every function 
https://github.com/zeta-chain/zeta-node/blob/e81d37068a35f8a0f3819c0dc6f2a6ab9c0fc965/x/authority/types/authorizations_test.go#L81-L90
Please correct me if you were referring to something else ?

x/authority/genesis.go
kingpinXD: 1fd660e5fbf39783b37daa8c0d4e2a4c356a0aa7

x/authority/keeper/authorization_list.go
lumtis: If we do it, we should address in a separate PR for consistency and I think this is not necessary, the setter/getter are the low-level API for the store to save/retrieve data, this level should not be aware of what represent the data. The validation should always be called in the high-level logic. Validating genesis, validating message, message processing

x/authority/types/authorizations.go
lumtis: I don't see any usage of `CheckAuthorizationExists` in the code outside of tests, so this could be moved to a helper in test, but even for this `GetAuthorizedPolicy` would be better to use because we should also assert the actual policy type in the tests, not just check msg url existence

x/authority/types/authorizations_test.go
lumtis: Test should be clear and efficient, not short.

We had issues in the past of missing cases to check all conditions and in this case TBT makes much easier to see what is being tested, if edge case are missing

I don't think this makes the tests less clear. In you example above we have long lines and usage of sample and default which makes harder to check what is being checked.
If we can use literal values and use new lines for every elements this would make easy to visualize
```
list: {
  {
    url: "foo"
    policy: "operation"
  }
}
newElement: {
  url: "bar"
  policy: "emergency"
}
newList: {
  {
    url: "foo"
    policy: "operation"
  },
  {
    url: "bar"
    policy: "emergency"
  }
}
```
Also in the current test, we don't check that the other elements are left untouched, and with `CheckAuthorizationExists` we don't assert the current policy types.

Finally, TDT allows to easily add new cases so you can check edge cases, here for example we don't check when setting to anempty array

x/authority/types/authorizations.go
kingpinXD: My next PR uses it to check if AuthorizationExists before removing it. 
it looks more readable if the logic is like 
- `CheckAuthorizationExists` -> `RemoveAuthourization`
rather than 
- `GetAuthorizedPolicy` -> `RemoveAuthorization`
But I can remove it here and add it back in if needed

x/authority/types/authorizations_test.go
kingpinXD: 4b198d44901073a5d21399510efc92943d7c01e6

proto/zetachain/zetacore/authority/authorization.proto
```diff
@@ -0,0 +1,18 @@
...
+import "zetachain/zetacore/authority/policies.proto";
+
+option go_package = "github.com/zeta-chain/zetacore/x/authority/types";
+
+message Authorization {
+  // The URL of the message that needs to be authorized
+  string msg_url = 1;
+  // The policy that is authorized to access the message
+  PolicyType authorized_policy = 2;
+}
```
lumtis, line 14:
```suggestion
}

```

proto/zetachain/zetacore/authority/authorization.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+package zetachain.zetacore.authority;
+
+import "gogoproto/gogo.proto";
+import "zetachain/zetacore/authority/policies.proto";
+
+option go_package = "github.com/zeta-chain/zetacore/x/authority/types";
+
+message Authorization {
```
lumtis, line 9: Would be good to add a comment for this type as well

x/authority/keeper/authorization_list.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/zeta-chain/zetacore/x/authority/types"
+)
+
+// SetAuthorizationList sets the authorization list to the store.It returns an error if the list is invalid.
+func (k Keeper) SetAuthorizationList(ctx sdk.Context, list types.AuthorizationList) {
+	store := ctx.KVStore(k.storeKey)
```
lumtis, line 11: We don't use a prefix, meaning it is written at the root of the module store

https://github.com/zeta-chain/node/blob/6423a338eb9a745315e477c0048c95eea5923593/x/authority/keeper/policies.go#L12

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,186 @@
...
+
+import (
+	"fmt"
+
+	"cosmossdk.io/errors"
+)
+
+// DefaultAuthorizationsList list is the list of authorizations that presently exist in the system.
+// This is the minimum set of authorizations that are required to be set when the authorization table is deployed
+func DefaultAuthorizationsList() AuthorizationList {
```
lumtis, line 11: Wondering if we couldn't rather have 3 global vars. `OperationalMsgs`, `EmergencyMsgs` and `AmindMsgs`, then `DefaultAuthorizationsList` computes the array dynamically from the 3 variables.

So looks more visual for the lists

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+func TestAuthorizationList_SetAuthorizations(t *testing.T) {
+	tt := []struct {
+		name             string
+		oldList          types.AuthorizationList
+		addAuthorization types.Authorization
+		expectedList     types.AuthorizationList
+	}{
+		{
```
lumtis, line 25: Can we add a test when `oldList` is empty?
Also might be relevant to add one where we modify an element in the middle of the list

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,306 @@
...
+func TestAuthorizationList_GetAuthorizations(t *testing.T) {
+	tt := []struct {
+		name            string
+		authorizations  types.AuthorizationList
+		getPolicyMsgUrl string
+		expectedPolicy  types.PolicyType
+		error           error
+	}{
+		{
+			name: "get authorizations successfully",
```
lumtis, line 86: Should add tests retrieve a admin and emergency policy
Add a test where we search for an element in a non empty list

x/authority/genesis.go
```diff
@@ -11,6 +11,7 @@ import (
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
 	k.SetPolicies(ctx, genState.Policies)
 	k.SetChainInfo(ctx, genState.ChainInfo)
+	k.SetAuthorizationList(ctx, genState.AuthorizationList)
```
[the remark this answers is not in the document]
kingpinXD: Yeah, we would need a migration, I would be adding that in the next pr

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,186 @@
...
+
+// GetAuthorizedPolicy returns the policy for the given message url.If the message url is not found,
+// it returns an error and the first value of the enum.
+func (a *AuthorizationList) GetAuthorizedPolicy(msgURL string) (PolicyType, error) {
+	for _, auth := range a.Authorizations {
+		if auth.MsgUrl == msgURL {
+			return auth.AuthorizedPolicy, nil
+		}
+	}
+	// Returning first value of enum, can consider adding a default value of `EmptyPolicy` in the enum.
```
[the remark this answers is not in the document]
kingpinXD: I was leaning more towards introducing a `PolicyType_None`

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,306 @@
...
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			policy, err := tc.authorizations.GetAuthorizedPolicy(tc.getPolicyMsgUrl)
+			require.ErrorIs(t, err, tc.error)
+			require.Equal(t, tc.expectedPolicy, policy)
+		})
+	}
+}
+
+func TestAuthorizationList_Validate(t *testing.T) {
```
lumtis, line 121: Need to check for empty list.

Also I think we can check Default list is valid as well. Here there is not a specific assertion to check, this is more like an integration test

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,186 @@
...
+		},
+	}
+
+	return AuthorizationList{
+		Authorizations: authorizations,
+	}
+}
+
+// SetAuthorizations adds the authorization to the list. If the authorization already exists, it updates the policy.
+func (a *AuthorizationList) SetAuthorizations(authorization Authorization) {
```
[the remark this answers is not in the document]
kingpinXD: Yes we need the ability to modify , update or remove authorizations , this would be controlled by the admin policy address

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,306 @@
...
+		}}
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.authorizations.Validate()
+			require.ErrorIs(t, err, tc.expectedError)
+		})
+	}
+}
+
+func TestAuthorizationList_RemoveAuthorizations(t *testing.T) {
```
lumtis, line 181: Need to check when removing from empty list, nothing happen

x/authority/types/authorizations_test.go
```diff
@@ -0,0 +1,306 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			tc.oldList.RemoveAuthorizations(tc.removeAuthorization)
+			require.Equal(t, tc.expectedList, tc.oldList)
+		})
+	}
+}
+
+func TestDefaultAuthorizationsList(t *testing.T) {
+	t.Run("default authorizations list", func(t *testing.T) {
+		var OperationalPolicyMessageList = []string{
```
lumtis, line 239: These are the variable I think it makes sense to define as global var

x/authority/types/authorizations.go
kingpinXD: c18bd7d3776754598a3ebc3a527995e71faf7f15

x/authority/types/authorizations_test.go
kingpinXD: c18bd7d3776754598a3ebc3a527995e71faf7f15

What do you mean by any integration test ?

x/authority/types/authorizations_test.go
kingpinXD: Added the global variables for building the DefaultMsgList , but I would prefer to keep this in the test so that the test fails if any of the values gets changed

x/authority/types/authorizations_test.go
lumtis: By testing `DefaultAuthorizationList` we don't test the validate function itself but the integration between default list and validate function

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
+package types
+
+import (
+	"fmt"
+
+	"cosmossdk.io/errors"
+)
+
+var (
+	OperationPolicyMessages = []string{
```
lumtis, line 10:
```suggestion
	// OperationPolicyMessages keeps track of the message URLs that can, by default, only be executed by operational policy address
	OperationPolicyMessages = []string{
```

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+		"/zetachain.zetacore.fungible.MsgUnpauseZRC20",
+		"/zetachain.zetacore.observer.MsgAddObserver",
+		"/zetachain.zetacore.observer.MsgRemoveChainParams",
+		"/zetachain.zetacore.observer.MsgResetChainNonces",
+		"/zetachain.zetacore.observer.MsgUpdateChainParams",
+		"/zetachain.zetacore.observer.MsgEnableCCTX",
+		"/zetachain.zetacore.observer.MsgUpdateGasPriceIncreaseFlags",
+		"/zetachain.zetacore.lightclient.MsgEnableHeaderVerification",
+	}
+	AdminPolicyMessages = []string{
```
lumtis, line 29:
```suggestion
	// AdminPolicyMessages keeps track of the message URLs that can, by default, only be executed by admin policy address
	AdminPolicyMessages = []string{
```

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+		"/zetachain.zetacore.crosschain.MsgMigrateTssFunds",
+		"/zetachain.zetacore.crosschain.MsgUpdateTssAddress",
+		"/zetachain.zetacore.fungible.MsgUpdateContractBytecode",
+		"/zetachain.zetacore.fungible.MsgUpdateSystemContract",
+		"/zetachain.zetacore.observer.MsgUpdateObserver",
+	}
+	EmergencyPolicyMessages = []string{
```
lumtis, line 36:
```suggestion
	// EmergencyPolicyMessages keeps track of the message URLs that can, by default, only be executed by emergency policy address
	EmergencyPolicyMessages = []string{
```

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+		index++
+	}
+
+	return AuthorizationList{
+		Authorizations: authorizations,
+	}
+}
+
+// SetAuthorizations adds the authorization to the list. If the authorization already exists, it updates the policy.
+func (a *AuthorizationList) SetAuthorizations(authorization Authorization) {
```
lumtis, line 83: Should be renamed `SetAuthorization`

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+		if auth.MsgUrl == authorization.MsgUrl {
+			a.Authorizations[i].AuthorizedPolicy = authorization.AuthorizedPolicy
+			return
+		}
+	}
+	a.Authorizations = append(a.Authorizations, authorization)
+}
+
+// RemoveAuthorizations removes the authorization from the list. It does not check if the authorization exists or not.
+func (a *AuthorizationList) RemoveAuthorizations(authorization Authorization) {
```
lumtis, line 94: Should be renamed `RemoveAuthorization`

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+	for i, auth := range a.Authorizations {
+		if auth.MsgUrl == authorization.MsgUrl {
+			a.Authorizations = append(a.Authorizations[:i], a.Authorizations[i+1:]...)
+		}
+	}
+}
+
+// GetAuthorizedPolicy returns the policy for the given message url.If the message url is not found,
```
lumtis, line 102:
```suggestion
// GetAuthorizedPolicy returns the policy for the given message url. If the message url is not found,
```

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+// it returns an error and the first value of the enum.
```
lumtis, line 103:
```
and the first value of the enum.
```
This comment is not necessary. When a function returns an error, all the other values should not be considered by the calling function, so its value doesn't matter

x/authority/types/authorizations.go
```diff
@@ -0,0 +1,129 @@
...
+func (a *AuthorizationList) GetAuthorizedPolicy(msgURL string) (PolicyType, error) {
+	for _, auth := range a.Authorizations {
+		if auth.MsgUrl == msgURL {
+			return auth.AuthorizedPolicy, nil
+		}
+	}
+	fmt.Println("Authorization not found", msgURL)
```
lumtis, line 110:
```suggestion
```
Was it for debugging?

x/authority/types/keys.go
```diff
@@ -24,7 +24,8 @@ func KeyPrefix(p string) []byte {
 
 const (
 	// PoliciesKey is the key for the policies store
-	PoliciesKey = "Policies-value-"
+	PoliciesKey          = "Policies-value-"
+	AuthorizationListKey = "AuthorizationList-value-"
```
lumtis, line 28: Need comment

lumtis (approved):

PR #2306, opened 2024-06-03
refactor: removed unused code and unified tx signing logic

ws4charlie:
# Description

The PR is a minimum refactor to prepare for EIP-1559 outbound transaction

- removed unused method `SignWhitelistTx`
- unified different tx signing (onReceive, on Revert, erc20 withdraw, gas token withdraw) to use a common method `Sign`.
- unified the tx verification logic in unit tests; added missing unit tests for `SignCancelTx` and `TestSigner_SignMigrateTssFundsCmd`
- fix 'amount' issue in `SignMigrateTssFundsCmd`

Closes: <PD-XXXX>

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


zetaclient/chains/evm/signer/signer.go
```diff
@@ -207,12 +213,13 @@ func (signer *Signer) SignOutbound(txData *OutboundData) (*ethtypes.Transaction,
 
 	tx, _, _, err := signer.Sign(data,
 		signer.zetaConnectorAddress,
+		big.NewInt(0),
```
kingpinXD, line 216: Consider defining the zero amount somewhere as a constant.

kingpinXD (approved):

zetaclient/chains/evm/signer/signer.go
```diff
@@ -39,7 +39,10 @@ import (
 	"github.com/zeta-chain/zetacore/zetaclient/zetacore"
 )
 
-var _ interfaces.ChainSigner = &Signer{}
+var (
+	zeroValue                        = int64(0)
```
lumtis, line 43: Seems like to be only used as big.Int, why not directly defining it as big.Int?

Also need comment why the variable is necessary

zetaclient/chains/evm/signer/signer.go
```diff
@@ -241,74 +246,57 @@ func (signer *Signer) SignRevertTx(txData *OutboundData) (*ethtypes.Transaction,
...
+		signer.tssSigner.EVMAddress(),
+		big.NewInt(zeroValue), // zero out the amount to cancel the tx
+		evm.EthTransferGasLimit,
+		txData.gasPrice,
+		txData.nonce,
+		txData.height,
+	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("SignCancelTx error: %w", err)
```
lumtis, line 278: Not covered by test

zetaclient/chains/evm/signer/signer.go
```diff
@@ -241,74 +246,57 @@ func (signer *Signer) SignRevertTx(txData *OutboundData) (*ethtypes.Transaction,
...
+		txData.to,
+		txData.amount,
+		evm.EthTransferGasLimit,
+		txData.gasPrice,
+		txData.nonce,
+		txData.height,
+	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("SignWithdrawTx error: %w", err)
```
lumtis, line 296: Not covered by test, sme for other sign error below

lumtis (commented):
This looks legit to me. Some few comments.

Wondering if we could think about a further refactoring with a declarative approach?
We have a single sign function and an array containing the different values depending on the tx to sign.
Then to support a new tx that can be signed, you only need to add an entry in this array

zetaclient/chains/evm/signer/signer.go
ws4charlie: added more tests in https://github.com/zeta-chain/node/pull/2306/commits/3db4bef06ed6dee6a142efe5152d1704d9084793

zetaclient/chains/evm/signer/signer.go
ws4charlie: same commit https://github.com/zeta-chain/node/pull/2306/commits/3db4bef06ed6dee6a142efe5152d1704d9084793

ws4charlie:
> This looks legit to me. Some few comments.
> 
> Wondering if we could think about a further refactoring with a declarative approach? We have a single sign function and an array containing the different values depending on the tx to sign. Then to support a new tx that can be signed, you only need to add an entry in this array

I think we can create a issue to track this: https://github.com/zeta-chain/node/issues/2318

lumtis (approved):

PR #2308, opened 2024-06-03
refactor(e2e): move addresses and private keys to config

### LINKED ISSUE #2238: Allow randomizing localnet secrets
I'm using the localnet docker compose setup for developnet. Since developnet is public, we need to randomize the secrets we use.

Use a depends_on container with `service_completed_successfully` condition to generate secrets and store on a volume during the first run.

gartnera:
# Description

Move all localnet secrets to the e2e config file to allow easy randomization. I still need to finish the TODO items, but please take an initial look. Notes:

- `--config` is now a required parameter to zetae2e. `DefaultConfig()` is no longer used for localnet e2e testing.
- `zetae2e init` now generates keypairs

TODO:

- [x] move from hardcoded addresses in `start-zetae2e.sh` to loading from config file
- [x] move from hardcoded addresses in `start-zetacored.sh` to loading from config file

Closes #2238 

Relates to DEVOP-642

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests


cmd/zetae2e/config/localnet.yml
```diff
@@ -0,0 +1,55 @@
+zeta_chain_id: "athens_101-1"
+accounts:
```
lumtis, line 2: To me it seems we're adding configuration opiniated to local tests. So far the config was fully configured whether you use for local or live networks.

I think we should either:
- Make the account list more generic like having a `additional_accounts` array `{name, address, private_key}`
- Putting this account in a explicit field like `localnet_accounts` to make it clear these doesn't need to be initialized for live network tests

lumtis (commented):
On https://github.com/zeta-chain/node/issues/2238

We want to use the `zetae2e local` command to initialize/test the devnet?
It might make sense to rename `local` into `dev` or something similar then IMO

e2e/config/config.go
```diff
@@ -1,26 +1,41 @@
...
+	RPCs                  RPCs      `yaml:"rpcs"`
+	Contracts             Contracts `yaml:"contracts"`
+	ZetaChainID           string    `yaml:"zeta_chain_id"`
+	FungibleAdminMnemonic string    `yaml:"fungible_admin_mnemonic"`
+}
+
+// Account contains configuration for an account
+type Account struct {
+	RawAddress    string `yaml:"address"`
+	RawPrivateKey string `yaml:"private_key"`
```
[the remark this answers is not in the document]
gartnera: Because naming it just `Address` would be confusing as it would conflict with the `Address()` method.

I could also call it `EncodedAddress` or `AddressString` or something like that.

Or I could rename the methods `GetAddress()` or `DecodeAddress()` or `GetEVMAddress()`

cmd/zetae2e/config/localnet.yml
gartnera: I think I prefer the first option as some of these tests could be adjusted to run on live networks in the future. I think this should be a general goal with all e2e tests.

> Make the account list more generic like having

That was my original approach in d2cb7a36130318ad5327fae33da209588dea0a95 but the loss of type safety did not feel quite right. Please advise if you prefer that approach.

lumtis: > I prefer a combination of both options where we use a strongly typed structure but move the optional accounts into a separate additional_accounts field.

Sounds good, I was thinking about using `localnet_accounts` but if we can consider using those in live networks in the future, `additional_accounts ` works

lumtis:
Is this one ready for re-review @gartnera ?

lumtis (approved):
We might want to introduce a ignore file for GitGuardian if possible

e2e/config/config.go
```diff
@@ -1,26 +1,73 @@
...
+	RawPrivateKey    DoubleQuotedString `yaml:"private_key"`
 }
 
-// Accounts contains the configuration for the accounts
+// Accounts are the required accounts to run any e2e test
 type Accounts struct {
-	EVMAddress string `yaml:"evm_address"`
-	EVMPrivKey string `yaml:"evm_priv_key"`
+	Deployer Account `yaml:"deployer"`
+}
```
gartnera, line 59: Maybe this should just be a top level value. Deployer isn't a great name anyway.

Maybe `DefaultAccount Account`

lumtis: Yes we should rename it

PR #2350, opened 2024-06-19
refactor(e2e): use private key rather than mnemonic in admin tests

gartnera:
# Description

#2308 is getting a bit large, so I'm gonna start breaking stuff out of it.

- use private key rather than pneumonic for fungible admin
- remove fungible admin parameters from e2e instead provide an optional `ZetaTxServer` 
- only populate the `ZetaTxServer` when it's actually needed (admin tests)

Part of https://github.com/zeta-chain/node/issues/2238

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/contrib/localnet/docker-compose-admin.yml
+++ b/contrib/localnet/docker-compose-admin.yml
@@ -13,6 +13,7 @@ services:
       context: ./anvil
     container_name: eth2
     hostname: eth2
+    platform: linux/amd64
     ports:
       - "8546:8545"
     networks:
--- a/contrib/localnet/scripts/start-zetacored.sh
+++ b/contrib/localnet/scripts/start-zetacored.sh
@@ -219,11 +219,11 @@ then
   fi
 
 # set admin account
-  zetacored add-genesis-account zeta1srsq755t654agc0grpxj4y3w0znktrpr9tcdgk 100000000000000000000000000azeta
+  zetacored add-genesis-account zeta1svzuz982w09vf2y08xsh8qplj36phyz466krj3 100000000000000000000000000azeta
   zetacored add-genesis-account zeta1n0rn6sne54hv7w2uu93fl48ncyqz97d3kty6sh 100000000000000000000000000azeta # Funds the localnet_gov_admin account
-  cat $HOME/.zetacored/config/genesis.json | jq '.app_state["authority"]["policies"]["items"][0]["address"]="zeta1srsq755t654agc0grpxj4y3w0znktrpr9tcdgk"' > $HOME/.zetacored/config/tmp_genesis.json && mv $HOME/.zetacored/config/tmp_genesis.json $HOME/.zetacored/config/genesis.json
-  cat $HOME/.zetacored/config/genesis.json | jq '.app_state["authority"]["policies"]["items"][1]["address"]="zeta1srsq755t654agc0grpxj4y3w0znktrpr9tcdgk"' > $HOME/.zetacored/config/tmp_genesis.json && mv $HOME/.zetacored/config/tmp_genesis.json $HOME/.zetacored/config/genesis.json
-  cat $HOME/.zetacored/config/genesis.json | jq '.app_state["authority"]["policies"]["items"][2]["address"]="zeta1srsq755t654agc0grpxj4y3w0znktrpr9tcdgk"' > $HOME/.zetacored/config/tmp_genesis.json && mv $HOME/.zetacored/config/tmp_genesis.json $HOME/.zetacored/config/genesis.json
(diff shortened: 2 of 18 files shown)


contrib/localnet/docker-compose-admin.yml
```diff
@@ -13,6 +13,7 @@ services:
       context: ./anvil
     container_name: eth2
     hostname: eth2
+    platform: linux/amd64
```
gartnera, line 16: fixes this:

```
cd contrib/localnet/ && docker compose -f docker-compose.yml -f docker-compose-admin.yml up -d
WARN[0000] /home/alex/workspace/github.com/zeta-chain/node/contrib/localnet/docker-compose.yml: `version` is obsolete
WARN[0000] /home/alex/workspace/github.com/zeta-chain/node/contrib/localnet/docker-compose-admin.yml: `version` is obsolete
[+] Building 1.2s (2/2) FINISHED                                                                                                                                                                       docker:default
 => [eth2 internal] load build definition from Dockerfile                                                                                                                                                        0.2s
 => => transferring dockerfile: 565B                                                                                                                                                                             0.0s
 => ERROR [eth2 internal] load metadata for ghcr.io/foundry-rs/foundry:latest                                                                                                                                    0.9s
------
 > [eth2 internal] load metadata for ghcr.io/foundry-rs/foundry:latest:
------
failed to solve: ghcr.io/foundry-rs/foundry:latest: failed to resolve source metadata for ghcr.io/foundry-rs/foundry:latest: no match for platform in manifest: not found
make: *** [Makefile:226: start-e2e-admin-test] Error 17
```

e2e/e2etests/e2etests.go
```diff
@@ -513,6 +507,12 @@ var AllE2ETests = []runner.E2ETest{
 		[]runner.ArgDefinition{},
 		TestRateLimiter,
 	),
+	runner.NewE2ETest(
+		TestERC20DepositAndCallRefundName,
```
gartnera, line 511: this test uses admin apis so it should run on the admin test runner

lumtis:
I guess we can close https://github.com/zeta-chain/node/pull/2308 ?

e2e/e2etests/e2etests.go
lumtis: Good catch.
Although the admin functionality used doesn't interfere with the network so would be fine to keep it in regular tests.

The idea of separating admin tests is that functions that interfere with network functionality (pausing cctxs for example) can't be run in parallel with the rest of the tests, hence running separately. And they also tests scoped functionalities in the protocol.

We might want to do some refactoring or renaming here, calling "admin special tests" for example.

The other thing with test using admin API is that it can be used for tests on live network. For this I think we should introduce a new boolean in the E2E test object to specify if the test can be run on live network or not.

lumtis (approved):
LGTM

I think we can extend the functional option pattern for other options in the runner struct in the future

contrib/localnet/scripts/start-zetacored.sh
```diff
@@ -219,11 +219,11 @@ then
   fi
 
 # set admin account
-  zetacored add-genesis-account zeta1srsq755t654agc0grpxj4y3w0znktrpr9tcdgk 100000000000000000000000000azeta
+  zetacored add-genesis-account zeta1svzuz982w09vf2y08xsh8qplj36phyz466krj3 100000000000000000000000000azeta
```
kingpinXD, line 222: Lets define a variable fo this address

e2e/runner/runner.go
```diff
@@ -33,17 +33,24 @@ import (
 	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
 )
 
+type E2ERunnerOption func(*E2ERunner)
```
kingpinXD, line 36: NIT : `E2ERunnerWithOptions` or `E2ERunnerOptions`

contrib/localnet/scripts/start-zetacored.sh
gartnera: Yeah I will do that in my upcoming PRs based on https://github.com/zeta-chain/node/pull/2308

e2e/runner/runner.go
gartnera: I don't think this should be plural unless it was a slice type or a struct with multiple members.

It's singular but becomes plural when you use it as a varadic argument.

Based on a quick survey, this is typically singular:
- https://medium.com/@omid.jn/the-option-pattern-in-go-9edf55e88320
- https://rednafi.com/go/dysfunctional_options_pattern/
- https://www.yellowduck.be/posts/the-functional-options-pattern-in-go

e2e/e2etests/e2etests.go
gartnera: Ok I can move it back to the erc20 test runner, I will just need to inject the txserver into the runner.

gartnera:
> I guess we can close #2308 ?

I still have 2-3 more PRs worth of changes I need to pull out of that PR.

kingpinXD (approved):

Issue #2259, opened 2024-05-24
Add command to run e2e test on imported data from mainnet (Port from v16)


kingpinXD:
Add a make command that does the following 
- Download the latest Genesis snapshot from `https://snapshots.zetachain.com/latest-state-export`
- Import the required data into the genesis file for e2e test network 
- Start the network and run the e2e test on it.

PR #2353, opened 2024-06-19
refactor: use docker healthcheck for localnet e2e

gartnera:
# Description

Use a docker health check rather than sleeping in e2e to wait for e2e start. Also reduce keygen height.

The keygen height parameters are very sensitive, the following combinations seem to result in random test failures:
- 20 25
- 20 30
- 25 30

I have selected parameters which appear to be stable, but I will need a few CI runs to know for sure.

Reduced CI time by 30-60 seconds, but is mostly to make local development faster.

Update: need to accelerate preparams generation a bit. Also cache it on a docker volume to avoid pointless recomputation when running locally.

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

- [ ] Tested CCTX in localnet
- [ ] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

PR #2360, opened 2024-06-20
test: add stateful end to end test

kingpinXD:
# Description

Ports changes from v16

Closes: https://github.com/zeta-chain/node/issues/2259

## Type of change

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:


cmd/zetacored/parse_genesis.go
```diff
@@ -124,7 +141,7 @@ func ImportDataIntoFile(genDoc *types.GenesisDoc, importFile *types.GenesisDoc,
 		if Copy[m] {
 			appState[m] = importAppState[m]
 		}
-		if Modify[m] {
+		if Modify[m] && modifyEnabled {
```
kingpinXD, line 144: Disabling modify as some of the proto files have changed since v17. 
Examples 
`inbound_tx_params`  -> `inbound_params`
`OutBoundTx` -> `OutboundTx`

contrib/localnet/docker-compose-import-data.yml
```diff
@@ -0,0 +1,37 @@
...
+      - ~/genesis_export/:/root/genesis_data
+
+  zetacore1:
+    entrypoint: ["/root/start-zetacored.sh", "2","import-data"]
+
+  zetaclient0:
+    entrypoint: ["/root/start-zetaclientd.sh", "import-data"]
+
+  zetaclient1:
+    entrypoint: ["/root/start-zetaclientd.sh", "import-data"]
```
gartnera, line 34: I don't think this is necessary? I don't see it used anywhere in zetaclient

contrib/localnet/docker-compose-import-data.yml
```diff
@@ -0,0 +1,37 @@
...
+# - An Ethereum node (eth)
+# - A Bitcoin node (bitcoin)
+# - A Rosetta API (rosetta)
+# - An orchestrator to manage interaction with the localnet (orchestrator)
+
+networks:
+  mynetwork:
+    ipam:
+      config:
+        - subnet: 172.20.0.0/24
```
gartnera, line 16: I don't think this is needed

contrib/localnet/scripts/start-zetacored.sh
gartnera, line 1: Don't we need to update the keygen block here too depending on the imported height?

gartnera (commented):
This will conflict slightly with #2353. We will need to tune the healthcheck parameters for this scenario.

contrib/localnet/scripts/start-zetacored.sh
kingpinXD: Why? The import happens before the network starts, so the keygen should still be fine .

gartnera: Wouldn't the height be at whatever the snapshot height is?

gartnera: Ok I guess not, this is genesis state so it will start at 0.

gartnera, line 1: I don't think there is much value to importing the state without doing an upgrade. The upgrade is the main thing we want to test. Plus, then you should have to do less work to modify the state before import right? Since the state will be exported from the same version as the import.

kingpinXD: Yeah I agree , the upgrade test is better , that would be added in a separate pr though

gartnera: This test will be really hard to maintain because you have to write dedicated migration logic for this specifically right?

Whereas the upgrade test will just rely on the normal upgrade logic that already has to be written.

lumtis (approved):

Issue #2273, opened 2024-05-28
Use `ChainInfo` in `authority` to get the list of static chain data


lumtis:
**Describe the Issue**
Depends on https://github.com/zeta-chain/node/issues/2272

In many part of the blcokchain and ZetaClient, we use `DefaultChainList` and `GetChainFromChainID` to get static information about chains (like consensus, etc...)

Currently the function fetch from a hardcoded list of chain info.

With the new `ChainInfo` structure in the `authority` module, new chain info can dynamically be added.

We should query this structure (along with using the hardcoded list) to get static chain info from chain ID.

The final list should be determined by using the hardcoded list and adding elements from authority's ChainInfo into this list, chain info in ChainInfo structure can overwrite the values in the hardcoded list

Issue #2080, opened 2024-04-25
`zetacored` : Remove pointers from functions returning a list of chains


kingpinXD:
Functions returning a list of chains, based on certain conditions, return pointers to the chains 
Example 
```
DefaultChainsList() []*Chain
ChainListByNetworkType(networkType NetworkType) []*Chain
ChainListByNetwork(network Network) []*Chain 
ExternalChainList() []*Chain
ChainListByConsensus(consensus Consensus) []*Chain
ChainListForHeaderSupport() []*Chain 
```

These can instead return chains directly , the pointers are not necessary for this. usecase

PR #2380, opened 2024-06-25
refactor: use `ChainInfo` in `authority` to allow support chains dynamically

lumtis:
# Description

Use the `ChainList` object to allow supporting chains where static info are added dynamically, without being hardcoded.

Mainly:
- Modify the method in the `chains` package to support using a list of additional chains from ones hardcoded
- Read the chain list directly from the authority keeper in the protocol when calling chains methods
- Add the additional chains in core context and use it in ZetaClient

On the design choice, the main idea is to keep the purity of the functions in `chains` package and keep using these methods.

Closes: https://github.com/zeta-chain/node/issues/2273


pkg/chains/chain.go
```diff
@@ -87,99 +76,68 @@ func (chain Chain) EncodeAddress(b []byte) (string, error) {
...
 	}
 	return nil, fmt.Errorf("chain (%d) not supported", chainID)
 }
 
 // IsEVMChain returns true if the chain is an EVM chain or uses the ethereum consensus mechanism for block finality
-func IsEVMChain(chainID int64) bool {
-	return ChainIDInChainList(chainID, ChainListByConsensus(Consensus_ethereum))
+// additionalChains is a list of additional chains to search from
+// in practice, it is used in the protocol to dynamically support new chains without doing an upgrade
+func IsEVMChain(chainID int64, additionalChains []Chain) bool {
```
lumtis, line 94: For this function and following ones it could be justified to use variadic parameter `...Chain` to simplify the signature and make it more visual when not providing chain

The motivation keeping a required argument is that, in practice, outside of tests, we'll always provide a custom value.
- GetChainList from authority in the protocol
- AdditionalChains in ZetaClient
Only in test as placeholder we provide `[]Chain{}`

x/crosschain/types/inbound_params.go
```diff
@@ -2,19 +2,13 @@ package types
...
-
-	"github.com/zeta-chain/zetacore/pkg/chains"
 )
 
 func (m InboundParams) Validate() error {
 	if m.Sender == "" {
 		return fmt.Errorf("sender cannot be empty")
 	}
 
-	if chains.GetChainFromChainID(m.SenderChainId) == nil {
```
lumtis, line 14: We're removing here and other places in `types` like packages checks for chain ID validity.

The reasoning is that chain ID validity check is no longer stateless as chain IDs can be added dynamically as part of the chain state

lumtis:
@coderabbitai review

lumtis:
https://github.com/coderabbitai review

lumtis:
@coderabbitai review

lumtis:
Seems like @coderabbitai won't trigger for this one 🤔

x/lightclient/types/message_disable_verification_flags.go
```diff
@@ -46,21 +44,12 @@ func (msg *MsgDisableHeaderVerification) ValidateBasic() error {
...
-	for _, chainID := range msg.ChainIdList {
-		if !chains.ChainIDInChainList(chainID, chainListForHeaderSupport) {
-			return cosmoserrors.Wrapf(
-				sdkerrors.ErrInvalidRequest,
-				"invalid chain id header not supported (%d)",
-				chainID,
-			)
-		}
+
+	if len(msg.ChainIdList) > 200 {
```
swift1337, line 51: What does that mean?

cmd/zetaclientd/debug.go
```diff
@@ -70,36 +72,39 @@ func DebugCmd() *cobra.Command {
...
-			chain := chains.GetChainFromChainID(chainID)
+
+			chain := chains.GetChainFromChainID(chainID, coreContext.GetAdditionalChains())
 			if chain == nil {
 				return fmt.Errorf("invalid chain id")
 			}
 
-			// get ballot identifier according to the chain type
-			if chains.IsEVMChain(chain.ChainId) {
+			// get ballot identifier according to the chain type\
```
ws4charlie, line 91:
```suggestion
			// get ballot identifier according to the chain type
```

cmd/zetaclientd/debug.go
```diff
@@ -70,36 +72,39 @@ func DebugCmd() *cobra.Command {
...
 				return err
 			}
+
 			tssEthAddress, err := client.GetEthTssAddress()
 			if err != nil {
 				return err
 			}
-			chain := chains.GetChainFromChainID(chainID)
+
+			chain := chains.GetChainFromChainID(chainID, coreContext.GetAdditionalChains())
```
ws4charlie, line 86: The term `AdditionalChains` confused me a bit at first glance. Do you mean something like a 'ChainBook' or 'ChainDictory'?

pkg/chains/chain.go
```diff
@@ -87,99 +76,68 @@ func (chain Chain) EncodeAddress(b []byte) (string, error) {
...
-func (chain Chain) DecodeAddress(addr string) ([]byte, error) {
-	return DecodeAddressFromChainID(chain.ChainId, addr)
-}
-
 // DecodeAddressFromChainID decode the address string to bytes
-func DecodeAddressFromChainID(chainID int64, addr string) ([]byte, error) {
-	if IsEVMChain(chainID) {
+// additionalChains is a list of additional chains to search from
+// in practice, it is used in the protocol to dynamically support new chains without doing an upgrade
+func DecodeAddressFromChainID(chainID int64, addr string, additionalChains []Chain) ([]byte, error) {
```
ws4charlie, line 82: Same, the parameter name `additionalChains` could confuse devs. The only names I could think of was `chainBook` or `chainDictionary`

pkg/chains/chains.go
```diff
@@ -368,14 +376,47 @@ func ZetaChainFromChainID(chainID string) (Chain, error) {
...
+// duplicated chain ID are overwritten by the second list
+func CombineDefaultChainsList(chains []Chain) []*Chain {
+	chainsPtr := ChainListPointers(chains)
+
+	return CombineChainList(DefaultChainsList(), chainsPtr)
+}
+
+// CombineChainList combines a list of chains with a list of chains
+// duplicated chain ID are overwritten by the second list
+func CombineChainList(chains1 []*Chain, chains2 []*Chain) []*Chain {
```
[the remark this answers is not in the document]
ws4charlie: I think converting the pointers of `*Chain` fetched from RPC to actual value of `Chain` makes sense. The downstream code won't need to double check a `nil` pointer. Copying a small Chain struct won't cause performance issue.

pkg/chains/chains.go
```diff
@@ -368,14 +376,47 @@ func ZetaChainFromChainID(chainID string) (Chain, error) {
...
+	combined := make([]*Chain, 0)
+	combined = append(combined, chains1...)
+
+	// map chain ID in chains1 to index in the list
```
ws4charlie, line 397: Seems the comment doesn't match the loop below. The loop below is not iterating on `chains1`.

x/authority/keeper/chain_info.go
```diff
@@ -24,3 +25,13 @@ func (k Keeper) GetChainInfo(ctx sdk.Context) (val types.ChainInfo, found bool)
 	k.cdc.MustUnmarshal(b, &val)
 	return val, true
 }
+
+// GetAdditionalChainList returns the list of chains in chain info object
```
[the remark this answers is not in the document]
ws4charlie: Seems to me there are two source of truth in this `AdditionalChainList` model?

I was thinking what would happen if both list have a same int64 field `chainID` but differ on other fields when we combining the `DefaultChainList` and and `AdditionalChainList`. Not a blocker.

x/crosschain/types/message_add_inbound_tracker.go
```diff
@@ -48,17 +47,7 @@ func (msg *MsgAddInboundTracker) ValidateBasic() error {
...
-		return errorsmod.Wrapf(ErrInvalidChainID, "chain id (%d)", msg.ChainId)
-	}
-	if msg.Proof != nil && !chain.SupportMerkleProof() {
-		return errorsmod.Wrapf(
-			ErrProofVerificationFail,
-			"chain id %d does not support proof-based trackers",
-			msg.ChainId,
-		)
-	}
+
```
ws4charlie, line 50: The method `SupportMerkleProof` is used no where. Should we remove?

x/lightclient/types/message_enable_verification_flags.go
```diff
@@ -46,21 +44,12 @@ func (msg *MsgEnableHeaderVerification) ValidateBasic() error {
...
-	for _, chainID := range msg.ChainIdList {
-		if !chains.ChainIDInChainList(chainID, chainListForHeaderSupport) {
-			return cosmoserrors.Wrapf(
-				sdkerrors.ErrInvalidRequest,
-				"invalid chain id header not supported (%d)",
-				chainID,
-			)
-		}
+
+	if len(msg.ChainIdList) > 200 {
```
ws4charlie, line 51: same constant

x/observer/types/message_vote_block_header.go
```diff
@@ -59,10 +58,6 @@ func (msg *MsgVoteBlockHeader) ValidateBasic() error {
 		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidAddress, err.Error())
 	}
 
-	if !chains.IsHeaderSupportedChain(msg.ChainId) {
```
ws4charlie, line 62: The `IsHeaderSupportedChain` is not used atm. Should remove or keep?

ws4charlie (commented):
LGTM. left a few minor comments

pkg/chains/chains.go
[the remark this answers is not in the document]
lumtis: We have an issue for this: https://github.com/zeta-chain/node/issues/2080
I'm gonna check the complexity removing these as this has actually be planned for sometimes

x/lightclient/types/message_disable_verification_flags.go
lumtis: This is an arbitrary value as a static check, as the value is stored as a singleton state. Can add comment for this one

pkg/chains/chain.go
```diff
@@ -63,13 +52,13 @@ func (chain Chain) IsExternalChain() bool {
 // on EVM chain, it is 20Bytes
 // on Bitcoin chain, it is P2WPKH address, []byte(bech32 encoded string)
 func (chain Chain) EncodeAddress(b []byte) (string, error) {
-	if IsEVMChain(chain.ChainId) {
+	if chain.Consensus == Consensus_ethereum {
```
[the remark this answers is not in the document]
lumtis:
```
IsEVMChain(chain.ChainId)
```
walkthrough the list of chains and find with chain ID to compare, which is unnecessary as the information is already in the chain object

pkg/chains/chain.go
```diff
@@ -87,99 +76,68 @@ func (chain Chain) EncodeAddress(b []byte) (string, error) {
...
-	return DecodeAddressFromChainID(chain.ChainId, addr)
-}
-
 // DecodeAddressFromChainID decode the address string to bytes
-func DecodeAddressFromChainID(chainID int64, addr string) ([]byte, error) {
-	if IsEVMChain(chainID) {
+// additionalChains is a list of additional chains to search from
+// in practice, it is used in the protocol to dynamically support new chains without doing an upgrade
+func DecodeAddressFromChainID(chainID int64, addr string, additionalChains []Chain) ([]byte, error) {
+	if IsEVMChain(chainID, additionalChains) {
```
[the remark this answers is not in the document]
lumtis: a7af5d8c6

pkg/crypto/pubkey.go
```diff
@@ -71,7 +71,7 @@ func (pubKey PubKey) String() string {
 
 // GetAddress will return an address for the given chain
 func (pubKey PubKey) GetAddress(chain chains.Chain) (chains.Address, error) {
-	if chains.IsEVMChain(chain.ChainId) {
+	if chain.Consensus == chains.Consensus_ethereum {
```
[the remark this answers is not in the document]
lumtis: a7af5d8c6

x/lightclient/types/message_disable_verification_flags.go
lumtis: a7af5d8c6

x/authority/keeper/chain_info.go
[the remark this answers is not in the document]
lumtis: `AdditionalChainList` overwrite chains in `DefaultChainList`
So let say a chain is changing its consensus mechanism, this information could be updated without an upgrade

x/observer/types/chain_params.go
```diff
@@ -90,38 +94,36 @@ func ValidateChainParams(params *ChainParams) error {
...
-	if chains.IsBitcoinChain(params.ChainId) {
-		if params.WatchUtxoTicker == 0 || params.WatchUtxoTicker > 300 {
-			return errorsmod.Wrapf(
-				sdkerrors.ErrInvalidRequest,
-				"WatchUtxoTicker %d out of range",
-				params.WatchUtxoTicker,
-			)
-		}
+	// if WatchUtxoTicker defined, check validity
+	if params.WatchUtxoTicker > 300 {
```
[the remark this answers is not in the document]
lumtis: I think we need to reassess if some of the value are still relevant.
Create a separate issue for this: https://github.com/zeta-chain/node/issues/2418

x/crosschain/types/message_add_inbound_tracker.go
lumtis: Replacing `ErrProofVerificationFail` with `sdkerrors.ErrInvalidRequest` there might have been a merge issue

x/observer/types/message_vote_block_header.go
lumtis: Removed it with SupportMerkleProof, as it is unclear once we need back

ws4charlie (approved):
lgtm

x/authority/keeper/chain_info.go
```diff
@@ -27,11 +27,13 @@ func (k Keeper) GetChainInfo(ctx sdk.Context) (val types.ChainInfo, found bool)
...
 
 // GetAdditionalChainList returns the list of chains in chain info object
+// additional chains are additional static chain information stored on-chain used in addition with the default chain information
+// this list allow to add new chain support without doing an upgrade
 // returns empty list if no chains are present
 func (k Keeper) GetAdditionalChainList(ctx sdk.Context) []chains.Chain {
 	chainInfo, found := k.GetChainInfo(ctx)
 	if !found {
-		return
+		return []chains.Chain{}
```
swift1337, line 36: `return nil`

swift1337 (approved):

PR #2393, opened 2024-06-27
feat: add zetae2e serve-addresses

gartnera:
# Description

In order for the toolkit to be usable on localnet and developnet, we need to be able to export the contract/tss/zrc20 addresses in the [format that is expected](https://github.com/zeta-chain/protocol-contracts/blob/main/data/addresses.mainnet.json). These addresses are not stable between localnet deployments, so we cannot simply export the data once. Add `zetae2e serve-addresses` command which will serve this data in the correct format.

This PR is currently an POC for feedback. I will need help correctly mapping all the contract addresses from the e2e config format to the format expected by the toolkit. I will coordinate this change with changes across the networks repo and toolkit repo.

I think it would also be useful to add another endpoint which would serve the raw(ish) config to allow you to easily run `zetae2e` commands locally. So you could do `zetae2e run --config http://localhost:9091/v1/zetae2econfig`.

Relates to:
- https://github.com/zeta-chain/protocol-contracts/pull/188
- https://github.com/zeta-chain/networks/pull/49

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

# Checklist:

- [ ] TSS Addresses
- [ ] ZRC20 Addresses
- [ ] uniswap Addresses (although I'm not sure we deploy the uniswap contracts on localnet eth)

### DIFF
--- a/cmd/zetae2e/root.go
+++ b/cmd/zetae2e/root.go
@@ -29,6 +29,7 @@ func NewRootCmd() *cobra.Command {
 		NewStressTestCmd(),
 		NewInitCmd(),
 		NewSetupBitcoinCmd(),
+		NewServeAddressesCmd(),
 	)
 
 	return cmd
--- a/cmd/zetae2e/serve_addresses.go
+++ b/cmd/zetae2e/serve_addresses.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/spf13/cobra"
+	"github.com/zeta-chain/zetacore/e2e/config"
+)
+
+type v1AddressItem struct {
+	Address   string `json:"address"`
+	Category  string `json:"category"`
+	ChainID   int    `json:"chain_id"`
+	ChainName string `json:"chain_name"`
+	Type      string `json:"type"`
+}
+
+func NewServeAddressesCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "serve-addresses",
+		Short: "Serve addresses of deployed contracts from config file",
+		RunE:  runServeAddresses,
+	}
+	cmd.Flags().StringVarP(&configFile, flagConfig, "c", "", "path to the configuration file")
+	if err := cmd.MarkFlagRequired(flagConfig); err != nil {
+		fmt.Println("Error marking flag as required")
+		os.Exit(1)
+	}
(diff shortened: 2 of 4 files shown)


fadeev:
@gartnera [addresses.testnet.json](https://github.com/zeta-chain/protocol-contracts/blob/9d0791ed765e8a8315a2454cc53fa59601a9af4e/data/addresses.testnet.json) is generated automatically by [a script](https://github.com/zeta-chain/protocol-contracts/blob/9d0791ed765e8a8315a2454cc53fa59601a9af4e/tasks/addresses.ts) that accepts ZetaChain RPC endpoint (mainnet or testnet) and [networks](https://github.com/zeta-chain/networks/blob/a15b2b334949d9138b6138b7d56d7238f61aa2b6/data/networks.json) for connected chains' RPC. I see no reason why the same script cannot be used to fetch contract addresses for any other environment.

fadeev:
We might also consider storing all contract addresses in the upcoming Gateway contract, which will make this even more streamlined. Right now we're making at least 9 requests per chain to get all addresses.

gartnera:
> I see no reason why the same script cannot be used to fetch contract addresses for any other environment.

@fadeev because the addresses are not stable on localnet and developnet. The addresses will be different for every copy of localnet. As it stands today, you'd need to do a new release of protocol-contracts and toolkit to update the addresses right?

fadeev:
> The addresses will be different for every copy of localnet. As it stands today, you'd need to do a new release of protocol-contracts and toolkit to update the addresses right?

@gartnera we can deploy a nightly (tagged) version of protocol contracts on npm with the latest version of developnet addresses. And you can run this script manually to generate localnet addresses. And modify Toolkit to accept those.

lumtis:
> > I think the new [ChainInfo](https://develop-cosmos-api-0.zetachain.network/#/Query/Query_ChainInfo) might be the best spot, so long as we only need it via RPC and not EVM.
> 
> @gartnera from an off-chain program point of view, sure, but if you have **a contract** on Ethereum that needs to know ZETA token address or something, they can't make a query to an API, so having these addresses on EVM is important.

Overall, having a registry contract that the TSS can write into would make the most sense to me for this.

kingpinXD:
> t the TSS can write into would make the most sense to me for this.

I think this is a good idea, but it would mean an additional tx/message on the Zeta core, which initiates this write . We should plan it accordingly .

lumtis:
@gartnera follow up regarding the conoversation with @fadeev 

Is this PR still to be merged?

PR #2394, opened 2024-06-27
chore: update PR template

gartnera:
# Description

- remove "type of change". This is not needed because of conventional naming.
- put instructions behind comments so they don't show up on the actual PR
- add note about linking additional github actions runs.


lumtis (approved):
We can also remove `Checklist` as redundant

kingpinXD (approved):

julianrubino (approved):

PR #2400, opened 2024-06-29
ci: expand performance testing to catch errors, fixed connection refused errors, added performance testing with state.

gzukel:
# Description

1. Expanded the logic the pipeline to look for errors in the report json if no failures we're reported via artillery as a secondary check to ensure we know when tests fail/succeed.
2. Fixed connection refused errors. These errors we're happening locally. The issue is using the composite for artillery uses docker backed composite so the network was separate. The reports weren't clear that it was failing or if it was reporting wrong. That is corrected now.
3. Added starting a local net from the state export to performance test with state.
4. Added slack channel notification when testing fails so the DevOps team can know about it when it happens. 

# How Has This Been Tested?

I executed multiple Github Action scenarios to ensure all logic functioned as intended through failure and success scenarios.

Failed Test: https://github.com/zeta-chain/node/actions/runs/9719615016
Failed Test w/State: https://github.com/zeta-chain/node/actions/runs/9719839293
Success Test: https://github.com/zeta-chain/node/actions/runs/9719810694
Success Tests With State: https://github.com/zeta-chain/node/actions/runs/9719842212/job/26830323773

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


lumtis (approved):
LGTM

.github/actions/performance-tests/art.yaml
```diff
@@ -26,8 +28,17 @@ scenarios:
           json:
             id: 1
             jsonrpc: "2.0"
-            method: "web3_clientVersion"
+            method: "web3_clientVersion"  # Deliberate typo to force failure
```
CharlieMc0, line 31: Is `# Deliberate typo to force failure` let over from testing?

CharlieMc0 (commented):
I think there is one comment leftover from your testing but other than that it looks good.

.github/workflows/ci-nightly-performance-testing.yaml
```diff
@@ -12,23 +12,57 @@ jobs:
...
-        if: success()
+        if: always()
         with:
           name: artillery-report
-          path: ./artillery_report.html
\ No newline at end of file
+          path: ./artillery_report.html
+
+      - name: "NOTIFY:SLACK:FAILURE"
+        if: failure()
```
gartnera, line 62: Please only send the alert if it's a ` schedule` type

gartnera (changes requested):

gartnera (approved):

Issue #2409, opened 2024-07-01
Update CODEOWNER


lumtis:
**Describe the Issue**
Add @fbac in codeowner

PR #2410, opened 2024-07-01
chore: add fbac as codeowner

fbac:
# Description

Add fbac as CODEOWNER.
Fixes https://github.com/zeta-chain/node/issues/2409


gartnera (approved):
@lumtis why do you not just use a github team? `zeta-chain/Protocol-Engineering`

lumtis (approved):

PR #2412, opened 2024-07-01
ci: add build cache to e2e

gartnera:
# Description

Add build caching to the e2e build:
- [cache container layers in the registry](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache)
- [inject go build cache](https://github.com/reproducible-containers/buildkit-cache-dance)

Only populate the cache on the push event. On a cache hit, this should reduce runtimes by about ~2 minutes ([example](https://github.com/gartnera/node/actions/runs/9737295434/job/26870606140))

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,38 +106,57 @@ jobs:
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@v4
+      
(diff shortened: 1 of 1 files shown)


lumtis (approved):
LGTM

CryptoFewka (approved):

PR #2387, opened 2024-06-26
feat(zetaclient): add support for EIP-1559 outbound transactions

swift1337:
# Description

`TBD`

Closes: https://github.com/zeta-chain/node/issues/1952

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Include instructions and any relevant details so others can reproduce. 

- [ ] Tested CCTX in localnet
- [ ] Tested in development environment
- [ ] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions 

# Checklist:

- [ ] I have added unit tests that prove my fix feature works

Issue #2373, opened 2024-06-24
Adding a new Make command running testing upgrade with admin tests


lumtis:
**Describe the Issue**
We currently have:
```
// Run regular E2E tests, upgrade, rerun regular E2E tests
start-upgrade-test

// Upgrade, run regular E2E tests
start-upgrade-test-light
```

We should add a new command that allows to test the admin API after upgrade.
Running 
```
start-e2e-admin-test
```
after completing an upgrade.

**Considered solution**
Adding a Makefile entry running
```
start-upgrade-test-admin
```
The test like the `light` one, proceed with the upgrade then run the admin tests
```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/2446
- [ ] https://github.com/zeta-chain/node/issues/2447
- [ ] https://github.com/zeta-chain/node/issues/2448
```

PR #2415, opened 2024-07-01
refactor: more docker compose deduplicating

gartnera:
# Description

Continue to reduce the number of docker compose files by using [profiles](https://docs.docker.com/compose/profiles/). This is cool because it you can request multiple profiles at the same time to test more complex scenarios. Imagine running stress + upgrade or stress+admin+eth2 etc.

Getting all the services in one compose file is good because you can easily see IP address conflicts. It's also more obvious what needs to be updated when you're refactoring things. The stress tests were actually broken because I forgot to update the `docker-compose-stress.yml`.

Also detect presence of other containers (eth2, zetacored replicas) via dns resolution rather that static config. Example output from zetaclient now:

```
Waiting for zetacore0 rpc
operatorAddress: zeta12x9lynrkr7y9q58kdtdapn3pvawcrnlldvzjpl
Start zetaclientd
eth2 has address 172.20.0.102
enabling additional evm (eth2)
```

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


contrib/localnet/docker-compose.yml
```diff
@@ -120,6 +187,20 @@ services:
         ipv4_address: 172.20.0.100
     entrypoint: ["geth", "--dev", "--http", "--http.addr", "172.20.0.100", "--http.vhosts", "*", "--http.api", "eth,web3,net", "--http.corsdomain", "https://remix.ethereum.org", "--dev.period", "2"]
 
+  eth2:
+    build:
+      context: ./anvil
```
swift1337, line 192: AFAIK `anvil` doesn't support pre EIP 1559 transactions. So this is blocked by https://github.com/zeta-chain/node/pull/2387

swift1337: Also, should we remove `rosetta` image?

contrib/localnet/docker-compose.yml
```diff
@@ -157,6 +238,38 @@ services:
...
+    depends_on:
+      - zetacore0
+      - upgrade-host
+    environment:
+      - UPGRADE_HEIGHT=${UPGRADE_HEIGHT}
+    volumes:
+      - ssh:/root/.ssh
 volumes:
   ssh:
   preparams:
```
swift1337, line 275: nit: new line

swift1337 (commented):
Looks good, left some comments

contrib/localnet/docker-compose.yml
gartnera: > AFAIK anvil doesn't support pre EIP 1559 transactions. So this is blocked by https://github.com/zeta-chain/node/pull/2387

This is just refactoring, so if it was broken before this it will continue to be broken. Not a blocker if it was broken before imo.

> Also, should we remove rosetta image?

I don't specifically know of anything that uses it

lumtis: Yes it doesn't work but it's currently not used, the test for chain migration is disabled

lumtis (approved):
I might see it could be used for https://github.com/zeta-chain/node/issues/2373?
cc @kingpinXD

CryptoFewka (approved):
looks good to me

Issue #2404, opened 2024-07-01
Add Solana static chain info


lumtis:
**Describe the Issue**
Add Solana static chain info in `chains` package.

As we're using EVM TSS signer, we need assess if new field needs to be added regarding consensus, etc..

Do it for all chains necessary:
- localnet
- testnet
- mainnet

PR #2416, opened 2024-07-01
feat: add Solana chain information

lumtis:
# Description

Close: https://github.com/zeta-chain/node/issues/2404


proto/zetachain/zetacore/pkg/chains/chains.proto
```diff
@@ -86,6 +92,7 @@ enum Consensus {
   tendermint = 1;
   bitcoin = 2;
   op_stack = 3;
+  solana_consensus = 4;
```
lumtis, line 95: This is a bit ugly and inconsistent with the rest. Thing is you can't have a enum value with the same name as one value of another enum in the same file (here solana in network enum) so we might need to do some refactoring to change all values

lumtis: > which also seems weird so probably better to just change all values

IMO this is the best.
We can also change all value without breaking compatibility in the future

fbac: What I've done in the past and is very intuitive on this case is to just prefix the values with the name of the enum, as in Consensus_solana, Network_solana, etc.

lumtis: Yeah but then in appear as `Consensus_consensus_solana` in the generated code.

Which is not a big problem per se

kingpinXD (approved):

proto/zetachain/zetacore/pkg/chains/chains.proto
lumtis: Created an issue for this one: https://github.com/zeta-chain/node/issues/2420

pkg/chains/chains.go
```diff
@@ -91,6 +91,18 @@ var (
 		CctxGateway: CCTXGateway_observers,
 	}
 
+	// SolanaMainnet is Solana mainnet
+	SolanaMainnet = Chain{
+		ChainName:   ChainName_solana_mainnet,
+		ChainId:     900,
```
ws4charlie, line 97: From which official source did we see this mainnet chain ID? An unofficial place I saw is: https://docs.expand.network/important-ids/chain-id

lumtis: Yeah, it's unofficial. Since the chain ID is used in the signature unlike Bitcoin, we might want to check carefully

pkg/chains/chains.go
```diff
@@ -179,6 +191,18 @@ var (
 		CctxGateway: CCTXGateway_observers,
 	}
 
+	// SolanaTestnet is Solana testnet
```
ws4charlie, line 194: Seems the `Devnet` is a commonly used name for Solana devs (user, app developer) and `Testnet` is for Solana core contributors stress test recent release features. See this: https://docs.solanalabs.com/clusters/available

lumtis: Good point, I will rename to avoid confusion

pkg/chains/chains.go
lumtis: Created issue to define final values here: https://github.com/zeta-chain/node/issues/2421

pkg/chains/chains.go
lumtis: a843cc4cc

pkg/chains/chains.go
lumtis: a843cc4cc

PR #2428, opened 2024-07-03
feat(zetaclient): propagate context across codebase & refactor zetacore client

swift1337:
# Description
This PR revamps the `zetaclient` codebase mainly by wiring `ctx context.Context` across the codebase.

- [x] Add `retrier` package with exponential backoff out-of-the-box
- [x] Add `bg` package for running go routines safely
- [x] Drastically refactor `zetacore` client package, separate each QueryClient into a different file
- [x] Propagate `ctx` across the codebase
- [x] Add several panic handlers for goroutines
- [x] Improve logging with log fields
- [x] Improve testing suite, switch to `mockery` for `ZetaClient` interface

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced a robust retry mechanism with exponential backoff for handling retryable errors.
  - Added new background task management utilities for improved task handling and error management.

- **Refactor**
  - Enhanced context handling across various functions to improve control over execution flow and pass context-specific information seamlessly.

- **Tests**
  - Added comprehensive test cases for new background task management utilities, including logging and panic recovery scenarios.


zetaclient/chains/evm/observer/inbound.go
```diff
@@ -192,13 +203,13 @@ func (ob *Observer) ObserveInbound(sampledLogger zerolog.Logger) error {
...
-	lastScannedZetaSent := ob.ObserveZetaSent(startBlock, toBlock)
+	lastScannedZetaSent, _ := ob.ObserveZetaSent(ctx, startBlock, toBlock)
 
 	// task 2: query evm chain for deposited logs (read at most 100 blocks in one go)
-	lastScannedDeposited := ob.ObserveERC20Deposited(startBlock, toBlock)
+	lastScannedDeposited := ob.ObserveERC20Deposited(ctx, startBlock, toBlock)
 
 	// task 3: query the incoming tx to TSS address (read at most 100 blocks in one go)
-	lastScannedTssRecvd := ob.ObserverTSSReceive(startBlock, toBlock)
+	lastScannedTssRecvd, _ := ob.ObserverTSSReceive(ctx, startBlock, toBlock)
```
lumtis, line 212: I think we should log the error here?

lumtis (commented):
Just add a first look, will look more in the created files, etc.. later

lumtis:
@coderabbitai can you please give a summary of the changes

pkg/bg/bg.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type Opt func(*config)
+
+func WithName(name string) Opt {
+	return func(cfg *config) { cfg.name = name }
+}
+
+func WithLogger(logger zerolog.Logger) Opt {
+	return func(cfg *config) { cfg.logger = &logger }
```
ws4charlie, line 24: not sure if we really need a pointer `*zerolog.Logger` type if it's already passed by value `logger zerolog.Logger`.

zetaclient/chains/evm/observer/inbound.go
ws4charlie: Ideally we need to handle the error if AppContext can't be retrieved (I know it will rarely/never happen).

zetaclient/chains/evm/observer/observer.go
```diff
@@ -486,13 +475,13 @@ func (ob *Observer) LoadLastBlockScanned() error {
...
 // postBlockHeader posts the block header to zetacore
 // TODO(revamp): move to a block header file
-func (ob *Observer) postBlockHeader(tip uint64) error {
+func (ob *Observer) postBlockHeader(ctx context.Context, tip uint64) error {
 	bn := tip
 
-	res, err := ob.ZetacoreClient().GetBlockHeaderChainState(ob.Chain().ChainId)
-	if err == nil && res.ChainState != nil && res.ChainState.EarliestHeight > 0 {
+	chainState, err := ob.ZetacoreClient().GetBlockHeaderChainState(ctx, ob.Chain().ChainId)
+	if err == nil && chainState != nil && chainState.EarliestHeight > 0 {
```
ws4charlie, line 482: In the bitcoin `observer.go` doesn't check `chainState != nil`. For consistency purpose, can we simply use same criterion for both?

zetaclient/chains/evm/observer/outbound.go
```diff
@@ -90,49 +99,74 @@ func (ob *Observer) WatchOutbound() {
...
-	} else if zetaTxHash != "" {
-		logger.Info().Msgf("PostVoteOutbound: posted vote for chain %d nonce %d outbound %s vote %s ballot %s", chainID, nonce, receipt.TxHash, zetaTxHash, ballot)
+		logger.Error().Err(err).Fields(lf).Msgf("PostVoteOutbound: error posting vote for chain %d", chainID)
+		return
 	}
+
+	lf["outbound.zeta_tx_hash"] = zetaTxHash
+	lf["outbound.ballot"] = ballot
+
+	logger.Info().Fields(lf).Msgf("PostVoteOutbound: posted vote for chain %d", chainID)
```
ws4charlie, line 159: Can we keep the previous logic which prints logs only when it qualifies `else if zetaTxHash != ""`? The reason is that if we don't check empty `zetaTxHash`, we'll be printing lots of logs.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -100,20 +103,20 @@ func NewOrchestrator(
...
 
 	// start cctx scheduler
-	go oc.StartCctxScheduler(appContext)
+	bg.Work(ctx, oc.StartCctxScheduler, bg.WithName("StartCctxScheduler"), bg.WithLogger(oc.logger.Std))
 
 	// watch for upgrade plan from zetacore
 	go func() {
 		// wait for upgrade plan signal to arrive
-		oc.zetacoreClient.Pause()
+		oc.zetacoreClient.Stop()
```
ws4charlie, line 119: Is this a real stop? The previous `Pause()` is waiting for upgrade plan to arrive its height?

ws4charlie (commented):
looks nice. A few minor comments.

zetaclient/chains/evm/observer/inbound.go
swift1337: @ws4charlie we indeed handle with in most cases in this PR.

pkg/bg/bg.go
swift1337: I used pointer to determine whether logger is set for this bg job. afaik there's no `logger.IsNop()`

ws4charlie: Does it panic when using a `Nop` logger to print logs?

swift1337: https://github.com/zeta-chain/node/pull/2428/commits/55626c964cf8850e36026aeca127059aa9821aa5

zetaclient/orchestrator/orchestrator.go
swift1337: This is a good question. Can you elaborate on initial `Pause()` use-case? Because the log says
```
Msgf("Active upgrade plan detected and upgrade height reached: %s at height %d; ZetaClient is stopped;"+
				"please kill this process, replace zetaclientd binary with upgraded version, and restart zetaclientd", plan.Name, plan.Height)
```


https://github.com/zeta-chain/node/blob/develop/zetaclient/zetacore/client.go#L206

swift1337: UPD: I followed the logic in `develop` branch. I found `Pause()` method confusing as it doesn't "pause" the client but blocks the current go-routine until someone pushes `c.pause <- struct{}{}`. 

I'll refactor code accordingly, thanks for spotting this!

ws4charlie (approved):

pkg/retry/retry.go
gartnera, line 1: The problem with this api is that there isn't a good way to handle context cancellations/timeouts. If the context is cancelled during the backoff sleep you will have to wait until that sleep concludes.

I think most of these functions need to take a context parameter?

pkg/bg/bg.go
gartnera, line 1: Why is this needed when https://pkg.go.dev/golang.org/x/sync/errgroup exists?

Background tasks need a way to report their status to the foreground task?

Regardless, needs unit tests.

gartnera (changes requested):

zetaclient/chains/bitcoin/observer/outbound.go
```diff
@@ -176,26 +188,52 @@ func (ob *Observer) IsOutboundProcessed(cctx *crosschaintypes.CrossChainTx, logg
...
 		chains.ReceiveStatus_success,
-		ob.Chain(),
+		ob.Chain().ChainId,
 		nonce,
 		coin.CoinType_Gas,
 	)
+
+	zetaHash, ballot, err := ob.ZetacoreClient().PostVoteOutbound(ctx, gasLimit, gasRetryLimit, msg)
+
+	lf := map[string]any{
```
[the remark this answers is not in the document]
swift1337: yes, I can rename to `logFields`

pkg/bg/bg.go
swift1337: `errgroup` is needed for tasks that end eventually e.g. "fetch 10 urls but no more than 5 at the same time" (~ waitGroup + semaphore). `bg` is more for background go-routines. Currently, this package has bare minimum feature-set, but I plan to extend it in the following PRs. Also `bg` has logging and **handles panics** in comparison to `errgroup`.

Another feature that I want to put in `bg` is proper "background service" wrapper what support generic `Start()`, `Stop()`, graceful shutdown, signaling, ticker-based background routines, etc...

swift1337: Test cases: https://github.com/zeta-chain/node/pull/2428/commits/d7ccb9c1ea49634133a5087d982bc950a613e5c3

pkg/retry/retry.go
swift1337: > You will have to wait until that sleep concludes.

It's up to the caller as `DoWithBackoff(call, backoff)` doesn't perform a retry if an error isn't explicitly wrapped into `Retry(err)`. However, your point is valid for other high-level aliases such as `DoTypedWithRetry`.

I'll handle `ctx` errors as a popular edge case, but keep the function signatures as is. Not all methods that might be retried require context.

swift1337: https://github.com/zeta-chain/node/pull/2428/commits/7e96f9d7bfd3b924ebc260fea2982593bde32763

pkg/bg/bg.go
lumtis: but I plan to extend it in the following PRs

@swift1337 let's create issues for these

lumtis (approved):

pkg/bg/bg.go
swift1337: https://github.com/zeta-chain/node/issues/2469

pkg/retry/retry.go
```diff
@@ -0,0 +1,143 @@
...
+		if !isRetryable {
+			return err
+		}
+
+		sleepFor := bo.NextBackOff()
+		if sleepFor == backoff.Stop {
+			return errors.Wrap(err, "retry limit exceeded")
+		}
+
+		time.Sleep(sleepFor)
```
gartnera, line 75: I'm saying this bit here specifically ignores context cancellations because this api is not context aware

PR #2434, opened 2024-07-03
fix: set pebbledb as default database

### LINKED ISSUE #2432: Update zetacored init config with pebbledb
When `zetacored init` is run, it creates a configuration which specifies leveldb. Pebbledb has higher performance and is what our snapshots use. 

We should attempt to get pebbledb set as the backend within the configuration files that are written to disk when running `zetacored init`.

gartnera:
# Description

Set pebbledb as the default database because that's what format our snapshots are stored in.

Closes #2432

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/cmd/zetacored/root.go
+++ b/cmd/zetacored/root.go
@@ -114,6 +114,8 @@ func initTmConfig() *tmcfg.Config {
 		cfg.Mempool.Version = tmcfg.MempoolV1
 	}
 
+	cfg.DBBackend = "pebbledb"
+
 	return cfg
 }
 


cmd/zetacored/root.go
```diff
@@ -114,6 +114,8 @@ func initTmConfig() *tmcfg.Config {
 		cfg.Mempool.Version = tmcfg.MempoolV1
 	}
 
+	cfg.DBBackend = "pebbledb"
```
kingpinXD, line 117: I think we moved to rocksDB at some point to improve performace ? 
And compared pebbleDB and rocksDB when doing that .  Has something changed since then ?

gartnera: All our nodes run on pebbledb, so if you want to restore from our snapshots (the only thing that's really supported) then you also have to use pebbledb.

kingpinXD (approved):

PR #2460, opened 2024-07-10
feat!: upgrade to go 1.22

gartnera:
# Description

Upgrade to go 1.22. Pretty simple on the node side, but please see https://github.com/zeta-chain/go-tss/pull/19.

This is a breaking change because QUIC P2P will no longer function. We don't think anyone is using it but technically other zetaclientd operators could have configured it.

TODO:

- [x] [cosmos gosec](https://github.com/cosmos/gosec) is unmaintained, I will rebase their ruleset on the upstream repo. Use [fork](https://github.com/zeta-chain/gosec) until upstream changes are merged and released.
    - [x] [G701](https://github.com/cosmos/gosec/blob/master/rules/rulelist.go#L115) is now upstream as [G115](https://github.com/securego/gosec/pull/1149/files#diff-d197038414d78239949c635273d8bb244c9ac660310e702438cfb8e3c9fe4c10). Update: need fix for [upstream issue](https://github.com/securego/gosec/pull/1167).
    - [x] [G703](https://github.com/cosmos/gosec/blob/bf28a33fadf2652753980200865d63512cde870e/rules/rulelist.go#L117) is supposed to be an improved version of [G104](https://github.com/securego/gosec/blob/master/rules/rulelist.go#L70) but there is just a [TODO](https://github.com/cosmos/gosec/blob/bf28a33fadf2652753980200865d63512cde870e/rules/sdk/errors.go#L80) to actually implement the functionality. I think it's ok to remove this check.
    - [x] [G702](https://github.com/cosmos/gosec/blob/bf28a33fadf2652753980200865d63512cde870e/rules/rulelist.go#L116C5-L116C9) is a dumber version of G103 and G404. I think it's ok to remove. It's also not currently enabled anyway.
    - [x] [G704](https://github.com/cosmos/gosec/blob/bf28a33fadf2652753980200865d63512cde870e/rules/rulelist.go#L118C5-L118C9) may be useful and I may try to upstream it. But the [described scenario](https://github.com/cosmos/gosec/tree/master/rules/sdk#strconv-unsigned-integers-cast-to-signed-integers-overflow) seems to be pretty dumb when `strconv.ParseInt()` exists.
- [x] Evaluate new gosec rule failures.
    - G115 false positives I had to suppress
    - Some `MkdirAll` 0o777 -> 0o750

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **Chores**
  - Updated the Go version from 1.20 to 1.22 across multiple configurations including GitHub Actions workflows and Dockerfiles.
  - Changed security annotations in comments from `G701` to `G115` to reflect updated security guidelines.

- **New Features**
  - Introduced a new `gosec.sh` script for running security scans using the `gosec` tool.

- **Documentation**
  - Updated comments in various files to align with the new security annotation standards.


.github/actions/install-dependencies/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
     - uses: actions/setup-go@v5
       if: ${{ inputs.skip_go == 'false' }}
       with:
-        go-version: '1.20'
+        go-version: '1.22'
```
fbac, line 36: Stefan and me were discussing this way of hardcoding versions in a https://github.com/zeta-chain/protocol-contracts/pull/212#discussion_r1674318232

To make this more dynamic we could have a .versions file in every one of our repositories, centralizing the versioning in just one place.

swift1337 (approved):
Awesome!

morde08 (approved):
LGTM

.github/actions/install-dependencies/action.yml
gartnera: I think having a central shared `actions/setup-go` that sets the version once would be reasonable.

You could read this out of the go.mod file directly rather than a separate file.

PR #2470, opened 2024-07-12
feat: add Polygon, Base and Base Sepolia in static chain info

lumtis:
# Description

As part of v18

[upgrade tests pass](https://github.com/zeta-chain/node/actions/runs/9910599831)


ws4charlie (approved):

kingpinXD (approved):

docs/openapi/openapi.swagger.yaml
```diff
@@ -53612,6 +53612,10 @@ definitions:
       - goerli_localnet
       - btc_regtest
       - amoy_testnet
+      - optimism_mainnet
```
swift1337, line 53615: Are OP stack keys placed on purpose?

app/setup_handlers.go
```diff
@@ -5,12 +5,9 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
-	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
-	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
-	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
 )
 
-const releaseVersion = "v17"
```
gartnera, line 13: Do we really need to do a full major version for this change? This seems more like a v17.1.0 but requires an upgrade proposal.

docs/openapi/openapi.swagger.yaml
lumtis: To keep consistency with the field numbers on `develop` although these values are not used here

app/setup_handlers.go
lumtis: This is a consensus breaking changes and we decided to do a major release for every consensus breaking changes, unless we discuss changing this recently?

gartnera: Ok

gartnera (approved):

lumtis:
Upgrade test successful completion: https://github.com/zeta-chain/node/actions/runs/9910599831/job/27381349189

pkg/chains/chain.go
```diff
@@ -91,7 +91,10 @@ func IsZetaChain(chainID int64) bool {
 func IsEVMChain(chainID int64) bool {
 	return chainID == 5 || // Goerli
 		chainID == AmoyChain().ChainId ||
-		chainID == SepoliaChain().ChainId || // Sepolia
+		chainID == PolygonChain().ChainId ||
+		chainID == SepoliaChain().ChainId ||
+		chainID == BaseMainnetChain().ChainId ||
```
brewmaster012, line 96: should we add optimism chains here

brewmaster012 (approved):
LGTM

CharlieMc0 (approved):

PR #2472, opened 2024-07-12
feat: add an IBC chain in the localnet

lumtis:
# Description

Add an IBC chain and a relayer container in the localnet to allow testing IBC related features

### DIFF
--- a/contrib/localnet/ibc-relayer/Dockerfile
+++ b/contrib/localnet/ibc-relayer/Dockerfile
@@ -1,0 +1,29 @@
+# Stage 1: Set permissions using a base image that allows it
+FROM alpine:latest as builder
+
+# Install necessary tools
+RUN apk add --no-cache bash
+
+# Set the working directory
+WORKDIR /root/.hermes
+
+# Copy entrypoint script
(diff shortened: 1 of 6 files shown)


contrib/localnet/ibc-relayer/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN apk add --no-cache bash
+
+# Set the working directory
+WORKDIR /root/.hermes
+
+# Copy entrypoint script
+COPY entrypoint.sh .
+
+# Set permissions
+RUN chmod +x entrypoint.sh
```
gartnera, line 14: This seems entirely redundant? You should just be able to copy it directly in?

lumtis: Thanks for the feedback. The work on this is on hold for now to prioritize Solana and new contract design.
I think for Cosmos we will end up in a potential similar setup as it is currently considered for Solana, with the containers not included in the default localnet but that can be optionally bootstrapped

lumtis:
Closing as the integration will be deprioritized for now

Issue #1952, opened 2024-03-27
Support `EIP-1559` transactions with ZetaClient


lumtis:
**Describe the Issue**
Send `EIP-1559` transactions using fee market mechanism on ZetaClient and prevent using legacy transactions

Context: https://github.com/zeta-chain/node/pull/1900#discussion_r1539131000

lumtis:
@ws4charlie would you see an issue with using EIP-1559 transactions with ZetaClient?

ws4charlie:
Here is a comparison of a legacy transaction and an EIP-1559 transaction:
<br />

**Transaction fee:**

-   A legacy transaction fee is calculated as: `gasPrice x gasUsed`

-   An EIP-1559 transaction fee is calculated as: `(baseFee + gasTipCap) x gasUsed`, see: [metamask doc](https://support.metamask.io/transactions-and-gas/gas-fees/how-to-estimate-the-gas-fee/#:~:text=As%20of%20EIP%2D1559%2C%20the,x%20units%20of%20gas%20used))

-   From the perspective of CCTX fee charging: `gasPrice  == baseFee + gasTipCap`


    In order to not touch the CCTX fee charging logic, zetaclient can simply post (`baseFee + maxPriorityFeePerGas`) as a replacement of `gasPrice` for EIP-1559 enabled chains.
<br />
<br />

**Transaction building:**

- A legacy transaction requires: `gasPrice`, `gasLimit`
![image](https://github.com/zeta-chain/node/assets/34498985/6b20f291-a4ef-4beb-aab7-3dc249cbce1c)
- An EIP-1559 transaction requires: `gasTipCap`, `maxFeePerGas`, `gasLimit`
![image](https://github.com/zeta-chain/node/assets/34498985/384f83a0-692a-4c54-aa5b-f6f2e5863e18)
- Sending a EIP-1559 transaction without explicit `gasTipCap` or `maxFeePerGas` will result in error (in live tests).

  Where to get deterministic numbers of `gasTipCap` and `maxFeePerGas` when signing a TSS transaction?

- Remember the equation `gasPrice == baseFee + gasTipCap`. If `gasTipCap` becomes a field of CCTX outbound parameter, the zetaclient can then calculate `baseFee` as
`gasPrice - gasTipCap`.

- The `maxFeePerGas` is the upper limit of gas price that the sender is willing to pay. The estimation of `maxFeePerGas` can be done by
`maxFeePerGas = gasTipCap + baseFee * 2`

[...]

PR #2483, opened 2024-07-15
feat(crosschain): store priorityFee (GasTipCap) for EVM chains

swift1337:
# Description

This PR adds priorityFee from EIP-1559 to ZetaChain state. Updates:
- Added `uint64 priorityFee` to `MsgVoteGasPrice` so observers can propose priority fees from `rpc.SuggestGasTipCap()`
- Added `[]uint64 priority_fee` to `crosschain.types.GasPrice{}`. The logic is the same: we pick **median** price
- Added `string cctx.outbound_params.gas_priority_fee` (same as `string cctx.outbound_params.gas_priority_fee`)
- Minor refactoring in `x/crosschain`

Note that this PR unblocks https://github.com/zeta-chain/node/pull/2387 which actually signs and broadcasts transactions with priority fees

Part of https://github.com/zeta-chain/node/issues/1952

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced `priorityFee` for EIP-1559 transactions to enhance gas fee handling.
  
- **Enhancements**
  - Updated various CLI commands and message structs to replace the `supply` parameter with `priorityFee`.
  - Improved error handling and message formatting in several functions.
  
- **Bug Fixes**
  - Fixed field assignments and updated test cases for gas price management.

- **Documentation**
  - Updated documentation to reflect changes related to `priorityFee` and gas price.

- **Refactor**
  - Refactored multiple functions to include and handle `priorityFee` for gas calculations and transactions.


proto/zetachain/zetacore/crosschain/gas_price.proto
```diff
@@ -9,6 +9,12 @@ message GasPrice {
...
   repeated string signers = 4;
   repeated uint64 block_nums = 5;
+
   repeated uint64 prices = 6;
+
+  // index of the median gas price in the prices array
   uint64 median_index = 7;
+
+  // priority fees for EIP-1559
+  repeated uint64 priority_fees = 8;
```
swift1337, line 19: Note that we don't need to store `median_index_priority_fees` here as we can calculate it on demand.

swift1337:
@lumtis @ws4charlie In some cases in `x/crosschain`, we alter `cctx.outbound_params` and bump gasPrice by either some percent or by a factor of `x2`. I kept priorityFee the same without changes. Should I apply those factors to priorityFee as well?

x/crosschain/keeper/abci_test.go
```diff
@@ -360,7 +360,7 @@ func TestCheckAndUpdateCctxGasPrice(t *testing.T) {
 				})
 
 				// ensure median gas price is set
-				medianGasPrice, isFound := k.GetMedianGasPriceInUint(ctx, chainID)
+				medianGasPrice, _, isFound := k.GetMedianGasPriceInUint(ctx, chainID)
```
lumtis, line 363: Can we test with some values here?

x/crosschain/keeper/gas_price.go
```diff
@@ -6,36 +6,72 @@ import (
...
 	return val, true
 }
 
-func (k Keeper) GetMedianGasPriceInUint(ctx sdk.Context, chainID int64) (sdk.Uint, bool) {
-	gasPrice, isFound := k.GetGasPrice(ctx, chainID)
-	if !isFound {
-		return math.ZeroUint(), isFound
+// GetMedianGasPriceInUint returns the median gas price (and median priority fee) from the store
+// or false if it doesn't exist.
+func (k Keeper) GetMedianGasPriceInUint(ctx sdk.Context, chainID int64) (math.Uint, math.Uint, bool) {
```
lumtis, line 39: Maybe it can be renamed as it now returns priority fee as well?

x/crosschain/keeper/gas_price.go
```diff
@@ -6,36 +6,72 @@ import (
...
+		gasPrice    = math.NewUint(entity.Prices[entity.MedianIndex])
+		priorityFee = math.NewUint(medianValue(entity.PriorityFees))
+	)
+
+	return gasPrice, priorityFee, true
+}
+
+// medianValue returns the median value of a slice
+// example: [ 1 7 5 2 3 6 4 ] => [ 1 2 3 4 5 6 7 ] => 4
+func medianValue(items []uint64) uint64 {
```
lumtis, line 55: Can we add some unit test for it? Also we can move it to `pkg`

x/crosschain/keeper/gas_payment.go
```diff
@@ -43,42 +56,43 @@ func (k Keeper) PayGasAndUpdateCctx(
...
 	// get the gas limit
-	gasLimitQueried, err := k.fungibleKeeper.QueryGasLimit(ctx, gasZRC20)
-	if err != nil {
-		return gasZRC20, gasLimit, gasPrice, protocolFee, err
+	gasLimit, err := k.fungibleKeeper.QueryGasLimit(ctx, gasZRC20)
+	switch {
+	case err != nil:
+		return ChainGasParams{}, errors.Wrap(err, "unable to get gas limit")
+	case gasLimit == nil:
+		return ChainGasParams{}, errors.New("gas limit is nil")
```
kingpinXD, line 71: NIT : Can consider defining a type for the new errors

zetaclient/chains/bitcoin/observer/observer.go
```diff
@@ -376,7 +376,7 @@ func (ob *Observer) PostGasPrice(ctx context.Context) error {
 	}
 
 	// #nosec G115 always positive
-	_, err = ob.ZetacoreClient().PostVoteGasPrice(ctx, ob.Chain(), feeRateEstimated, "100", uint64(blockNumber))
+	_, err = ob.ZetacoreClient().PostVoteGasPrice(ctx, ob.Chain(), feeRateEstimated, 0, uint64(blockNumber))
```
kingpinXD, line 379: lets add a comment to clarify why we are setting the 0 value here .

x/crosschain/keeper/msg_server_migrate_tss_funds.go
```diff
@@ -136,6 +136,7 @@ func (k Keeper) MigrateTSSFundsForChain(
 			TssNonce:               0,
 			GasLimit:               1_000_000,
 			GasPrice:               medianGasPrice.MulUint64(2).String(),
+			GasPriorityFee:         priorityFee.String(),
```
kingpinXD, line 139: We can have a multiplier for important transactions like this , so as to increase the chances of their success.

proto/zetachain/zetacore/crosschain/cross_chain_tx.proto
```diff
@@ -65,6 +65,7 @@ message OutboundParams {
   uint64 tss_nonce = 5;
   uint64 gas_limit = 6;
   string gas_price = 7;
+  string gas_priority_fee = 13;
```
ws4charlie, line 68: Has `13` already been used before at some point? Maybe `23` is for the next new field ? @lumtis

lumtis: I don't remember why we jumped on the number like this. I don't think we ever used `13`

x/crosschain/keeper/msg_server_migrate_tss_funds.go
```diff
@@ -181,6 +181,7 @@ func (k Keeper) MigrateTSSFundsForChain(
 				),
 			)
 		}
+		cctx.GetCurrentOutboundParam().GasPrice = multipliedGasPrice.String()
```
ws4charlie, line 184: We need to fill `GasPriorityFee` as well, do we? zetaclient will use it to transfer TSS funds thru a EIP-1559 transaction.

zetaclient/chains/evm/observer/observer.go
```diff
@@ -332,16 +332,24 @@ func (ob *Observer) PostGasPrice(ctx context.Context) error {
 		ob.Logger().GasPrice.Err(err).Msg("Err SuggestGasPrice:")
 		return err
 	}
+
+	priorityFee, err := ob.evmClient.SuggestGasTipCap(ctx)
+	if err != nil {
+		ob.Logger().GasPrice.Err(err).Msg("unable to load priority fee")
+		// we want fallback to zero if we can't get the priority fee
```
ws4charlie, line 339: The `err` can be any one of RPC failures. To tell if a network supports EIP-1559 or not, we need to fetch block header and check if the `baseFee` field exists. You can refer to [Chainsafe's implementation](https://github.com/ChainSafe/chainbridge-core/blob/main/chains/evm/calls/evmgaspricer/london.go#L24) .

The logic would look like

1. Fetch header and check if the baseFee exists.
2. If it exists, then we MUST succeed in the `SuggestGasTipCap` query.
3. If it doesn't exist, the network doesn't support EIP-1559 and we fall back to zero.

x/crosschain/keeper/gas_price.go
```diff
@@ -6,36 +6,72 @@ import (
...
 	}
-	mi := gasPrice.MedianIndex
-	return sdk.NewUint(gasPrice.Prices[mi]), true
+
+	// We don't want to modify the original slice
+	copiedItems := make([]uint64, len(items))
+	copy(copiedItems, items)
+
+	slices.Sort(copiedItems)
+	mv := copiedItems[len(copiedItems)/2]
```
fbac, line 68: Given a data set of [1, 1, 1, 1, 7, 7, 7, 7] this would return 1 as the median value, is this expected? shouldn't we return 4 on that case?

x/crosschain/keeper/msg_server_migrate_tss_funds.go
fbac: Should we cover the case where the chain does not support 1559?

x/crosschain/keeper/msg_server_migrate_tss_funds.go
fbac: I like this idea the multiplier could be dynamically calculated depending on the receiving chain usage.

x/crosschain/keeper/gas_price.go
swift1337: This is how it currently works for gasPrice 😅  Let's indeed use the "real" median there with some rounding to the closest integer. I'll also move this to a separate package.

proto/zetachain/zetacore/crosschain/cross_chain_tx.proto
swift1337: Okay, I'll bump it. 

To prevent this in the future, we can integrate `buf breaking` into CI (https://buf.build/docs/breaking/overview). It works great, however sometimes it makes life harder rather than easier 😅  cc @gartnera

x/crosschain/keeper/gas_price.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/aab53c875f033855ac4ba568f0ec7b59fc2860f4

x/crosschain/keeper/gas_price.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/7c16387dd93d3863b09849169d6de3ead7c35b39

x/crosschain/keeper/abci_test.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/7c16387dd93d3863b09849169d6de3ead7c35b39

proto/zetachain/zetacore/crosschain/cross_chain_tx.proto
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/fa69f4873705dd2418563a58a49253c9c609b1c5

zetaclient/chains/bitcoin/observer/observer.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/4193e83b33a82eb2f701249f7fc1492a945e5987

x/crosschain/keeper/gas_payment.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/4193e83b33a82eb2f701249f7fc1492a945e5987

x/crosschain/keeper/msg_server_migrate_tss_funds.go
swift1337: I'll add a hard-coded multiplier for now, opened an issue for a more complex logic. https://github.com/zeta-chain/node/issues/2528

x/crosschain/keeper/msg_server_migrate_tss_funds.go
swift1337: It's there, check comments above. https://github.com/zeta-chain/node/pull/2483/files#r1679509456

x/crosschain/keeper/msg_server_migrate_tss_funds.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/2ceb5b0b0290da52b3b6c0860afe828b4c58e3e0

zetaclient/chains/evm/observer/observer.go
swift1337: https://github.com/zeta-chain/node/pull/2483/commits/aa6de92d7d8dae05c29c029453f2b64228b34d0f

lumtis:
Missed this comment

> Should I apply those factors to priorityFee as well?

Let's keep it without factor for now

lumtis (commented):
Looks good to me
Will approve once upgrade test passes

ws4charlie:
> @lumtis @ws4charlie In some cases in `x/crosschain`, we alter `cctx.outbound_params` and bump gasPrice by either some percent or by a factor of `x2`. I kept priorityFee the same without changes. Should I apply those factors to priorityFee as well?

I think you're talking about gas stability pool updating the gas price. If it's stability pool, the answer would be `Yes`.
To unblock a EIP-1559 transaction stuck in mempool, we'll have to bump both gas price and gasTipCap. Otherwise, transaction replacement won't be accepted by Ethereum.

Here is the bump percentage in Ethereum source code: https://github.com/ethereum/go-ethereum/blob/87246f3cbaf10f83f56bc4d45f0f3e36e83e71e9/core/txpool/legacypool/legacypool.go#L146

Here is a doc from Alchemy: https://docs.alchemy.com/docs/retrying-an-eip-1559-transaction

The Github issue contains the details in section `Pending Transaction replacement`.

lumtis (approved):

pkg/math/cosmos_uint.go
```diff
@@ -0,0 +1,21 @@
+package math
+
+import "cosmossdk.io/math"
```
fbac, line 3: Is `Uint64()` from cosmossdk implemented as arbitrary-precision? Similar to big ints from math/big.
If it's not we would have to check for overflows as Add() could return a panic when trying to return a number that is out of bounds.

swift1337: it's a wrapper for `*big.Int`

swift1337:
@ws4charlie priorityFee bump: https://github.com/zeta-chain/node/pull/2483/commits/f50abea8734d9d287dec1067eba151a3267fb7e5

zetaclient/chains/evm/observer/observer_gas.go
```diff
@@ -0,0 +1,146 @@
...
+		return false, errors.Wrap(err, "unable to get base fee")
+	}
+
+	ob.Logger().GasPrice.Info().
+		Int64("observer.chain_id", ob.Chain().ChainId).
+		Str("observer.base_fee", baseFee.String()).
+		Msg("Fetched base fee for chain")
+
+	// EIP-1559 is supported if base fee is not zero.
+	isSupported := baseFee.Uint64() > 0
```
ws4charlie, line 122: I think we should just follow ChainSafe's logic (similar to below). For example, Binance mainnet returns `0` baseFee but it still supports EIP-1559 transactions.

`isSupported := baseFee != nil`

swift1337: Yep, you're right. https://github.com/zeta-chain/node/pull/2483/commits/ccc3598ead10d16816f4adeadd320030e9ff31aa

ws4charlie (approved):

Issue #1904, opened 2024-03-15
feat: Zetaclient should be able to enable newly added chain dynamically at runtime


ws4charlie:
**Describe the Issue**
Dynamically add a new chain is a manual process which will at least involve:

1. Adding new chains to zetacore (whitelisting new chain, etc.)
2. Observers runners update their zetaclient config files and restart zetaclient.

**Expected Outcome**
Zetaclient should be able to detect newly added chains and start working for new chains (EVM chain only for now) dynamically (at runtime). Zetaclient should use `IsSupported` flag in Chain Parameters to detect newly added chains.

Issue #2019, opened 2024-04-15
Reduce logic in `cmd/zetaclientd`


lumtis:
**Describe the Issue**
Currently `cmd/zetaclientd` has to much logic:
- Initialization of ZetaBridge, Signers, clients
- TSS generation
- Initiazation logic in `start.go`

We should keep the logic in `cmd` as minimal as possible. Making an interface interaction with packages in `zetaclient`
We also don't include the logic in `cmd` for the code coverage scope, which is decreasing coverage of potentially imporant logic.

We should move relevant logic in packages in ZetaClient and keep cmd minimal

swift1337:
![image](https://github.com/user-attachments/assets/b6f793fa-4aaa-4ca4-b226-f51ddd7976b7)

## Simplify and consolidate files

We have too many of them, that creates distractions e.g. 

- `main.go` and `root.go` could be a single file
- Same for `utils` and `start_utils`
- Make sure there’s ONLY a single `init()` definition


## debug.go

Either refactor or drop `debug.go` (if not used) https://github.com/zeta-chain/node/blob/8d17118882f50003654e5d582f0b4049f913998f/cmd/zetaclientd/debug.go#L56

- Rename `debug.go` to `get_inbound_ballot.go`
- Remove logic for voting as it is supposed to only fetch the ballot
- Move the logic into the service layer, `cmd/` should only initialize the command and not contain business logic

## tss commands

Consolidate tss commands into a single `tss.go` https://github.com/zeta-chain/node/blob/9b704a5bf924b164ad1472f028323b8b59a102a0/cmd/zetaclientd/keygen_tss.go#L31

- Move tss generation logic into `zetaclient/tss/...`

swift1337:
WIP

swift1337:
start.go will be fixed in this issue https://github.com/zeta-chain/node/issues/3112

PR #2411, opened 2024-07-01
feat: zetaclient should be able to activate/deactivate chains at runtime

ws4charlie:
# Description

- [x] Clean up code and simplified `zetaclient` initialization logics.
- [x] Simplified zetaclient side `AppContext` by consolidating `[ZetacoreContext + AppContext] ==> AppContext`. `ZetacoreContext` is removed.
- [x] Use separate go-routine to detect upgrade plan in zetacore.
- [x] Use separate go-routine to refresh zetaclient side `AppContext`.
- [x] Use separate go-routine to check enabled chains at runtime and activate/deactivate observers/signers dynamically.

Closes: [1904](https://github.com/zeta-chain/node/issues/1904)
Closes: [1397](https://github.com/zeta-chain/node/issues/1397)

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Tested in development environment
- [x] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions

## Summary by CodeRabbit

- **New Features**
  - Added functionality for activating and deactivating chains at runtime in the Zeta client.
  - Introduced tests for the `AppContext` functionality.
  - Added orchestrator methods for watching, updating, activating, and deactivating chains.

- **Bug Fixes**
  - Updated method calls and logic for improved clarity and consistency.

- **Refactor**
  - Replaced usage of logging libraries and updated method signatures in the debug and start commands for improved functionality.
  - Modified the `Orchestrator` struct with new fields and methods.

- **Tests**
  - Updated code generation tool `mockery` to version `v2.42.2` for various test utility packages.
  - Added tests for chain activation and deactivation functionalities.

- **Chores**
  - Reorganized imports and adjusted logger initialization and compliance configuration loading in the start command.

PR #2497, opened 2024-07-17
feat(zetaclient)!: support for runtime chain (de)provisioning 

swift1337:
# Description

This PR implements observers&signers runtime (de)provisioning. So then zetacore lacks `$chain`, `$chainParams`, or `$chainParams.IsSupported == false` zetaclient stops affected workers. Same for the opposite: then a net-new chain exists in the config of zetaclient AND it eventually pops up in chainParams or "becomes" supported, necessary workers are bootstrapped.

It also refactors code a bit and adds couple of useful packages, e.g. `testrpc`.

- Replaces https://github.com/zeta-chain/node/pull/2411
- Closes https://github.com/zeta-chain/node/issues/1904

## Context
In order to add a new chain, the following steps should happen:
- gov proposal for new `Chain` creation
- gov proposal for new `ChainParam` creation
- `zetaclient` config update (e.g. RPC endpoint)
- restart of `zetaclient`

This PR adds logic that periodically checks for updates in chains/chainParams and provisions signers & observers if net-new chain exists in zetaclient config. It also implements automatic deprovisioning  

## Changes
- [x] Refactor SQLite logic move it of `observer` package to `db`
- [x] Simplify appContext configuration, make it more streamlined  
- [x] Move signers map creation to `orchestrator`. Support signers map sync based on `AppContext`
- [x] Move observers map creation to `orchestrator`. Support observers map sync based on `AppContext`
- [x] `zetaclientd`: program exits if operator is NOT an observer 🚨 
- [x] Add BTC, EVM, and SOL RPCs mocks using `httptest`

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Added support for runtime chain (de)provisioning.
  - Introduced a database management system for handling SQLite connections and migrations.

- **Bug Fixes**
  - Improved error handling and logging throughout various modules to provide clearer diagnostics.

- **Documentation**
  - Updated changelog to reflect new features and changes.

- **Tests**
  - Enhanced test cases for various functionalities, utilizing in-memory databases for improved reliability and performance.


zetaclient/orchestrator/bootstap_test.go
```diff
@@ -0,0 +1,492 @@
+package orchestrator
```
lumtis, line 1: nit - Can we consider maybe having a separate package for the test? `orchestrator_test`

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		return 0, 0, errors.Wrapf(err, "failed to get app context")
+	}
+
+	var (
+		added, removed int
+
+		presentChainIDs = make([]int64, 0)
+
+		onAfterSet = func(chainID int64, _ interfaces.ChainSigner) {
```
lumtis, line 67: Can we reuse added/removed terminology here instead of set/unset?

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,343 @@
...
+			config.GetERC20CustodyABI(),
+			mpiAddress,
+			erc20CustodyAddress,
+		)
+		if err != nil {
+			logger.Std.Error().Err(err).Msgf("Unable to construct signer for EVM chain %d", chainID)
+			continue
+		}
+
+		mapSet[int64, interfaces.ChainSigner](signers, chainID, signer, onAfterSet)
```
lumtis, line 125: Might be good to add comment to explain as`mapSet` is generic name

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,343 @@
...
+
+		cfg, _ := app.Config().GetBTCConfig()
+
+		utxoSigner, err := btcsigner.NewSigner(btcChain, tss, ts, logger, cfg)
+		if err != nil {
+			logger.Std.Error().Err(err).Msgf("Unable to construct signer for UTXO chain %d", chainID)
+			continue
+		}
+
+		mapSet[int64, interfaces.ChainSigner](signers, chainID, utxoSigner, onAfterSet)
```
lumtis, line 158: same comment as above

zetaclient/orchestrator/mapping.go
```diff
@@ -0,0 +1,49 @@
+package orchestrator
```
lumtis, line 1: We can maybe move to `pkg`?

zetaclient/orchestrator/orchestrator.go
```diff
@@ -518,3 +565,58 @@ func (oc *Orchestrator) ScheduleCctxBTC(
 		}
 	}
 }
+
+func (oc *Orchestrator) runObserverSignerSync(ctx context.Context) error {
```
lumtis, line 569: Might be good to give comment even if it's private method

lumtis (commented):
Overall changes look good to me, just a few comments

Seems like this issue can close https://github.com/zeta-chain/node/issues/2019 ?

zetaclient/orchestrator/mapping.go
swift1337: I'd keep it here for now, not mature enough to be moved IMO

zetaclient/orchestrator/bootstap_test.go
swift1337: Go distinguishes $package and $package_test. Generally, I'd use $package_test if regular $package conflicts with imports.

swift1337:
> Seems like this issue can close #2019 ?

I would say it closes it only partially as we still lots of `todo: revamp`

zetaclient/chains/base/observer.go
```diff
@@ -121,18 +121,36 @@ func NewObserver(
...
+	ob.started = true
+
+	return false
+}
+
 // Stop notifies all goroutines to stop and closes the database.
 func (ob *Observer) Stop() {
-	ob.logger.Chain.Info().Msgf("observer is stopping for chain %d", ob.Chain().ChainId)
+	ob.logger.Chain.Info().Msgf("Stopping observer for chain %d", ob.Chain().ChainId)
 	close(ob.stop)
```
ws4charlie, line 142: move `close(ob.stop)` to the line after acquiring the lock? The `Stop()` is called by both
https://github.com/zeta-chain/node/blob/feat%2Fzetaclient-runtime-chains/cmd/zetaclientd/start.go#L334
and
https://github.com/zeta-chain/node/blob/feat%2Fzetaclient-runtime-chains/zetaclient/zetacore/client.go#L369

calling `close(ob.stop)` twice (the upgrade plan monitor and listener to system signals) will panic even though the chance is low.

zetaclient/orchestrator/orchestrator.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/2f06e5639910af4fbfeb4261ff0faf4a86b53af1

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/2f06e5639910af4fbfeb4261ff0faf4a86b53af1

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/2f06e5639910af4fbfeb4261ff0faf4a86b53af1

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/2f06e5639910af4fbfeb4261ff0faf4a86b53af1

zetaclient/chains/base/observer.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/cc5f6530f18f26bb26cabfd0428c74f9ff7e31c9

zetaclient/chains/evm/observer/observer.go
```diff
@@ -166,6 +167,11 @@ func FetchZetaTokenContract(
 
 // Start all observation routines for the evm chain
 func (ob *Observer) Start(ctx context.Context) {
+	if noop := ob.Observer.Start(); noop {
+		ob.Logger().Chain.Info().Msgf("observer is already started for chain %d", ob.Chain().ChainId)
```
ws4charlie, line 171: Will this log print a lot when the bootstrapper keep starting a chain that have already started? Just curious

swift1337: It calculated the diff only. New observers/signers won't start if present.
- https://github.com/zeta-chain/node/blob/2f06e5639910af4fbfeb4261ff0faf4a86b53af1/zetaclient/orchestrator/bootstrap.go#L152
- https://github.com/zeta-chain/node/blob/f2d14dfe0023cad7502ba4e2a663d2b81cd484a9/zetaclient/orchestrator/bootstap_test.go#L205

zetaclient/chains/solana/observer/observer.go
```diff
@@ -108,6 +112,11 @@ func (ob *Observer) GetChainParams() observertypes.ChainParams {
 
 // Start starts the Go routine processes to observe the Solana chain
 func (ob *Observer) Start(ctx context.Context) {
+	if noop := ob.Observer.Start(); noop {
+		ob.Logger().Chain.Info().Msgf("observer is already started for chain %d", ob.Chain().ChainId)
```
ws4charlie, line 116: same question

zetaclient/context/app.go
```diff
@@ -81,14 +64,17 @@ func (a *AppContext) Config() config.Config {
 
 // GetBTCChainAndConfig returns btc chain and config if enabled
 func (a *AppContext) GetBTCChainAndConfig() (chains.Chain, config.BTCConfig, bool) {
-	btcConfig, configEnabled := a.Config().GetBTCConfig()
-	btcChain, _, paramsEnabled := a.GetBTCChainParams()
+	cfg, configEnabled := a.Config().GetBTCConfig()
```
ws4charlie, line 67: Here is place for future refactor you can track

1. Get rid of `GetBTCChainAndConfig` and `GetSolanaChainAndConfig` and use one single method similar to below to return a deep copy of ChainParams for chain bootstrap or chain parameter updating. Not for this PR.

`GetExternalChainParams(chainID int64) (observertypes.ChainParams, bool)`

swift1337: Correct, we can resolve everything based on `chainID` + `IsEVMChain()`, `IsUTXOChain()`, `IsSolanaChain()`, ...

zetaclient/context/app.go
```diff
@@ -306,25 +292,33 @@ func (a *AppContext) Update(
...
+		}
+
+		freshEvmChainParams[cp.ChainId] = cp
+	}
+
+	a.evmChainParams = freshEvmChainParams
+
 	// update chain params for bitcoin if it has config in file
-	if a.bitcoinChainParams != nil && btcChainParams != nil {
+	if btcChainParams != nil {
```
ws4charlie, line 313: do we shut down Bitcoin observer/signer if newly fetched `btcChainParams` is nil? or a future plan?

zetaclient/context/app.go
```diff
@@ -306,25 +292,33 @@ func (a *AppContext) Update(
...
 		a.bitcoinChainParams = btcChainParams
 	}
 
 	// update chain params for solana if it has config in file
-	if a.solanaChainParams != nil && solChainParams != nil {
+	if solChainParams != nil {
```
ws4charlie, line 318: similar

zetaclient/db/db.go
```diff
@@ -0,0 +1,112 @@
...
+
+// read/write/execute for user
+// read/write for group
+const dirCreationMode = 0o750
+
+var (
+	defaultGormConfig = &gorm.Config{
+		Logger: logger.Default.LogMode(logger.Silent),
+	}
+	migrationEntities = []any{
```
ws4charlie, line 29: can we have 3 groups of migration Entities. One for EVM chain, one for Bitcoin and last one for Solana? They need their own tables.

zetaclient/orchestrator/mapping.go
```diff
@@ -0,0 +1,53 @@
...
+	if beforeUnset != nil {
+		beforeUnset(key, (*m)[key])
+	}
+
+	delete(*m, key)
+
+	return true
+}
+
+// mapDeleteMissingKeys removes signers from the map IF they are not in the presentKeys.
```
ws4charlie, line 41: // mapDeleteMissingKeys removes observers/signers from the map IF they are not in the presentKeys.

zetaclient/orchestrator/mapping.go
```diff
@@ -0,0 +1,53 @@
...
+func mapDeleteMissingKeys[K cmp.Ordered, V any](m *map[K]V, presentKeys []K, beforeUnset func(K, V)) {
+	presentKeysSet := make(map[K]struct{}, len(presentKeys))
```
ws4charlie, line 43: Maybe using type `make(map[K]bool, len(presentKeys))` will help to simplify the statement `if _, isPresent := presentKeysSet[key]; !isPresent` ?

zetaclient/orchestrator/orchestrator.go
```diff
@@ -518,3 +565,60 @@ func (oc *Orchestrator) ScheduleCctxBTC(
...
+			Int("observer.removed", removed).
+			Msg("synced observers")
+	}
+
+	added, removed, err = syncSignerMap(ctx, oc.tss, oc.baseLogger, oc.ts, &oc.signerMap)
+	if err != nil {
+		return errors.Wrap(err, "syncSignerMap failed")
+	}
+
+	if added+removed > 0 {
```
ws4charlie, line 616: Are newly created or newly deleted observer and signer pair always exactly match?

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+		if evmConfig.Chain.IsZetaChain() {
+			continue
+		}
+
+		evmChainParams, found := app.GetEVMChainParams(chainID)
+		switch {
+		case !found:
+			logger.Std.Warn().Msgf("Unable to find chain params for EVM chain %d", chainID)
+			continue
+		case !evmChainParams.IsSupported:
```
ws4charlie, line 97: The `IsSupported` flag is supposed to be `true` because it is fetched [from the query GetSupportedChains](https://github.com/zeta-chain/node/blob/feat%2Fzetaclient-runtime-chains/zetaclient/zetacore/client.go#L402). If we do want to double check it, we can just check it right in `client.go`.

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+			logger.Std.Info().Msgf("Removing signer for chain %d", chainID)
+			removed++
+		}
+	)
+
+	// EVM signers
+	for _, evmConfig := range app.Config().GetAllEVMConfigs() {
+		chainID := evmConfig.Chain.ChainId
+
+		if evmConfig.Chain.IsZetaChain() {
```
ws4charlie, line 88: Can we move this check to upstream code in the method `UpdateZetacoreContext` in `client.go` with similar logic like
```
// skip ZetaChain
if !chain.IsExternalChain() {
	continue
}
```

Note: we'll probably remove Zetachain in the future.

zetaclient/context/app_test.go
```diff
@@ -69,37 +70,48 @@ func TestNew(t *testing.T) {
 	})
 
 	t.Run("should create new zetacore context with config containing evm chain params", func(t *testing.T) {
-		testCfg := config.New(false)
+		// ARRANGE
```
ws4charlie, line 73: There are some remaining `zetacore context`, `coreContext` in the codebase. Since we removed the concept of zetacore context, leaving only app context, we need a text search and replace them with `app context` and `appContext`.

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+	evmsigner "github.com/zeta-chain/zetacore/zetaclient/chains/evm/signer"
+	"github.com/zeta-chain/zetacore/zetaclient/chains/interfaces"
+	solbserver "github.com/zeta-chain/zetacore/zetaclient/chains/solana/observer"
+	"github.com/zeta-chain/zetacore/zetaclient/config"
+	zctx "github.com/zeta-chain/zetacore/zetaclient/context"
+	"github.com/zeta-chain/zetacore/zetaclient/db"
+	"github.com/zeta-chain/zetacore/zetaclient/metrics"
+)
+
+// backwards compatibility
```
ws4charlie, line 25: here is a bit more description about the constant
`// btcDatabaseFilename is the Bitcoin database file name now used in mainnet so we keep using it here for backward compatibility`

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+			continue
+		case !chainParams.IsSupported:
+			logger.Std.Error().Msgf("EVM chain %d is not supported", chainID)
+			continue
+		}
+
+		presentChainIDs = append(presentChainIDs, chainID)
+
+		// noop
+		if mapHas(observerMap, chainID) {
```
ws4charlie, line 255: As mentioned earlier in Slack. For incoming refactor, once getting rid of `GetAllEVMConfigs()`, `GetBTCChainAndConfig` etc., we don't need to check `mapHas` any more as there will be only one single map in the app context exposing two methods:
1. `GetEnabledBTCChains() []chains.Chain`
2. `GetExternalChainParams(chainID int64)`

we'll simply iterate all supported chains and get their corresponding chain params by a chain ID. You can still refer to the implementation [here](https://github.com/zeta-chain/node/blob/orchestrator-add-new-chain-at-runtime/zetaclient/context/app_context.go#L123)

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+			logger.Std.Error().Err(err).Msgf("NewObserver error for EVM chain %s", evmConfig.Chain.String())
+			continue
+		}
+
+		addObserver(chainID, observer)
+	}
+
+	// Emulate same loop semantics as for EVM chains
+	// create BTC chain observer
+	for i := 0; i < 1; i++ {
```
ws4charlie, line 294: Not for this PR. For tracking future refactor, the hardcoded loop can be improved similar to [this loop](https://github.com/zeta-chain/node/blob/orchestrator-add-new-chain-at-runtime/zetaclient/orchestrator/chain_activate.go#L215)

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+			database,
+			logger,
+			ts,
+		)
+		if err != nil {
+			logger.Std.Error().Err(err).Msgf("NewObserver error for EVM chain %s", evmConfig.Chain.String())
+			continue
+		}
+
+		addObserver(chainID, observer)
```
ws4charlie, line 289: As mentioned in Slack. It's not easy to guarantee that the observer and signer are always added in pairs as we use separate function `syncObserverMap` and `syncSignerMap` to add them separately.

For incoming refactor, once we can rid of `GetAllEVMConfigs()`, `GetBTCChainAndConfig` etc. and use only one single map in the app context exposing two methods:
1. `GetEnabledExternalChains() []chains.Chain`
2. `GetExternalChainParams(chainID int64)`

we'll simply iterate all supported chains and get their corresponding chain params by a chain ID to create observer/signer in pairs. You can still refer to the implementation [here](https://github.com/zeta-chain/node/blob/orchestrator-add-new-chain-at-runtime/zetaclient/context/app_context.go#L123)

To make function shorter (now `syncObserverMap` spans 200+ lines and will increase if more chain comes in) and easier to unit test, you can create smaller functions like `syncObserverSignerMapEVM`, `syncObserverSignerMapBTC`, `syncObserverSignerMapSolana` to sync observer/signer pairs by consensus type such as `Consensus_ethereum`, `Consensus_bitcoin` and `Consensus_solana_consensus` etc.

zetaclient/orchestrator/bootstrap.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+
+		onBeforeRemove = func(chainID int64, _ interfaces.ChainSigner) {
+			logger.Std.Info().Msgf("Removing signer for chain %d", chainID)
+			removed++
+		}
+	)
+
+	// EVM signers
+	for _, evmConfig := range app.Config().GetAllEVMConfigs() {
```
ws4charlie, line 85: Just to track. The config file reloader would be a piece of future refactor to allow the observer signer to feed config (e.g. their own endpoint for the chain) of newly enabled chains.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -518,3 +565,60 @@ func (oc *Orchestrator) ScheduleCctxBTC(
 		}
 	}
 }
+
+// runObserverSignerSync runs a blocking ticker that observes chain changes from zetacore
+// and optionally (de)provisions respective observers and signers.
+func (oc *Orchestrator) runObserverSignerSync(ctx context.Context) error {
+	const cadence = 5 * time.Second
```
ws4charlie, line 572: If you want to use a constant, `ZetaBlockTime` (or `ZetaBlockTime * 2`) are good options as the chain state (chains, parameters are part of chain state) changes block by block.

zetaclient/orchestrator/bootstrap.go
swift1337: Yes, 100%

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/issues/2519

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/issues/2519

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/issues/2519

zetaclient/context/app.go
```diff
@@ -277,15 +263,15 @@ func (a *AppContext) Update(
 	blockHeaderEnabledChains []lightclienttypes.HeaderSupportedChain,
 	init bool,
 ) {
+	if len(newChains) == 0 {
```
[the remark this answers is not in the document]
swift1337: Correct, but I don't think we can hit this case. I can only imagine a gov proposal that disables support for all ChainParams

zetaclient/context/app.go
swift1337: Let's tackle this in the following PRs.

zetaclient/db/db.go
swift1337: https://gorm.io/docs/migration.html

GORM will create all required tables  based on this list. I don't think this is an issue.

Also, using GORM is an overkill in our case, we can consider simplifying zetaclient's persistent store. Also, we need to build the client in a way that allows loss/flushing sqlite db.

zetaclient/orchestrator/mapping.go
swift1337: Using `map[K]struct{}{}` is a common Go way to avoid unnecessary allocation if maps's values.

zetaclient/orchestrator/orchestrator.go
swift1337: That's a good question. We can continue the discussion here https://github.com/zeta-chain/node/issues/2519

zetaclient/orchestrator/bootstrap.go
swift1337: I prefer to keep this because this code is not aware of the source of ChainParams.

zetaclient/context/app_test.go
swift1337: Created an issue https://github.com/zeta-chain/node/issues/2526

zetaclient/orchestrator/bootstrap.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/322af8c88b4b5aad1c1fb9119f5ee3125817d82e

zetaclient/orchestrator/mapping.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/322af8c88b4b5aad1c1fb9119f5ee3125817d82e

zetaclient/orchestrator/mapping.go
swift1337: https://github.com/zeta-chain/node/pull/2497/commits/322af8c88b4b5aad1c1fb9119f5ee3125817d82e

lumtis (approved):

zetaclient/chains/evm/observer/observer.go
ws4charlie: Not sure, just curious. When will `ob.Observer.Start()` returns a `true` and print the log?

It seems the callback function `onAfterAdd` is called on a newly created observer on [this line](https://github.com/zeta-chain/node/blob/feat%2Fzetaclient-runtime-chains/zetaclient/orchestrator/bootstrap.go#L266), which contains a `false` on its `started` field.

zetaclient/orchestrator/bootstrap.go
ws4charlie: If we really want to enforce something, I think enforcing it in the very beginning source could stop spreading checking in all downstream code. Just my opinion.

zetaclient/db/db.go
ws4charlie: Not sure if I clarified it well or not. Each chain creates its own database, they just need their own specific data structures in database.

Bitcoin needs two structs: LastBlockSQLType, TransactionResultSQLType, OutboundHashSQLType
Solana needs one struct:   LastTransactionSQLType
EVM needs three structs:  LastBlockSQLType, TransactionSQLType, ReceiptSQLType

The current impl is creating same group of tables for all chains (for example, created receipt table for Bitcoin, and bitcoin doesn't have a concept or receipt). Not sure what's the benefit of it if the chain doesn't need that table (even in future).

ws4charlie (commented):
looks good. Just a few clarifications.

Issue #2478, opened 2024-07-15
Add support for passing data from Bitcoin as inscriptions


fadeev:
**Is your feature request related to a problem? Please describe.**

Currently, ZetaChain supports passing data using the op_return opcode, which has an 80-byte limit. This limitation restricts the amount of data that can be included in cross-chain transactions, posing challenges for applications that require more data, such as omnichain NFTs and complex smart contract interactions.

**Describe the solution you'd like**

To address this limitation, we propose adding support for passing data as inscriptions. Inscriptions allow for longer messages by utilizing the witness portion of a Bitcoin transaction, enabling applications to include more comprehensive data.

Inscriptions involve wrapping data into a Taproot script and injecting it into the witness portion of a Bitcoin transaction. This process allows for data pushes up to 520 bytes, and larger data files can be managed through multiple data pushes. Creating an inscription requires a commit transaction (hash reference) and a reveal transaction (revealing the entire script).

We propose maintaining support for both op_return and inscriptions. The advantage of op_return is that it allows for a single transaction, making it simpler and potentially less costly for users. On the other hand, inscriptions support longer messages, providing more flexibility for applications that need to pass larger amounts of data. A user-facing app could potentially choose which method to use depending on the length of the data being passed.
```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/2728
- [ ] https://github.com/zeta-chain/node/pull/2524
- [ ] https://github.com/zeta-chain/node/pull/2533
- [ ] https://github.com/zeta-chain/node/pull/2727
- [ ] https://github.com/zeta-chain/node/issues/2534
```

bitSmiley:
# Zetachain Extending EVM Calldata Size From Bitcoin

## Introduction
Currently btc to zeta ominchain call cannot contain arbitrary long evm contract call data length. The current limit is 80
bytes and this is probably too much of a restriction.

From a technical point of view, this comes from BTC "[OP_RETURN](https://github.com/zeta-chain/node/blob/a0868bda2106cc958eea5b65d258a0bacdc6a766/zetaclient/chains/bitcoin/observer/inbound.go#L446)" opcode that allows only 80 bytes.

This document proposals a possible solution to extend to a larger calldata size by utilizing tapscript based solution,
very much like inscription.

## Tapscript
Similar to inscriptions used, this approach makes use of the tapscript in bitcoin. Unlike inscription, the current protocol
[...]

Issue #2506, opened 2024-07-18
Integrate Solana Mainnet


lumtis:
**Describe the Issue**
This EPIC is an extension of https://github.com/zeta-chain/node/issues/2401

It is used to track every tasks necessary to fully support Solana on Mainnet
- Integration
- Optimization to do after testnet integration
- Any bugs occuring during testnet experimentation

```[tasklist]
### Tasks
- [ ] https://github.com/zeta-chain/node/issues/2401
- [ ] https://github.com/zeta-chain/protocol-contracts-solana/issues/8
- [ ] https://github.com/zeta-chain/node/issues/2659
- [ ] https://github.com/zeta-chain/node/issues/2507
- [ ] https://github.com/zeta-chain/node/issues/2508
- [ ] https://github.com/zeta-chain/node/issues/2614
- [ ] https://github.com/zeta-chain/node/issues/2606
- [ ] https://github.com/zeta-chain/node/issues/2692
```

PR #2524, opened 2024-07-21
feat: parse inscription like witness data

bitSmiley:
# Description

This is the start of a series of PRs that implements the solution mentioned in #2478. The implementation is broken down into several parts:
- [x] Extract memo from the witness script (This PR) 
- [ ] Detect targetted zeta btc address and extract corresponding witness script
- [ ] Javascript client that submits inscription transactions
- [ ] Integration testing in code base and in mainet

This PR adds a `DecodeScript` function that decodes memo wrapped in an inscription-like script.  The format of the script is following that of "inscription" defined in ordinal theory. However, to separate from inscription (as this use case is not an NFT), simplifications are made. The bitcoin envelope script is as follows:
```
OP_DATA_32 <32 byte of public key> OP_CHECKSIG
OP_FALSE
OP_IF
  OP_PUSH 0x...
  OP_PUSH 0x...
OP_ENDIF
```
The public key is the `internalKey` used in the taproot transaction. This follows `ord`.

The `DecodeScript` method take the script bytes and parse them according. Internally it is utilizing `ScriptTokenizer`. Ideally, it should directly use `txscript.ScriptTokenizer`, however, the current codebase does not have the versions for `ScriptTokenizer`. A simplified version is used. If `btcsuite` is upgraded, then `scriptTokenizer` can be removed and the code base will be simplified further.

// There are no content-type or any other attributes, it's just raw bytes.
# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced Bitcoin event processing with a new function to handle transactions containing witness data.
	- Added functionalities for decoding Bitcoin scripts, improving the robustness and validation of transaction scripts.

- **Tests**
	- Introduced a comprehensive test suite for the new script decoding functionality, validating various scenarios and error conditions.


zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -192,6 +193,36 @@ func DecodeOpReturnMemo(scriptHex string, txid string) ([]byte, bool, error) {
 	return nil, false, nil
 }
 
+// DecodeScript decodes memo wrapped in an inscription like script in witness
```
lumtis, line 196: Not related to this PR directly but I would consider moving this entire file in an independent package: https://github.com/zeta-chain/node/issues/2525

But we can let as it in this PR to keep it minimal

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,39 @@ func GetBtcEvent(
...
+		tx,
+		tssAddress,
+		blockNumber,
+		logger,
+		netParams,
+		depositorFee,
+	)
+
+	if err != nil { // should never happen
+		return nil, err
```
swift1337, line 501:
```suggestion
		return nil, errors.Wrap(err, "unable to get btc event")
```

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,39 @@ func GetBtcEvent(
 	}
 	return nil, nil
 }
+
+// GetBtcEventWithWitness either returns a valid BTCInboundEvent or nil.
+// This method supports data with more than 80 bytes by scanning the witness for possible presence of a tapscript.
+// It will first prioritize OP_RETURN over tapscript.
+// The format of the tapscript is
+func GetBtcEventWithWitness(
+	rpcClient interfaces.BTCRPCClient,
```
swift1337, line 481:
```suggestion
	client interfaces.BTCRPCClient,
```

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -245,3 +276,211 @@
...
+
+		if next < txscript.OP_DATA_1 || next > txscript.OP_PUSHDATA4 {
+			return nil, fmt.Errorf("expecting data push, found %d", next)
+		}
+
+		memo = append(memo, t.Data()...)
+	}
+
+	return memo, nil
+}
```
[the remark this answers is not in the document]
lumtis: I think this suggestion makes sense

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,39 @@ func GetBtcEvent(
 	}
 	return nil, nil
 }
+
+// GetBtcEventWithWitness either returns a valid BTCInboundEvent or nil.
+// This method supports data with more than 80 bytes by scanning the witness for possible presence of a tapscript.
+// It will first prioritize OP_RETURN over tapscript.
+// The format of the tapscript is
```
swift1337, line 479: Should this comment be extended?

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -245,3 +276,211 @@ func DecodeTSSVout(vout btcjson.Vout, receiverExpected string, chain chains.Chai
...
+
+// checkInscriptionEnvelope decodes the envelope for the script monitoring. The format is
+// OP_PUSHBYTES_32 <32 bytes> OP_CHECKSIG <Content>
+func checkInscriptionEnvelope(t *scriptTokenizer) error {
+	if !t.Next() || t.Opcode() != txscript.OP_DATA_32 {
+		return fmt.Errorf("cannot obtain public key bytes")
+	}
+
+	if !t.Next() || t.Opcode() != txscript.OP_CHECKSIG {
+		return fmt.Errorf("cannot parse OP_CHECKSIG")
```
lumtis, line 334: Maybe we can write the current opcode in the error?

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -245,3 +276,211 @@ func DecodeTSSVout(vout btcjson.Vout, receiverExpected string, chain chains.Chai
...
+	}
+
+	return nil
+}
+
+func makeScriptTokenizer(script []byte) scriptTokenizer {
```
lumtis, line 340:
```suggestion
func newScriptTokenizer(script []byte) scriptTokenizer {
```
I think new termninology is more adapted?

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -245,3 +276,211 @@ func DecodeTSSVout(vout btcjson.Vout, receiverExpected string, chain chains.Chai
...
+		}
+
+		// Move the offset forward and set the opcode and data accordingly.
+		t.offset += length
+		t.op = op
+		t.data = script[1:length]
+
+		return true
+	case op > txscript.OP_PUSHDATA4:
+		t.err = fmt.Errorf("unexpected op code")
```
lumtis, line 435: Same, could we add the value of the opcode here?

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -192,6 +193,36 @@ func DecodeOpReturnMemo(scriptHex string, txid string) ([]byte, bool, error) {
...
+//	OP_PUSH 0x...
+//	OP_PUSH 0x...
+//
+// OP_ENDIF
+// There are no content-type or any other attributes, it's just raw bytes.
+func DecodeScript(script []byte) ([]byte, bool, error) {
+	t := makeScriptTokenizer(script)
+
+	if err := checkInscriptionEnvelope(&t); err != nil {
+		return nil, false, err
```
swift1337, line 215:
```suggestion
		return nil, false, errors.Wrap(err, "unable to check the envelope")
```

lumtis (commented):
>This is the start of a series of PRs that implements the solution mentioned in https://github.com/zeta-chain/node/issues/2478. The implementation is broken down into several parts:

Sounds good. For task that can be broken down in subset, we usually create issue tagged as `epic`

Example https://github.com/zeta-chain/node/issues/2506

zetaclient/chains/bitcoin/tx_script.go
swift1337, line 340: can we move this to `zetaclient/chains/bitcoin/tokenizer`?

`t := tokenizer.New(script) *Tokenizer`

zetaclient/chains/bitcoin/tx_script_test.go
```diff
@@ -491,3 +491,66 @@ func TestDecodeTSSVoutErrors(t *testing.T) {
...
...1dd05f27993d22c42988758882d20080241074462884c8774e1cdf4b04e5b3b74b6568bd1769722708306c66270b6b2a7f68baced83627eeeb2d494e8a1749277b92a4c5a90b1b4f6038e5f704405515109d4d0021612ad298b8dad6e12245f8f0020e11a7a319652ba6abe261958201ce5e83131cd81302c0ecec60d4afa9f72540fc84b6b9c1f3d903ab25686df263b192a403a4aa22b799ba24369c49ff4042012589a07d4211e05f80f18a1262de5a1577ce0ec9e1fa9283cfa25d98d7d0b4217951dfcb8868570318c63f1e1424cfdb7d7a33c6b9e3ced4b2ffa0178b3a5fac8bace2991e382a402f56a2c6a9191463740910056483e4fd0f5ac729ffac66bf1b3ec4570c4e75c116f7d9fd65718ec3ed6c7647bf335b77e7d6a4e2011276dc8031b78403a1ad82c92fb339ec916c263b6dd0f003ba4381ad5410e90e88effbfa7f961b8e8a6011c525643a434f7abe2c1928a892cc57d6291831216c4e70cb80a39a79a3889211070e767c23db396af9b4c2093c3743d8cbcbfcb73d29361ecd3857e94ab3c800be1299fd36a5685ec60607a60d8c2e0f99ff0b8b9e86354d39a43041f7d552e95fe2d33b6fc0f540715da0e7e1b344c778afe73f82d00881352207b719f67dcb00b4ff645974d4fd7711363d26400e2852890cb6ea9cbfe63ac43080870049b1023be984331560c6350bb64da52b4b81bc8910934915f0a96701f4c50646d5386146596443bee9b2d116706e1687697fb42542196c1d764419c23a914896f9212946518ac59e1ba5d1fc37e503313133ebdf2ced5785e0eaa9738fe3f9ad73646e733931ebb7cff26e96106fe68"
```
swift1337, line 498: can we generate this data on the fly? Also would be cool to have some real-world use-case of having the following logic:
```
Given an ABI
Encode some call data using inscriptions 
Assert memo 
Make sure that passing memo further allows proper ABI decoding and calling

**Review by swift1337 — COMMENTED** [review]
Thank you for contributing! Looks awesome, left some comments.

File: zetaclient/chains/bitcoin/tx_script_test.go (hunk 10 above)
Comment (bitSmiley) [reply zetaclient/chains/bitcoin/tx_script_test.go:498]: In the following PR will contain parsing the txn data to extract the witness. Then a real mainet txn will be tested. There will be a nodejs repo for the use case you mentioned. Regarding the generate this data on the fly, might need to look into the existing library script building capability, currently it's generated using nodejs.

File: zetaclient/chains/bitcoin/tx_script.go (hunk 7 above)
Comment (bitSmiley) [reply zetaclient/chains/bitcoin/tx_script.go:340]: Actually it would be great to update the btcsuite, then this code can be deleted, haha

File: zetaclient/chains/bitcoin/observer/inbound.go (hunk 5 above)
Comment (bitSmiley) [reply zetaclient/chains/bitcoin/observer/inbound.go:479]: oops, will update

**Comment by bitSmiley** [comment]
@lumtis @swift1337 I have addressed the review feedbacks. I'm not familiar with the "resolve" convention you have, so I just leave them open first. Please take another look and let me know.

File: zetaclient/chains/bitcoin/observer/inbound.go [hunk 11]
```diff
@@ -472,3 +472,39 @@ func GetBtcEvent(
...
+	)
+
+	if err != nil { // should never happen
+		return nil, errors.Wrap(err, "unable to get btc event")
+	}
+	if event != nil {
+		return event, nil
+	}
+
+	// TODO: integrate parsing script
```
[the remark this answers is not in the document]
lumtis: @coderabbitai yes, can you create an issue for this and suggest a comment line here to link the issue?

lumtis (commented):
Looks like we need to run
```
make generate
```
in the repo and commit the changes

bitSmiley:
@lumtis Ran `make generate` and pushed the results

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -477,3 +477,38 @@ func GetBtcEvent(
...
+		netParams,
+		depositorFee,
+	)
+
+	if err != nil { // should never happen
+		return nil, errors.Wrap(err, "unable to get btc event")
+	}
+	if event != nil {
+		return event, nil
+	}
```
fbac, line 509: nit: just identation changes. The comment does not add a lot of value, so we can skip it.
```suggestion

	if err != nil {
		return nil, errors.Wrap(err, "unable to get btc event")
	}

	if event != nil {
		return event, nil
	}
```

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
+package bitcoin
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+)
+
+func newScriptTokenizer(script []byte) scriptTokenizer {
```
fbac, line 10: Does `script` have a min or max bytes size that we should be checking here?

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+		t.op = op
+		t.data = nil
+		return true
+
+	// Data pushes of specific lengths -- OP_DATA_[1-75].
+	case op >= txscript.OP_DATA_1 && op <= txscript.OP_DATA_75:
+		script := t.script[t.offset:]
+
+		// add 2 instead of 1 because script includes the opcode as well
+		length := int(op) - txscript.OP_DATA_1 + 2
```
fbac, line 91: question: to understand this line, why OP_DATA_1 has to be subtracted?

It seems like OP_DATA_1 has value 1, so we're doing `X - 1 + 2` instead of simply `X + 1`? That would give us always the script length plus the opcode?

Not sure if I'm missing something.

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+		t.offset += length
+		t.op = op
+		t.data = script[1:length]
+
+		return true
+	case op > txscript.OP_PUSHDATA4:
+		t.err = fmt.Errorf("unexpected op code %d", op)
+		return false
+	// Data pushes with parsed lengths -- OP_PUSHDATA{1,2,4}.
+	default:
```
fbac, line 108: Just some linting.
```suggestion
		t.data = script[1:length]
		return true

	case op > txscript.OP_PUSHDATA4:
		t.err = fmt.Errorf("unexpected op code %d", op)
		return false

	// Data pushes with parsed lengths -- OP_PUSHDATA{1,2,4}.
	default:
```

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Data pushes of specific lengths -- OP_DATA_[1-75].
+	case op >= txscript.OP_DATA_1 && op <= txscript.OP_DATA_75:
+		script := t.script[t.offset:]
+
+		// add 2 instead of 1 because script includes the opcode as well
+		length := int(op) - txscript.OP_DATA_1 + 2
+		if len(script) < length {
+			t.err = fmt.Errorf("opcode %d requires %d bytes, but script only "+
+				"has %d remaining", op, length, len(script))
```
fbac, line 94: nit: 
Consider shortening and making the error more specific, as it's repeated in the code. Concatenating strings is not optimal.
```suggestion
			t.err = fmt.Errorf("opcode %d detected, but script only %d bytes remaining", op, len(script))
```

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+			length = 1
+		case txscript.OP_PUSHDATA2:
+			length = 2
+		default:
+			length = 4
+		}
+
+		script := t.script[t.offset+1:]
+		if len(script) < length {
+			t.err = fmt.Errorf("opcode %d requires %d bytes, but script only "+
```
fbac, line 121: Same here, avoid concatenating.

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+	default:
+		var length int
+		switch op {
+		case txscript.OP_PUSHDATA1:
+			length = 1
+		case txscript.OP_PUSHDATA2:
+			length = 2
+		default:
+			length = 4
+		}
```
fbac, line 117: Default to a defensive approach:

```suggestion
		switch op {
		case txscript.OP_PUSHDATA1:
			length = 1
		case txscript.OP_PUSHDATA2:
			length = 2
		case txscript.OP_PUSHDATA4:
		        length = 4
		default:
			return fmt.Errorf("unexpected op code %d", op)
		}
```

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+
+		// Next -length bytes are little endian length of data.
+		var dataLen int
+		switch length {
+		case 1:
+			dataLen = int(script[0])
+		case 2:
+			dataLen = int(binary.LittleEndian.Uint16(script[:2]))
+		case 4:
+			dataLen = int(binary.LittleEndian.Uint32(script[:4]))
```
fbac, line 134: Avoiding magic numbers when possible.

```suggestion
		case 2:
			dataLen = int(binary.LittleEndian.Uint16(script[:length]))
		case 4:
			dataLen = int(binary.LittleEndian.Uint32(script[:length]))
```

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+			t.err = fmt.Errorf("invalid opcode length %d", length)
+			return false
+		}
+
+		// Move to the beginning of the data.
+		script = script[length:]
+
+		// Disallow entries that do not fit script or were sign extended.
+		if dataLen > len(script) || dataLen < 0 {
+			t.err = fmt.Errorf("opcode %d pushes %d bytes, but script only "+
```
fbac, line 145: Same consideration on the error as before.

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+				"has %d remaining", op, dataLen, len(script))
+			return false
+		}
+
+		// Move the offset forward and set the opcode and data accordingly.
+		t.offset += 1 + length + dataLen
```
fbac, line 151: question: is `1` the opcode byte? should we `t.offset += int(op) + length + dataLength`?

zetaclient/chains/bitcoin/tokenizer.go
```diff
@@ -0,0 +1,156 @@
...
+	script []byte
+	offset int
+	op     byte
+	data   []byte
+	err    error
+}
+
+// Done returns true when either all opcodes have been exhausted or a parse
+// failure was encountered and therefore the state has an associated error.
+func (t *scriptTokenizer) Done() bool {
```
fbac, line 31: Consideration for all the scriptTokenizer methods: are we going to use them outside the `bitcoin` package? If that's not the case we can simply make them private.

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -245,3 +275,44 @@ func DecodeTSSVout(vout btcjson.Vout, receiverExpected string, chain chains.Chai
...
+func decodeInscriptionPayload(t *scriptTokenizer) ([]byte, error) {
+	if !t.Next() || t.Opcode() != txscript.OP_FALSE {
+		return nil, fmt.Errorf("OP_FALSE not found")
+	}
+
+	if !t.Next() || t.Opcode() != txscript.OP_IF {
+		return nil, fmt.Errorf("OP_IF not found")
+	}
+
+	memo := make([]byte, 0)
```
fbac, line 288: What is the maximum byte size a bitcoin message can carry?

zetaclient/chains/bitcoin/tokenizer.go
bitSmiley: It's coming from Bitcoin, it should be contained. I think we can just take whatever comes from BTC.

zetaclient/chains/bitcoin/tx_script.go
bitSmiley: From what I know, taproot script txn size is lifted, so to play safe we assume the max block size, 4MB. But usually a few hundred KB is already very expensive.

zetaclient/chains/bitcoin/tokenizer.go
bitSmiley: I would say keep it as it is for now. Ideally we should be using the bitcoin script functionalities from btcsuite. The current version in the code base does not have these functionalities. But higher versions do. In this case, we should keep the method name as of now and once versions are upgraded, just swap to the library implementation.

zetaclient/chains/bitcoin/tokenizer.go
bitSmiley: 1 is the opcode size, which is just 1 byte. `int(op)` is the opcode value, it should not be mixed with the size.

zetaclient/chains/bitcoin/tokenizer.go
bitSmiley: Yeah, that makes sense. When this was written, I was thinking it might be easier to interpret. I will update accordingly and put in comments.

bitSmiley:
@fbac I have updated based on your feedbacks. Please help take a look.

zetaclient/chains/bitcoin/tokenizer.go
fbac: Oh yes, thank you I confused it here, definitely not using `int(op)` here. In that case could you extend the comment and mention 1 refers to the 1 byte used by the opcode?

zetaclient/chains/bitcoin/tokenizer.go
fbac: Makes sense, I'm not familiar with this scripts so that's why I was wondering if there's a possibility of script being 0 bytes, on which case we can simply discard it and save resources.

zetaclient/chains/bitcoin/tokenizer.go
fbac: Makes sense, thanks!

fbac:
> @fbac I have updated based on your feedbacks. Please help take a look.

Thank you for the changes, it looks good to me now. Could you fix the two errors concatenating strings and expand the comment where we add 1? After that this can be merged!

bitSmiley:
@fbac Updated so. Funny thing, seems like my IDE added that concatenation, fix that already.

fbac (approved):
LGTM, thank you!

fbac:
Rebasing the branch with main, if we pass all the checks this will be merged.

Also @bitSmiley could you update the changelog introducing this change?

PR #2527, opened 2024-07-22
fix(`e2e`): remove hardcoded value for approval in withdrawal tests

lumtis:
# Description

Replace the hardcoded value with the amount provided in argument during the approval to allow using the test with any value.

[EDIT]: hardcoded values where ok because this is only amount for the fees that we need to give for the approval. This PR however remove the check allowing to set up a higher number for the actual withdraw

## Summary by CodeRabbit

- **New Features**
	- Streamlined withdrawal tests for ERC20 and Ether by simplifying approval logic.
  
- **Bug Fixes**
	- Removed redundant comparison checks that could lead to potential issues with withdrawal limits.

### DIFF
--- a/e2e/e2etests/test_erc20_withdraw.go
+++ b/e2e/e2etests/test_erc20_withdraw.go
@@ -15,8 +15,8 @@ func TestERC20Withdraw(r *runner.E2ERunner, args []string) {
 	withdrawalAmount, ok := new(big.Int).SetString(args[0], 10)
 	require.True(r, ok, "Invalid withdrawal amount specified for TestERC20Withdraw.")
 
-	// approve
-	tx, err := r.ETHZRC20.Approve(r.ZEVMAuth, r.ERC20ZRC20Addr, withdrawalAmount)
+	// approve 1 unit of the gas token to cover the gas fee
+	tx, err := r.ETHZRC20.Approve(r.ZEVMAuth, r.ERC20ZRC20Addr, big.NewInt(1e18))
 	require.NoError(r, err)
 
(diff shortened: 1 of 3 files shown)


swift1337 (approved):

e2e/e2etests/test_erc20_withdraw.go
```diff
@@ -12,14 +12,11 @@ import (
...
-	approvedAmount := big.NewInt(1e18)
-
 	withdrawalAmount, ok := new(big.Int).SetString(args[0], 10)
 	require.True(r, ok, "Invalid withdrawal amount specified for TestERC20Withdraw.")
-	require.Equal(r, -1, withdrawalAmount.Cmp(approvedAmount))
 
-	// approve
-	tx, err := r.ETHZRC20.Approve(r.ZEVMAuth, r.ERC20ZRC20Addr, approvedAmount)
+	// approve 1 unit of the gas token to cover the gas fee
+	tx, err := r.ETHZRC20.Approve(r.ZEVMAuth, r.ERC20ZRC20Addr, big.NewInt(1e18))
```
fbac, line 19: nit: we could create an `e2e/const.go` file which defines units, such as `const one = big.NewInt(1e18)`, so we reuse it everywhere we need it.

fbac (approved):

e2e/e2etests/test_erc20_withdraw.go
lumtis: Yeah we might have many values now for this. Created an issue: https://github.com/zeta-chain/node/issues/2540

PR #2533, opened 2024-07-23
feat: detect memo in btc txn from OP_RETURN and inscription

bitSmiley:
# Description

This is the start of a series of PRs that implements the solution mentioned in https://github.com/zeta-chain/node/issues/2478.  This is a follow up to #2524. 

This PR integrated inscription detection into the `GetBTCEvent`. The key changes are:
- Parsing the script from input witness
- Extract common methods, such as check recipient and check amount, so that less code duplication as compared to original solution outline in #2478
- Prioritising `OP_RETURN` then check for inscription. 

The "inscription" is tested in mainnet through:
- Commit: https://mempool.space/tx/213403e1efb29349a48ea9717096cf20d6e19091e496052ab591f310f0deebd6
- Reveal: https://mempool.space/tx/7a57f987a3cb605896a5909d9ef2bf7afbf0c78f21e4118b85d00d9e4cce0c2c
The input of the `7a57f987a3cb605896a5909d9ef2bf7afbf0c78f21e4118b85d00d9e4cce0c2c` is actually part of the go test.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [X] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced Bitcoin event processing with improved handling of witness data and memo extraction.
	- Introduced a script tokenizer for parsing Bitcoin script data, improving transaction processing capabilities.
	- Added functionality to decode specific Bitcoin envelope scripts related to inscriptions.

- **Bug Fixes**
	- Improved validations to ensure correct transaction amounts and recipient addresses.

- **Tests**
	- Expanded testing suite for Bitcoin event handling, including new tests for script parsing and decoding functionality.

- **Chores**
	- Added two new JSON files representing sample Bitcoin transactions for testing and validation purposes.


bitSmiley:
It seems that cross repo PR cannot be stacked on top of each other, ideally it should commit into #2524 but that branch is not found and only to `develop`. Maybe this PR will appear simplier when #2524 is merged.

lumtis:
> It seems that cross repo PR cannot be stacked on top of each other, ideally it should commit into #2524 but that branch is not found and only to `develop`. Maybe this PR will appear simplier when #2524 is merged.

Can we address this https://github.com/zeta-chain/node/pull/2524#pullrequestreview-2193268296 for the other PR? Otherwise the change looks good to me. Thanks for the contribution

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
 	}
 	return nil, nil
 }
+
+// GetBtcEventWithWitness either returns a valid BTCInboundEvent or nil.
```
lumtis, line 476: Can we move this function and below into a file like `witness.go`?

If we want to reorganized the code internally once the feature is completed, this will simplify in our case the process

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+		FromAddress: fromAddress,
+		ToAddress:   tssAddress,
+		Value:       amount,
+		MemoBytes:   memo,
+		BlockNumber: blockNumber,
+		TxHash:      tx.Txid,
+	}, nil
+}
+
+func tryExtractOpRet(tx btcjson.TxRawResult, logger zerolog.Logger) []byte {
```
lumtis, line 536: I think it would be great to document these methods, what they're doing if if private.

Can we also group the private method all at the end of the file?

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+	}
+
+	if !isValidRecipient(tx.Vout[0].ScriptPubKey.Hex, tssAddress, netParams, logger) {
+		logger.Debug().Msgf("irrelevant recipient %s for inbound: %s", tx.Vout[0].ScriptPubKey.Hex, tx.Txid)
+		return nil, nil
+	}
+
+	isAmountValid, amount := isValidAmount(tx.Vout[0].Value, depositorFee)
+	if !isAmountValid {
+		logger.Info().
```
lumtis, line 504: Why using Info instead of Debug here?

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+	var memo []byte
+	if candidate := tryExtractOpRet(tx, logger); candidate != nil {
+		memo = candidate
+		logger.Debug().Msgf("GetBtcEventWithWitness: found OP_RETURN memo %s in tx %s", hex.EncodeToString(memo), tx.Txid)
+	} else if candidate = tryExtractInscription(tx, logger); candidate != nil {
+		memo = candidate
+		logger.Debug().Msgf("GetBtcEventWithWitness: found inscription memo %s in tx %s", hex.EncodeToString(memo), tx.Txid)
+	} else {
+		return nil, nil
+	}
```
lumtis, line 518: Would be great to add comment, what we're doing in this block

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: I'm thinking if the value passed is invalid, maybe this is some critical information. Maybe a `Warn` is more appropriate. Maybe a refund is needed? `Debug` would just be easily ignored, unless there are some extra log scanning happening for Debug as well.

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+	if len(tx.Vin) == 0 { // should never happen
+		logger.Debug().Msgf("no input found for inbound: %s", tx.Txid)
+		return nil, nil
+	}
+
+	if !isValidRecipient(tx.Vout[0].ScriptPubKey.Hex, tssAddress, netParams, logger) {
+		logger.Debug().Msgf("irrelevant recipient %s for inbound: %s", tx.Vout[0].ScriptPubKey.Hex, tx.Txid)
+		return nil, nil
+	}
```
fbac, line 500: I think we could have a cleaner api on this case that avoids repeating the same logging information.

We currently logs the errors in `isValidRecipient`, then here, and then we return nil.

Instead of this approach:
- First, `isValidRecipient` should return an error instead of a bool. Modify the returned errors to include the address as well.
- Then we can proceed to modify this bit:
```suggestion
	if err := isValidRecipient(tx.Vout[0].ScriptPubKey.Hex, tssAddress, netParams, logger); err != nil {
		logger.Warn().Msgf("unexpected inbound %s: %s", tx.Txid, err)
		return nil, nil
	}
```

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+		logger.Debug().Msgf("txn %s has fewer than 2 outputs, not target OP_RETURN txn", tx.Txid)
+		return nil
+	}
+
+	vout1 := tx.Vout[1]
+	memo, found, err := bitcoin.DecodeOpReturnMemo(vout1.ScriptPubKey.Hex, tx.Txid)
+	if err != nil {
+		logger.Error().Err(err).Msgf("tryExtractOpRet: error decoding OP_RETURN memo: %s", vout1.ScriptPubKey.Hex)
+		return nil
+	}
```
fbac, line 547:
```suggestion
	memo, found, err := bitcoin.DecodeOpReturnMemo(tx.Vout[1].ScriptPubKey.Hex, tx.Txid)
	if err != nil {
		logger.Error().Err(err).Msgf("tryExtractOpRet: error decoding OP_RETURN memo: %s", tx.Vout[1].ScriptPubKey.Hex)
		return nil
	}
```

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+	lastElement, err := hex.DecodeString(witness[length-1])
+	if err != nil {
+		logger.Debug().Msgf("invalid witness element")
+		return nil
+	}
+
+	// From BIP341:
+	// If there are at least two witness elements, and the first byte of
+	// the last element is 0x50, this last element is called annex a
+	// and is removed from the witness stack.
```
fbac, line 574: nit: really curious, what's the use of annex?

zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -472,3 +472,173 @@ func GetBtcEvent(
...
+
+	var memo []byte
+	if candidate := tryExtractOpRet(tx, logger); candidate != nil {
+		memo = candidate
+		logger.Debug().Msgf("GetBtcEventWithWitness: found OP_RETURN memo %s in tx %s", hex.EncodeToString(memo), tx.Txid)
+	} else if candidate = tryExtractInscription(tx, logger); candidate != nil {
+		memo = candidate
+		logger.Debug().Msgf("GetBtcEventWithWitness: found inscription memo %s in tx %s", hex.EncodeToString(memo), tx.Txid)
+	} else {
+		return nil, nil
```
fbac, line 517: Is this case possible? If so, we should return an error. If it's not we could simply omit it.

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: I dont think it should be necessary to return an error if invalid recipient is detected. This is not an error, it's just not for zeta, pretty normal situation. I think perhaps what's worth more special treatment is when the recipient is correct, but other things go wrong, like invalid script or invalid amount. Maybe for data analytics or UX alert. Maybe.

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: From the spec, it's really reserved for future changes. I dont think it's use for now.

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: Done!

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: Updated!

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: Added!

zetaclient/chains/bitcoin/observer/inbound.go
bitSmiley: Possible. I will add an error then. But I think a log could also do it.

bitSmiley:
@lumtis @fbac I have updated accordingly, please help take a look.

lumtis (approved):

fbac:
Please do not forget that this branch has to be updated and requires resolving the conflicts.

bitSmiley:
@fbac The conflicts are resolved, should be ok now.

fbac (approved):

fbac:
Thanks @bitSmiley, LGTM. ;)

fbac:
@bitSmiley Seems like there is some cleaning up to do yet, when the PR passes all the CI we can merge it.

PR #2537, opened 2024-07-23
chore: use v18 in upgrade tests

gartnera:


lumtis (approved):

fbac (approved):
LGTM.

As a side note, the Makefile could contain an envvar that calculates the release version from tags, so this would be an automatic operation.

gartnera:
> As a side note, the Makefile could contain an envvar that calculates the release version from tags, so this would be an automatic operation.

If the release was cut from develop, `app/setup_handlers.go` will need to be updated with the new upgrade logic.

Issue #2455, opened 2024-07-09
Move restart logic from supervisor into the zetaclient directly


kingpinXD:
TSS migration needs the following restarts to happen for successful completion
- Restart when new keygen is set 
- Restart when TSS is generated 
- Restart when TSS is updated 

PResently this logic to shutdown the client is part of zeta supervisor , but can be moved to the  zetaclient binary

lumtis:
@kingpinXD can we please add some description on each issues.
- Giving some details about the goal of the issue and reasoning behind it
- Add associated labels to simplify triaging

Same for
https://github.com/zeta-chain/node/issues/2446
https://github.com/zeta-chain/node/issues/2447
https://github.com/zeta-chain/node/issues/2455

Issue #2492, opened 2024-07-16
Use retry package for queries on zetaclient


kingpinXD:
Use retry package when refactoring this logic to move into the zetaclient
https://github.com/zeta-chain/zeta-node/blob/8e1cfa7589055522045699077e2e7a11a24c6f36/cmd/zetaclientd-supervisor/lib.go#L186-L193

https://github.com/zeta-chain/zeta-node/blob/8e1cfa7589055522045699077e2e7a11a24c6f36/cmd/zetaclientd-supervisor/lib.go#L227-L234

lumtis:
@kingpinXD can we give more details on the related part that need to be refactored?

kingpinXD:
> @kingpinXD can we give more details on the related part that need to be refactored?

Yeah I will add in after the two pr;s I have are merged so that I have a clear idea of the refactor

PR #2538, opened 2024-07-24
feat: background worker routines to shutdown client for migration

kingpinXD:
# Description

This pr adds three background threads which can trigger the zeta-client to shutdown to help the process of migration
Thread 1 : Check for keygen and restart if new keygen is set 
Thread 2 : Check TSS list and restart if a new tss is added 
Thread 3 : Check current TSS , and restart if it is updated 

Closes https://github.com/zeta-chain/node/issues/2455
https://github.com/zeta-chain/node/issues/2492

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Introduced structured background processing for handling TSS updates and key generation.
	- Added a new `Logger()` method to access logging capabilities within the application context.
	- Implemented a new method `GetTSS` to retrieve TSS objects directly from the observer client.
	- Added a constant backoff strategy for retry mechanisms.
  
- **Improvements**
	- Enhanced gRPC connection security.
	- Improved error handling and logging during shutdown processes.
	- Updated restart policies in Docker configuration to ensure service resilience.

- **Bug Fixes**
	- Addressed issues with task cancellation and shutdown sequence management.


cmd/zetaclientd/start.go
```diff
@@ -210,6 +213,12 @@ func start(_ *cobra.Command, _ []string) error {
 	// Set P2P ID for telemetry
 	telemetryServer.SetP2PID(server.GetLocalPeerID())
 
+	// Create a notification context for background threads.These threads are responsible for sending shutdown signals to the main thread.
+	notifyCtx, cancel := context.WithCancelCause(ctx)
```
fbac, line 217: This could be refined so we reduce the codebase:
```suggestion
	ctxwc, cancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
```

So we can remove all the code regarding the sig channel, saving resources and logic.

And replace this:
```go
	startLogger.Info().Msgf("awaiting shutdown signals")
	select {
	case <-notifyCtx.Done():
		cause := context.Cause(notifyCtx)
		startLogger.Info().Msgf("shutdown signal received , cause : %s", cause)

	case sig := <-ch:
		startLogger.Info().Msgf("stop signal received: %s", sig)
	}

	//stop chain observers
	for _, observer := range observerMap {
		observer.Stop()
	}
	zetacoreClient.Stop()

	return nil
```

for this leaner code:
```go
	startLogger.Info().Msgf("zetaclientd is running")
	<-notifyCtx.Done()

        startLogger.Info().Msgf("zetaclientd shutting down")
	for _, observer := range observerMap {
		observer.Stop()
	}

	zetacoreClient.Stop()

	return nil
```

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
...
+				if err != nil {
+					logger.Warn().Err(err).Msg("unable to get new tss")
+					continue
+				}
+
+				if tssNew.TssPubkey == tss.TssPubkey {
+					continue
+				}
+				logger.Info().Msgf("tss address is updated from %s to %s", tss.TssPubkey, tssNew.TssPubkey)
+				tss = tssNew
```
fbac, line 49: what is the use of this line? Not sure if I'm missing something but we don't return `tss` so this line seems to not be doing anything.

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
...
+				if err != nil {
+					continue
+				}
+				tssLenUpdated := len(tssHistoricalListNew)
+
+				if tssLenUpdated <= tssLen {
+					continue
+				}
+				logger.Info().Msgf("tss list updated from %d to %d", tssLen, tssLenUpdated)
+				tssLen = tssLenUpdated
```
fbac, line 100: Same here, this line seems disposable.

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
...
+				if keygenUpdated.Status != observertypes.KeygenStatus_PendingKeygen {
+					continue
+				}
+
+				if keygen.BlockNumber == keygenUpdated.BlockNumber {
+					continue
+				}
+
+				keygen = keygenUpdated
+				logger.Info().Msgf("got new keygen at block %d", keygen.BlockNumber)
```
fbac, line 152: As well this seems disposable:
```suggestion
				logger.Info().Msgf("got new keygen at block %d", keygenUpdated.BlockNumber)
```

cmd/zetaclientd/start.go
```diff
@@ -210,6 +213,12 @@ func start(_ *cobra.Command, _ []string) error {
 	// Set P2P ID for telemetry
 	telemetryServer.SetP2PID(server.GetLocalPeerID())
 
+	// Create a notification context for background threads.These threads are responsible for sending shutdown signals to the main thread.
```
lumtis, line 216:
```suggestion
	// Create a notification context for background threads.
	// These threads are responsible for sending shutdown signals to the main thread.
```

cmd/zetaclientd/start.go
```diff
@@ -210,6 +213,12 @@ func start(_ *cobra.Command, _ []string) error {
...
+	notifyCtx, cancel := context.WithCancelCause(ctx)
+
+	// Start background threads
+	defer startBackgroundThreads(notifyCtx, cancel, zetacoreClient, masterLogger)
```
lumtis, line 220: Why using defer here?

cmd/zetaclientd/start.go
```diff
@@ -419,3 +435,37 @@ func promptPasswords() (string, string, error) {
 
 	return hotKeyPass, TSSKeyPass, err
 }
+
+// startBackgroundThreads: This function will start background threads.
+// These threads are responsible for handling TSS updates, new keygen, and new TSS key generation.
+// These threads are provided with a cancel function which is used to restart the main thread based on the outcome of the background task.
+func startBackgroundThreads(
```
lumtis, line 442: I think "routine" is a better terminology than "thread" for Go

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
+package zetacore
```
lumtis, line 1: `backgroud_threads` would be too abstract for a name
We should have filename reflecting content
- tss_update_listeners
- etc..

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"cosmossdk.io/errors"
+
+	"github.com/zeta-chain/zetacore/pkg/retry"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+	zctx "github.com/zeta-chain/zetacore/zetaclient/context"
+)
+
+// HandleTSSUpdate is a background thread that listens for TSS updates; it returns when the TSS address is updated
+func (c *Client) HandleTSSUpdate(ctx context.Context) error {
```
lumtis, line 15: Name is not reflective, it doesn't handle a TSS update, just wait for a TSS update
- ListenTSSUpdate
- WaitForTSSUpdate

zetaclient/zetacore/background_threads.go
```diff
@@ -0,0 +1,162 @@
...
+				logger.Info().Msg("HandleTSSUpdate stopped")
+				return nil
+			}
+		}
+	}
+}
+
+// HandleNewTSSKeyGeneration is a background thread that listens for new TSS key generation; it returns when a new key is generated
+// It uses the length of the TSS list to determine if a new key is generated
+func (c *Client) HandleNewTSSKeyGeneration(ctx context.Context) error {
```
lumtis, line 63: Same comment as above

lumtis (changes requested):
Would be good to have a review from @swift1337 for this PR. 🙏

cmd/zetaclientd-supervisor/main.go
```diff
@@ -69,10 +69,18 @@ func main() {
...
+		eg.Go(func() error {
+			err := cmd.Run()
+			if err != nil {
+				logger.Error().Err(err).Msg("zetaclient process exited with error")
+			} else {
+				logger.Info().Msg("zetaclient process exited")
+			}
+			cancel() // Signal other goroutines to exit
+			return err
+		})
```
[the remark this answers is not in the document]
swift1337: Also, let's not log errors twice. 
```go
if err := cmd.Run(); err != nil {
				return errors.Wrap(err, "zetaclient process exited with error")
}
```

....


```go
if err := eg.Wait(); err != nil {
    logger.Error().Err(err).Msg("error while waiting")
}
```

zetaclient/orchestrator/orchestrator.go
```diff
@@ -116,9 +116,6 @@ func (oc *Orchestrator) MonitorCore(ctx context.Context) error {
 	shutdownOrchestrator := func() {
 		// now stop orchestrator and all observers
 		close(oc.stop)
-		for _, c := range oc.observerMap {
-			c.Stop()
```
swift1337, line 120: Could you explain this change?

zetaclient/zetacore/background_threads.go
kingpinXD: The reason for adding the `handle` keyword is the fact that when the Update happens, this routine triggers a shutdown for the main thread, using the `cancel()` function of the parent context.
Happy to change it if you still feel otherwise , resolving the other similar comments for now

zetaclient/context/app.go
```diff
@@ -70,6 +70,10 @@ func (a *AppContext) Config() config.Config {
 	return a.config
 }
 
+func (a *AppContext) Logger() zerolog.Logger {
```
swift1337, line 73: Let's not expose context's logger to the outer scope.

cmd/zetaclientd/start.go
kingpinXD: The defer is to stop the background threads in case of a shutdown via os signals. I don't think its mandatory to have this , but can be good to have

zetaclient/zetacore/background_threads.go
swift1337: - Let's create a `zetaclient/maintenance` package and move TSS chores there. `zetaclient/zetacore/client.go` should be a client (i.e. req/res only), not a worker (there's 1 worker there atm which we need to move to orchestrator).
- Naming:
   - `(*TSSChore).WaitForUpdate(ctx) error` - blocking call. Returns `nil` on change
   - `(*TSSChore).WaitForNewTSSGeneration(ctx) error` - blocking call. Returns `nil` on change
   - `(*TSSChore).WaitForNewKeygen(ctx) error` - blocking call. Returns `nil` on change

swift1337 (changes requested):

zetaclient/zetacore/background_threads.go
swift1337: Also, the logic with cancel function is confusing, it's hard to grasp who actually calls cancel. I would not use `bg.WithCancel()` right now as well as this: (start.go)
```go
	// Create a notification context for background threads.These threads are responsible for sending shutdown signals to the main thread.
	notifyCtx, cancel := context.WithCancelCause(ctx)

	// Start background threads
	defer startBackgroundThreads(notifyCtx, cancel, zetacoreClient, masterLogger)
```

How I envision it:
```go
tssChores := maintenance.NewTSSChores(func(){
   // callback that is being invokes if we detect any changes in `Wait*` calls
   sigChannel <-  syscall.SIGINT
})


// invokes bg task
tssChores.Start(ctx)

```

and then we can simply implement bg opt `bg.OnComplete(callback)`








**Note**: Orchestrator is responsible for (de)provisioning observers/signers, no need to manually stop observers in start.go as it should happen within orchestrator's lifecycle.

cmd/zetaclientd/start.go
```diff
@@ -318,15 +327,16 @@ func start(_ *cobra.Command, _ []string) error {
 	}
 
 	// Orchestrator wraps the zetacore client and adds the observers and signer maps to it . This is the high level object used for CCTX interactions
-	orchestrator := orchestrator.NewOrchestrator(
+	cctxOrchestrator := orchestrator.NewOrchestrator(
```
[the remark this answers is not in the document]
swift1337: oh, I see that it haven't been merged with the latest `develop`. I fixed this

zetaclient/orchestrator/orchestrator.go
kingpinXD: the observervers are already being stopped here 
https://github.com/zeta-chain/zeta-node/blob/59515eb811ab940564b336cf08a5ac328986ebeb/cmd/zetaclientd/start.go#L356-L358

swift1337: https://github.com/zeta-chain/node/pull/2538#discussion_r1689904775

zetaclient/zetacore/background_threads.go
kingpinXD: I agree that using bg.Cancel() is not the cleanest way to do this, but the arch mentioned in this comment should be part of a separate refactor .
In my mind , background tasks is fine , as a package , We should ideally have two types of tasks , ones that read only and ones that write something back to the parent thread (naming can be decided later ), that way it can be more general , rather that adding something like `TSSChores` which is too specific .

This PR tries to keep changes to the zetaclient minimal .

cmd/zetaclientd/start.go
```diff
@@ -419,3 +435,37 @@ func promptPasswords() (string, string, error) {
...
+// These threads are responsible for handling TSS updates, new keygen, and new TSS key generation.
+// These threads are provided with a cancel function which is used to restart the main thread based on the outcome of the background task.
+func startBackgroundThreads(
+	ctx context.Context,
+	cancelFunc context.CancelCauseFunc,
+	client *zetacore.Client,
+	masterLogger zerolog.Logger,
+) context.CancelFunc {
+	backgroundContext, cancel := context.WithCancel(ctx)
+	bg.Work(
```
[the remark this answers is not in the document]
kingpinXD: it should stop the thread here 
https://github.com/zeta-chain/zeta-node/blob/2cce32c2fefc3ef7be37c797a47d3a6cd9a51905/pkg/bg/bg.go#L57-L59

I know its not the cleanest solution , and planning to implement a bigger refactor for background threads .
https://github.com/zeta-chain/node/pull/2538#discussion_r1689939726

Wil add a task once I finalize the details

zetaclient/context/app.go
kingpinXD: is there an issue you see with that , 
I am using it here https://github.com/zeta-chain/zeta-node/blob/2cce32c2fefc3ef7be37c797a47d3a6cd9a51905/zetaclient/zetacore/client_tss_migration_listners.go#L55-L55

cmd/zetaclientd/start.go
kingpinXD: I think , we would have to define a custom error , to fetch the cause in this case .
I want to use the cause to print out which job sends the shutdown request . 
What do you think ?

fbac: In the case of an error that would be printed anyway, right? as we execute the `bg.Work` routines with `WithLogger` and `WithName` the first error log we have before "zetaclientd shutting down" we can assume it's the one calling `cancel()`.

Another way to do it, more deterministic, is to modify the context that we create in StartTssMigrationRoutines and return it along with the cancelCauseFunc:
```go
	causeCtx, causeCancel := context.WithCancelCause(ctx)
	bg.Work(
		migrationRoutinesContext,
		c.HandleTSSUpdate,
		bg.WithName("HandleTSSUpdate"),
		bg.WithLogger(masterLogger),
		bg.WithCallback(causeCancel),
...

return migrationRoutinesContext, cancel
```
When the first task fails will cancel causeCtx:
- all the other tasks should receive a ctx.Done() and close.
- the `causeCtx` would carry the error and we can recover it from the main goroutine.

This would require some prior testing, as I'm just talking from memory and haven't tested it myself.

zetaclient/zetacore/background_threads.go
kingpinXD: Summarizing points discussed and additional thoughts . From what I understand, there are two main things that you are suggesting 
- Use a callback instead of the cancel function: I agree that cancel function is not the ideal option, but it provides some advantages, such as canceling all sub-routines when the parent is canceled. The cancelFunction provided here is actually the callback (renamed so that the terms make more sense ) 

- Made the functions as part of a separate struct and not `client` :  I don't see a lot of advantage of doing this , considering we are the functions that these routines use are all part of the client right now . Unless we are planning to modify functions such as `GetTss(ctx)` to `GetTss(ctx,client)`

cmd/zetaclientd/start.go
kingpinXD: I think your first point makes sense , we are printing inside the `Work` function , I can remove the cancelCause

kingpinXD:
Moving to draft for now , TSS migration test is broken on current develop . Will be fixing that first

kingpinXD:
> Moving to draft for now , TSS migration test is broken on current develop . Will be fixing that first

The pr is ready for a review

changelog.md
```diff
@@ -36,6 +36,7 @@
 * [2518](https://github.com/zeta-chain/node/pull/2518) - add support for Solana address in zetacore
 * [2483](https://github.com/zeta-chain/node/pull/2483) - add priorityFee (gasTipCap) gas to the state
 * [2567](https://github.com/zeta-chain/node/pull/2567) - add sign latency metric to zetaclient (zetaclient_sign_latency)
+* [2538](https://github.com/zeta-chain/node/pull/2538) - add background worker to shutdown zetaclientd when needed for tss migration
```
lumtis, line 39: To be moved to unreleased

zetaclient/zetacore/client_query_observer.go
```diff
@@ -167,6 +167,14 @@ func (c *Client) GetBTCTSSAddress(ctx context.Context, chainID int64) (string, e
 	return resp.Btc, nil
 }
 
+func (c *Client) GetTSS(ctx context.Context) (types.TSS, error) {
```
lumtis, line 170: Can we please add comment

lumtis, line 170: I think we have a test suite for `client_query` and we can add unit test for this function

lumtis:
Let's wait on https://github.com/zeta-chain/node/pull/2560 before considering merging this one

cmd/zetaclientd/start.go
```diff
@@ -341,14 +349,18 @@ func start(_ *cobra.Command, _ []string) error {
...
-	zetacoreClient.Stop()
+	<-notifyCtx.Done()
 
+	startLogger.Info().Msgf("initiating zetaclientd shut down")
+
+	//stop chain observers
+	for _, observer := range observerMap {
+		observer.Stop()
+	}
+	zetacoreClient.Stop()
```
swift1337, line 362: Latest `develop` doesn't have observer.Stop(), only `zetacoreClient.Stop()` https://github.com/zeta-chain/node/blob/develop/cmd/zetaclientd/start.go#L350. Please make sure this is up-to date with develop

zetaclient/zetacore/background_threads.go
swift1337: Several points from my side:
- By implementing reusable code components, we can simplify the mental model behind the project.
- In order to keep the codebase maintainable, we need to reduce the technical debt or at least stop increasing it. In my point of view, adding net-new worker logic in the client's package (a package that should handle req/res only) is indeed a technical debt increase due to making a "shortcut."
- `bg` package can and should be improved in the future PRs, but for this PR,`bg.OnSuccess(func())` is more than enough (IMO)

swift1337: I implemented proposed changes, feel free to ask any questions:
- https://github.com/zeta-chain/node/pull/2538/commits/f993da5aa5db578556741711bbab7f9ae8b732b7
- https://github.com/zeta-chain/node/pull/2538/commits/7c113a5059da6ce9e6e7a87dc4c1fcecaeacbfe8

TLDR:
```go
// cmd/zetaclient/start.go
func start(_ *cobra.Command, _ []string) error {
  // ...
  
  // Creating a channel to listen for os signals (or other signals)
  signalChannel := make(chan os.Signal, 1)
  signal.Notify(signalChannel, syscall.SIGINT, syscall.SIGTERM)
  
  // Maintenance workers ============
  maintenance.NewTSSListener(zetacoreClient, masterLogger).Listen(ctx, func() {
      masterLogger.Info().Msg("TSS listener received an action to shutdown zetaclientd.")
      signalChannel <- syscall.SIGTERM
  })
  
  sig := <-signalChannel
  startLogger.Info().Msgf("Stop signal received: %q", sig)
  
  zetacoreClient.Stop()
}
```

cc @lumtis

swift1337 (approved):
LGTM

lumtis (approved):
Code organization and implementation looks good to me, I guess this can be tested E2E in the TSS migration test?

kingpinXD:
> Code organization and implementation looks good to me, I guess this can be tested E2E in the TSS migration test?

Yes , the TSS migration test would use these routines

Issue #2555, opened 2024-07-25
`zetacored` : tss migration migrators check using consensus type


kingpinXD:
Currently we check for the length of migrators list against the number of supported chains .
This is correct , as every chain should have its own migratory before we update the TSS address .

However presently, we have Solana added as a chain but the Solana cctx lifecycle is still WIP.

It would make sense to refactor the original check and make it based on consensus types `ETH` and  `Bitcoin`

```
	supportedChains := append(k.zetaObserverKeeper.GetSupportedChainsByConsensus(ctx, chains.Consensus_ethereum),
		k.zetaObserverKeeper.GetSupportedChainsByConsensus(ctx, chains.Consensus_bitcoin)...)
	// Each connected chain should have its own tss migrator
	if len(supportedChains) != len(tssMigrators) {
		return nil, errorsmod.Wrap(
			types.ErrUnableToUpdateTss,
			"cannot update tss address incorrect number of migrations have been created and completed",
		)
	}
```

PR #2556, opened 2024-07-25
refactor: migrator length check to use consensus type

kingpinXD:
# Description

Closes https://github.com/zeta-chain/node/issues/2555

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Introduced methods to retrieve supported foreign chains based on specific consensus types, enhancing the chain management functionality.
	- Added a functionality to get a list of chains that support migration, improving migration-related processes.
  
- **Bug Fixes**
	- Refined test logic to ensure accurate behavior when retrieving chains relevant to migration.

- **Tests**
	- Expanded test coverage for the `Keeper` functionality regarding foreign chains and migration, ensuring robust performance across consensus types.


ws4charlie (approved):

x/crosschain/keeper/msg_server_update_tss.go
```diff
@@ -70,3 +72,9 @@ func (k msgServer) UpdateTssAddress(
 
 	return &types.MsgUpdateTssAddressResponse{}, nil
 }
+
+// GetChainsSupportingMigration returns the chains that support migration.
```
lumtis, line 76: Can we leave more details in the comments? Why specifically filtering ethereum and bitcoin?

x/crosschain/keeper/msg_server_update_tss.go
```diff
@@ -70,3 +72,9 @@ func (k msgServer) UpdateTssAddress(
...
+func (k *Keeper) GetChainsSupportingMigration(ctx sdk.Context) []chains.Chain {
```
lumtis, line 77: We should add another check:
chain.CCTXGateway == observer

Chains not using the observer solution don't need TSS migration

lumtis, line 77:
```suggestion
func (k *Keeper) GetChainsSupportingTSSMigration(ctx sdk.Context) []chains.Chain {
```
We should explicitely mention TSS migration everywhere as there can be any kind of migration

x/crosschain/keeper/msg_server_update_tss.go
```diff
@@ -70,3 +72,9 @@ func (k msgServer) UpdateTssAddress(
...
+	return append(k.zetaObserverKeeper.GetSupportedForeignChainsByConsensus(ctx, chains.Consensus_ethereum),
```
lumtis, line 78: I would prefer a functional programming approach.

We implement a stateless function `FilterChainsSupportingTSSMigration(chainList) -> chainList`

GetChainsSupportingTSSMigration can call `GetSupportedForeignChains` and then apply this method to it.

Unit tests are simpler to develop and maintain.

lumtis (changes requested):

x/crosschain/keeper/msg_server_update_tss.go
swift1337: And use can also combine these like so:
```go
type ChainFilter func (c Chain) bool

func SupportsTSS(c Chain) bool {
 // ...
}

func (Keeper) GetSupportedForeignChains(filters ...ChainFilter) []Chain {
   // ...
}

// ....

chainsWithTSSMIgration := keeper.GetSupportedForeignChains(SupportsTSS)

```

x/crosschain/keeper/msg_server_update_tss.go
kingpinXD: I am not sure what details to add, bitcoin and Ethereum, consensus  are the only ones that support TSS migration right now

x/crosschain/keeper/msg_server_update_tss.go
kingpinXD: Thanks, both good ideas. Ill implement them

x/crosschain/keeper/msg_server_update_tss.go
kingpinXD: e32c2a87997922c99ad8a9e7709df90d10050b69

x/crosschain/keeper/msg_server_update_tss.go
kingpinXD: e32c2a87997922c99ad8a9e7709df90d10050b69

x/crosschain/keeper/msg_server_update_tss.go
kingpinXD: e32c2a87997922c99ad8a9e7709df90d10050b69

kingpinXD: e32c2a87997922c99ad8a9e7709df90d10050b69

pkg/chains/chain_filters.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		chainList = filteredChains
+	}
+
+	// Return the filtered list of chains
+	return chainList
+}
+
+// CombineFilterChains combines multiple lists of chains into a single list
+func CombineFilterChains(chainLists ...[]Chain) []Chain {
```
swift1337, line 45: Why do we need to dedup chains here?

pkg/chains/chain_filters.go
```diff
@@ -0,0 +1,60 @@
...
+// FilterConsensusBitcoin filters chains that have the bitcoin consensus
+func FilterConsensusBitcoin(c Chain) bool { return c.Consensus == Consensus_bitcoin }
+
+// FilterConsensusSolana filters chains that have the solana consensus
+func FilterConsensusSolana(c Chain) bool { return c.Consensus == Consensus_solana_consensus }
+
+// FilterChains applies a list of filters to a list of chains
+func FilterChains(chainList []Chain, filters ...ChainFilter) []Chain {
+	// Apply each filter to the list of supported chains
+	for _, filter := range filters {
```
swift1337, line 30: If we have 5 chains and 3 filters that match ALL chains, doesn't this mean that we'll have 15 chains with duplicates?

x/crosschain/keeper/msg_server_update_tss_test.go
```diff
@@ -329,3 +330,26 @@ func TestMsgServer_UpdateTssAddress(t *testing.T) {
 		require.Equal(t, len(k.GetObserverKeeper().GetSupportedChains(ctx)), len(migrators))
 	})
 }
+
+func TestKeeper_GetChainsSupportingTSSMigration(t *testing.T) {
+	t.Run("should return only ethereum and bitcoin chains", func(t *testing.T) {
```
lumtis, line 335: We could maybe use TDT for this specific one?

pkg/chains/chain_filters.go
swift1337: What do you think about this?
```go
func FilterChains(items []Chain, filters ...ChainFilter) []Chain {
	// noop
	if len(filters) == 0 {
		return items
	}

	var result []Chain
	for _, item := range items {
	TOP:
		for _, filter := range filters {
			if !filter(item) {
				continue TOP
			}
		}

		result = append(result, item)
	}

	return result
}
```

kingpinXD: What do you mean ?
`chainList = filteredChains` we are not appending to the returned list . Does this clarify your question

pkg/chains/chain_filters.go
kingpinXD: why would we need duplicates ?
I don't see any harm in deduplicating

pkg/chains/chain_filters.go
kingpinXD: added a test case with duplicate filters

x/crosschain/keeper/msg_server_update_tss_test.go
kingpinXD: I think this test covers all the scenarios already , and we can add more composite filters to that as and when needed
https://github.com/zeta-chain/zeta-node/blob/54d32d9cb2f4cf9184950b01bc1fb38bc421fde8/pkg/chains/chain_filters_test.go#L145-L150

lumtis (approved):

pkg/chains/chain_filters.go
swift1337: Oh, okay, my bad. Still initial implementation would iterate though chains multiple times and do up to `len(chains) * len(filters)` allocations in the worst case. In proposed implementation it would do only up to `len(chains)`

pkg/chains/chain_filters.go
swift1337: Naming got me confused 😅 We can implement this as a generic `slices.Unique([]E)` under `pkg/slices`. Otherwise please name smth like `UniqueChains(list []Chain) []Chain`

pkg/chains/chain_filters.go
```diff
@@ -0,0 +1,60 @@
...
+	return c.IsExternal
+}
+
+// FilterGatewayObserver filters chains that are gateway observers
+func FilterGatewayObserver(c Chain) bool {
+	return c.CctxGateway == CCTXGateway_observers
+}
+
+// FilterConsensusEthereum filters chains that have the ethereum consensus
+func FilterConsensusEthereum(c Chain) bool {
```
swift1337, line 17: There's also a more generic way
```go
func FilterByConsensus(cs Consensus) ChainFilter {
  return func(chain Chain) bool { return chain.Consensus == cs }
}
```

pkg/chains/chain_filters.go
kingpinXD: The length of chains is fixed , I wouldn't worry too much about the number of iterations and prioritize readabilty .
IMO using labels makes it less readable

swift1337 (approved):
Approved. Left some non-blocking comments.

pkg/chains/chain_filters.go
kingpinXD: Its combines the output of multiple filters , and the naming is okay for now since its in the chain_filters file . Happy to implement a generic function though , I think that makes sense , ill add a issue for it .

pkg/chains/chain_filters.go
kingpinXD: 88a47ea2b1105249ae793de57c72b69fcd5dabe7

PR #2583, opened 2024-07-29
fix(ci): propagate timeout-minutes

gartnera:
`timeout-minutes` was set in the matrix, but not actually injected into the job. This resulted in everything being run with the default 25 minute timeout. The mainnet imports need more time.

## Summary by CodeRabbit

- **Chores**
	- Enhanced configuration for job execution time in the GitHub Actions workflow, allowing for customizable timeout settings.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -168,6 +168,7 @@ jobs:
       make-target: ${{ matrix.make-target }}
       runs-on: ${{ matrix.runs-on}}
       run: ${{ matrix.run }}
+      timeout-minutes: "${{ matrix.timeout-minutes || 25 }}"
     secrets: inherit
   # this allows you to set a required status check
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -168,6 +168,7 @@ jobs:
       make-target: ${{ matrix.make-target }}
       runs-on: ${{ matrix.runs-on}}
       run: ${{ matrix.run }}
+      timeout-minutes: "${{ matrix.timeout-minutes || 25 }}"
```
gartnera, line 171: We have to set the default value on this side too since setting an empty string will overwrite the default value set in resusable-e2e.yml

lumtis (approved):

CharlieMc0 (approved):

julianrubino (approved):

Issue #2493, opened 2024-07-16
Use a declarative design to tell the ZetaClient how to observe external chains


lumtis:
**Describe the Issue**

There is already an issue mention this idea https://github.com/zeta-chain/node/issues/2318
But I'm create a dedicated issue as this might be a larger refactoring out of the scope of the issue above.

This issue can also be focused for EVM chains at first.

The way the observer observes the different types of event for inbound/outbound/outbound reverts should be pretty generic for all.

The idea of this issue is to separate the "what" from the "how", and keeping a list of events to observe in a declarative way. 

This issue would simplify adding new events for example in the context of supporting the new smart contract design.

The issue is specific for observation, but a similar design could be thought for the signing part. Separating what to sign and how to sign

**Considered Outcome**

Keeping on the observer package a map/list of the event with cctx type associated (to initiate the inbound/outbound vote), when then have a method to observer the event.

This example is specifically for the new smart contract design.

```

var map eventMap {
  "deposit": [DepositCCTXData], // allow to initiate the inbound vote for a deposit to ZetaChain
  "depositAndCall": ...,
  "call": ...,
}

...

func Observer(evmClient, eventMap) {
  // actual logic parsing the events
}

```

swift1337:
I'll take this into account when revamping the orchestrator.

Based on my experience, this is not particularly true.
> The way the observer observes the different types of events for inbound/outbound/outbound reverts should be pretty generic for all.

Different chains have different flows.
- We can't broadcast more than 1 BTC outbound.
- We can send as many outbound as we want for EVM.
- TON mixes inbounds with outbound due to how tx pagination is implemented in this chain.
- Eth allows us to replace tx.
- and so on ...

I think that we'd still stick with a custom per-chain approach, but (!) all common operations should go through the same pipeline, i.e. we need to figure out the common denominator and keep chain-specific implementations as thin as possible. Ideally, each chain observer&signer should work with cctx and have a certain "outcome" that would determine whether we should process the next cctx, do a retry, etc...


I also would push back on having an intermediate "inbound event", "outbound event" ([link](https://github.com/zeta-chain/node/blob/b516a46959a730bb1dd04dd6a8c0b8b6e05714c4/zetaclient/types/event.go#L10)) because it introduces unnecessary intermediate abstraction that IMO should be directly a cctx (also need a revamp).

lumtis:
Would not be a universal object used for all chains but one per specific chain type (in particular EVM)

Example:
[...]

PR #2616, opened 2024-08-01
feat: implement gas token deposit with protocol contract v2

lumtis:
Added:
- Add a `ProtocolContractVersion` field in CCTX to store which workflow should be used
- Add a new observe function in EVM observer to parse Gateway Deposit events
- Add a new function in fungible to process deposits marked V2

Note:
To prevent interfering too much with the rest of the development, most new additions are implement in `v2` prefixed files and in functions that mention V2. This will simplify conflicts resolution and also allow to simplify work to consolidate all the functionalities into the new design.
The V2 terminology will be removed in the future: https://github.com/zeta-chain/node/issues/2627

To test:
```
make start-v2-test
```


x/crosschain/client/cli/tx_vote_inbound.go
```diff
@@ -87,3 +93,16 @@ func CmdVoteInbound() *cobra.Command {
 
 	return cmd
 }
+
+func parseProtocolContractVersion(version string) (types.ProtocolContractVersion, error) {
+	switch version {
+	case "V1":
+		return types.ProtocolContractVersion_V1, nil
+	case "V2":
+		return types.ProtocolContractVersion_V2, nil
```
fbac, line 102: nit: Technically golang convention for constants is snake case without any _.

lumtis: This is code generated by proto when using `enum`

lumtis:
> looks good, just changelog and e2e tests checks failing how are gateway and other contracts deployed atm?

Used to pass, I think this is a isolated error, we'll try to check why v2 are not running in the CI as well

zetaclient/chains/evm/observer/v2_inbound.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/zeta-chain/zetacore/zetaclient/chains/evm"
+	"github.com/zeta-chain/zetacore/zetaclient/compliance"
+	"github.com/zeta-chain/zetacore/zetaclient/config"
+	"github.com/zeta-chain/zetacore/zetaclient/metrics"
+	"github.com/zeta-chain/zetacore/zetaclient/zetacore"
+)
+
+// ObserveGateway queries the gateway contract for deposit/call events
+// returns the last block successfully scanned
+func (ob *Observer) ObserveGateway(ctx context.Context, startBlock, toBlock uint64) uint64 {
```
swift1337, line 26: I think this method should return `(uint64, error)` and use **sentinel** errors inside. 

Callers of `ob.ObserverGateway()` should handle those errors. Ideally we also want to have a return of some kind of `outcome` enum. i.e. `noop`, `processed`, `processedPartially`, `...` i.e. `(uint64, outcome, error)`

lumtis: I put the error handling at a higher level as cleaner: [63c4698](https://github.com/zeta-chain/node/pull/2616/commits/63c4698c7cad0beafb23035692782cf52c590fcf)

For the other changes containing `outcome`, I agree but I think it can be part of a separate PR for general refactoring of the observation logic. I put a TODO for https://github.com/zeta-chain/node/issues/2493 in the comments

e2e/runner/logger.go
```diff
@@ -199,6 +201,29 @@ func (l *Logger) ZRC20Withdrawal(
 	}
 }
 
+// GatewayDeposit prints a GatewayDeposit event
+func (l *Logger) GatewayDeposit(
+	contract interface {
```
swift1337, line 206:
```suggestion
	contract DepositParser
```
```go
type DepositParser interface {...}
```

swift1337 (approved):

Issue #2671, opened 2024-08-09
`zetacore` : fix events for add observer


kingpinXD:
Message AddObserver currently emits events when a new observer is not added, this should be refactored to make sure this event is only emitted when a new observer gets added to the set

kingpinXD:
In my opinon its just better to refactor this message and break it down into two different messages 
- AddSignerNodeAccount
- AddObserverToObserver 

The `AddNodeAccountOnly` flag is just extra logic we dont need

Issue #2670, opened 2024-08-09
`zetacore` : check for duplicate addresses when adding or updating the observer set


kingpinXD:
![image](https://github.com/user-attachments/assets/b5a0bdd1-8c72-4e94-aede-af3b7d10e7e5)

PR #2672, opened 2024-08-09
fix: check for duplicates in observer set

kingpinXD:
# Description
- The pr adds a check for duplicates in the observer.Validate() function.
- The validate function is called before setting the observer set during an update and add .
- The validate is not needed to call on removal, as removing a value from a valid set cannot make it invalid . 
Closes https://github.com/zeta-chain/security/issues/85
https://github.com/zeta-chain/node/issues/2670
https://github.com/zeta-chain/node/issues/2671

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for the authorization process, improving security against unauthorized access.
	- Added checks to prevent duplicate entries in the observer set, ensuring data integrity.
	- Introduced new error types for improved error handling in observer management.

- **Bug Fixes**
	- Improved error handling in observer addition and updating functions to provide clearer feedback on failures.

- **Tests**
	- Expanded test coverage, including new cases for validating edge scenarios related to observer management.

- **Documentation**
	- Updated changelog to reflect recent changes and improvements in functionality.


kingpinXD:
The observer set is also set during genesis , but I think we can assume the genesis state to be valid , as Validate should already have been run on the genesisState

x/observer/keeper/observer_set.go
```diff
@@ -36,23 +37,26 @@ func (k Keeper) IsAddressPartOfObserverSet(ctx sdk.Context, address string) bool
 	return false
 }
 
-func (k Keeper) AddObserverToSet(ctx sdk.Context, address string) {
+// AddObserverToSet adds an observer to the observer set.It makes sure the updated observer set is valid.
+func (k Keeper) AddObserverToSet(ctx sdk.Context, address string) error {
 	observerSet, found := k.GetObserverSet(ctx)
-	if !found {
-		k.SetObserverSet(ctx, types.ObserverSet{
+	switch {
```
lumtis, line 43: Why using a switch for a single condition? if is more adapted for this case

x/observer/keeper/observer_set.go
```diff
@@ -36,23 +37,26 @@ func (k Keeper) IsAddressPartOfObserverSet(ctx sdk.Context, address string) bool
...
-	}
-	for _, addr := range observerSet.ObserverList {
-		if addr == address {
-			return
 		}
+	default:
+		observerSet.ObserverList = append(observerSet.ObserverList, address)
+	}
+	err := observerSet.Validate()
+	if err != nil {
```
lumtis, line 52:
```suggestion
	
	if err := observerSet.Validate(); err != nil {
```

x/observer/keeper/observer_set.go
```diff
@@ -36,23 +37,26 @@ func (k Keeper) IsAddressPartOfObserverSet(ctx sdk.Context, address string) bool
...
-		})
-		return
-	}
-	for _, addr := range observerSet.ObserverList {
-		if addr == address {
-			return
 		}
+	default:
+		observerSet.ObserverList = append(observerSet.ObserverList, address)
+	}
```
lumtis, line 50: I think it's more clear to do
```
if !found {
  observerSet = types.ObserverSet{}
}
append(...)

```

x/observer/keeper/msg_server_add_observer.go
```diff
@@ -52,10 +52,15 @@ func (k msgServer) AddObserver(
 		return &types.MsgAddObserverResponse{}, nil
 	}
 
-	k.AddObserverToSet(ctx, msg.ObserverAddress)
+	// Add observer to the observer set
+	err = k.AddObserverToSet(ctx, msg.ObserverAddress)
+	if err != nil {
+		return &types.MsgAddObserverResponse{}, err
+	}
 	observerSet, _ := k.GetObserverSet(ctx)
```
lumtis, line 60: We should think this, if not found we should explicitely set ObserverSet empty
```
if !found {
  observerSet = types.ObserverSet{}
}
```

x/observer/keeper/msg_server_add_observer.go
```diff
@@ -52,10 +52,15 @@ func (k msgServer) AddObserver(
...
 
 	k.SetLastObserverCount(ctx, &types.LastObserverCount{Count: observerSet.LenUint()})
+
 	EmitEventAddObserver(
```
lumtis, line 64: This looks these could be done in `AddObserverToSet`

x/observer/keeper/observer_set.go
kingpinXD: I would need to use a `if` with an `else` as I think it is a bit cleaner to append only when the set is found.

However, we should call Validate() either way. IMO it looks slightly more readable using the switch

x/observer/keeper/msg_server_add_observer.go
kingpinXD: Not sure what you mean, if not found we set it it with the address provided 
https://github.com/zeta-chain/zeta-node/blob/d4048a53b75a178d8519057443d1c85afddc08ba/x/observer/keeper/observer_set.go#L44-L47

Do you mean , for the case !found , and address is not valid ?

x/observer/keeper/observer_set.go
swift1337:
```go
func (k Keeper) AddObserverToSet(ctx sdk.Context, address string) error {
	observerSet, found := k.GetObserverSet(ctx)
	if !found {
		observerSet = types.ObserverSet{
			ObserverList: []string{},
		}
	}

	observerSet.ObserverList = append(observerSet.ObserverList, address)
	
	if err := observerSet.Validate(); err != nil {
		return err
	}

	k.SetObserverSet(ctx, observerSet)
	
	return nil
}
```

swift1337 (approved):
Approved with minor comments

x/observer/keeper/msg_server_update_observer_test.go
```diff
@@ -73,6 +73,61 @@ func TestMsgServer_UpdateObserver(t *testing.T) {
...
+			//ACT
+			_, err = srv.UpdateObserver(sdk.WrapSDKContext(ctx), &types.MsgUpdateObserver{
+				Creator:            accAddressOfValidator.String(),
+				OldObserverAddress: accAddressOfValidator.String(),
+				NewObserverAddress: newOperatorAddress.String(),
+				UpdateReason:       types.ObserverUpdateReason_Tombstoned,
+			})
+
+			// ASSERT
+			require.ErrorContains(t, err, types.ErrDuplicateObserver.Error())
```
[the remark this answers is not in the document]
kingpinXD: The error is wrapped like so 
`return errors.Wrap(types.ErrUpdateObserver, err.Error())`

x/observer/types/observer_set.go
```diff
@@ -14,14 +15,25 @@ func (m *ObserverSet) LenUint() uint64 {
 	return uint64(len(m.ObserverList))
 }
 
-// Validate observer mapper contains an existing chain
+// Validate observer set verifies that the observer set is valid
+// - All observer addresses are valid
+// - No duplicate observer addresses
 func (m *ObserverSet) Validate() error {
```
fbac, line 21: Validate() can be refactores to achieve O(n) complexity.
Also it would be nice to add `ErrInvalidObserverAddress` as a custom error.
```suggestion
func (m *ObserverSet) Validate() error {
	observers := make(map[string]struct{})

	for _, observerAddress := range m.ObserverList {
		// Check address is valid.
		_, err := sdk.AccAddressFromBech32(observerAddress)
		if err != nil {
			return errors.Wrapf(ErrInvalidObserverAddress, "invalid observer address: %s", observerAddress)
		}
		
		// Check for duplicate observers.
		if _, ok := observers[observerAddress]; ok {
			return errors.Wrapf(ErrDuplicateObserver, "observer %s", observerAddress)
		}

		// Update observers map.
		observers[observerAddress] = struct{}{}
	}

	return nil
}
```

x/observer/keeper/observer_set.go
```diff
@@ -67,17 +72,28 @@ func (k Keeper) RemoveObserverFromSet(ctx sdk.Context, address string) {
...
 }
 
+// UpdateObserverAddress updates an observer address in the observer set.It makes sure the updated observer set is valid.
 func (k Keeper) UpdateObserverAddress(ctx sdk.Context, oldObserverAddress, newObserverAddress string) error {
 	observerSet, found := k.GetObserverSet(ctx)
 	if !found {
 		return types.ErrObserverSetNotFound
 	}
+	found = false
 	for i, addr := range observerSet.ObserverList {
```
fbac, line 82: nit: why ObserverList is a slice instead of a map?
With a map we would avoid having duplicates by default and would read/write from it with O(1).

x/observer/keeper/msg_server_add_observer.go
fbac: I agree with @lumtis but I lack some context:
- If the only way to add an observer is through `AddObserverToSet`, then emitting the event should be handled inside AddObserver, as the logic is intrinsic to it.
- If there are more ways of adding observers then we should refactor the event to be an optional callback to be passed to the N functions that add observers, such as `func SetObserver(foo, bar, eventToBeEmitted)`

The key thing here imho is to rely less and less with domain logic (If I invoke X, I have immediately to invoke Y).

kingpinXD: I agree with both of your points, however, I feel that emitting events in the main message is a pattern that we have been following across all the messages. And I would like to keep it the same for this message as well.

I did add the refactor the usage for `SetLastObserverCount` as it makes the the main message look cleaner 
83aed121f5b8beaaf55676da29d8b67f15f70c2f

x/observer/keeper/observer_set.go
kingpinXD: Maps are not ordered in go, which might lead to complexities. 
We would need to sort everytime before saving, I think using a list is better

fbac: Gotcha, I didn't know we need them ordered. Thanks!

lumtis (approved):

x/observer/keeper/msg_server_add_observer.go
lumtis: My point is if `found == false`, `observerSet` should be deemed as undefined and not used

PR #2674, opened 2024-08-09
fix: modify message vote tss to make sure the correct ballot is finalized for a keygen

kingpinXD:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/x/observer/keeper/msg_server_vote_tss.go
+++ b/x/observer/keeper/msg_server_vote_tss.go
@@ -43,9 +43,9 @@ func (k msgServer) VoteTSS(goCtx context.Context, msg *types.MsgVoteTSS) (*types
 	}
 
 	// Use a separate transaction to update keygen status to pending when trying to change the TSS address.
-	if keygen.Status == types.KeygenStatus_KeyGenSuccess {
-		return &types.MsgVoteTSSResponse{}, errorsmod.Wrap(types.ErrKeygenCompleted, voteTSSid)
-	}
+	//if keygen.Status == types.KeygenStatus_KeyGenSuccess {
+	//	return &types.MsgVoteTSSResponse{}, errorsmod.Wrap(types.ErrKeygenCompleted, voteTSSid)
+	//}
 
 	// GetBallot checks against the supported chains list before querying for Ballot.
 	ballotCreated := false
@@ -92,6 +92,14 @@ func (k msgServer) VoteTSS(goCtx context.Context, msg *types.MsgVoteTSS) (*types
 			BallotCreated: ballotCreated,
 			KeygenSuccess: false,
 		}, nil
+	}
+
+	if keygen.Status != types.KeygenStatus_PendingKeygen {
+		return &types.MsgVoteTSSResponse{}, nil
+	}
+
+	if msg.KeygenZetaHeight != keygen.BlockNumber {
+		return &types.MsgVoteTSSResponse{}, nil
 	}
 
 	// Set TSS only on success, set keygen either way.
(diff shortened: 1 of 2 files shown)


x/observer/keeper/msg_server_vote_tss.go
```diff
@@ -42,11 +42,6 @@ func (k msgServer) VoteTSS(goCtx context.Context, msg *types.MsgVoteTSS) (*types
 		return &types.MsgVoteTSSResponse{}, errorsmod.Wrap(types.ErrKeygenNotFound, voteTSSid)
 	}
 
-	// Use a separate transaction to update keygen status to pending when trying to change the TSS address.
-	if keygen.Status == types.KeygenStatus_KeyGenSuccess {
-		return &types.MsgVoteTSSResponse{}, errorsmod.Wrap(types.ErrKeygenCompleted, voteTSSid)
-	}
-
```
fbac, line 49: nit: Can you clarify why this is not needed anymore? Lacking context so I'd like to get this straight. :)

x/observer/keeper/msg_server_vote_tss.go
```diff
@@ -94,6 +89,21 @@ func (k msgServer) VoteTSS(goCtx context.Context, msg *types.MsgVoteTSS) (*types
...
+	// 1. The keygen is still pending
+	// 2. The keygen block number matches the ballot block number ,which makes sure this the correct ballot for the current keygen
+
+	// Return without error so the vote is added to the ballot
+	if keygen.Status != types.KeygenStatus_PendingKeygen {
+		return &types.MsgVoteTSSResponse{}, nil
+	}
+
+	// For cases when an observer tries to vote for an older pending ballot , associated with a keygen that was discarded , we would return at this check while still adding the vote to the ballot
+	if msg.KeygenZetaHeight != keygen.BlockNumber {
```
fbac, line 103: Should this be `msg.KeygenZetaHeight > keygen.BlockNumber`?

kingpinXD: The check is added so that we can add votes to an older ballot. 
When that happens the `msg.KeygenZetaHeight ` would always be less than the `keygen.BlockNumber` .

We could modify this to `msg.KeygenZetaHeight < keygen.BlockNumber` but I think using `!=` is fine as well

fbac (approved):

x/observer/keeper/msg_server_vote_tss.go
kingpinXD: We are removing this check , so as to allow observers to vote for ballots that have happened in the past and had to be discarded for some reason.
THe newly added checks below , would prevent such a vote from modifying the keygen status

x/observer/keeper/msg_server_vote_tss.go
```diff
@@ -94,6 +89,21 @@ func (k msgServer) VoteTSS(goCtx context.Context, msg *types.MsgVoteTSS) (*types
...
+		return &types.MsgVoteTSSResponse{}, nil
```
lumtis, line 104: We should add status values in the response
```
			VoteFinalized: isFinalized,
			BallotCreated: ballotCreated,
			KeygenSuccess: false,
```

lumtis (changes requested):

x/observer/keeper/msg_server_vote_tss.go
kingpinXD: I was actually considering removing the status object , but did not want to change existing code 
Wondering why do we need this status at all ?

lumtis: For assertions in unit tests

lumtis (approved):

PR #2703, opened 2024-08-14
feat: add e2e test stateful contract

fbac:
# Description

Include the first test stateful contract.

- [x] Add first contract
- [x] Add e2e testing
- [x] Opened https://github.com/zeta-chain/node/issues/2710 to handle future work around events.
- [x] Add unit testing

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [x] Go integration tests


precompiles/types/errors.go
```diff
@@ -0,0 +1,17 @@
...
+	// ErrDifferentOrigin is raised when an approval is set but the origin address is not the same as the spender.
+	ErrDifferentOrigin = "tx origin address %s does not match the delegator address %s"
+	// ErrInvalidAddr is raised when the origin address is invalid
+	ErrInvalidAddr = "invalid addr: %s"
+	// ErrInvalidDelegator is raised when the delegator address is not valid.
+	ErrInvalidDelegator = "invalid delegator address: %s"
+	// ErrInvalidNumberOfArgs is raised when the number of arguments is not what is expected.
+	ErrInvalidNumberOfArgs = "invalid number of arguments; expected %d; got: %d"
+	// ErrInvalidMethod is raised when the method is invalid.
+	ErrInvalidMethod = "unknown method: %s"
```
swift1337, line 16: Is it copied directly from Evmos repo? Doesn't look like idiomatic go 😓

precompiles/regular/regular.go
```diff
@@ -0,0 +1,248 @@
...
+	gasRequiredByMethod = map[[4]byte]uint64{}
+	if err := abi.UnmarshalJSON([]byte(RegularModuleMetaData.ABI)); err != nil {
+		panic(err)
+	}
+	for methodName := range abi.Methods {
+		var methodID [4]byte
+		copy(methodID[:], abi.Methods[methodName].ID[:4])
+		switch methodName {
+		case RegularCallMethodName:
+			gasRequiredByMethod[methodID] = 10
```
swift1337, line 55: can you explain `10` here? Also consider moving this to const

swift1337 (approved):

precompiles/types/errors.go
fbac: Yes, some parts are a mix between EVMOS and Cronos. I'll be modifying those to adhere to our standards.

precompiles/regular/regular.go
fbac: It's just the gas used in this specific method. Godo idea to move it to `RegularCallMethodGasRequired`.

precompiles/regular/IRegular.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.7;
```
lumtis, line 2: NIT: we now use `pragma solidity 0.8.26;` in protocol contract. Maybe we can do the same here?

precompiles/regular/testutil/RegularCaller.go
```diff
@@ -0,0 +1,286 @@
+// Code generated - DO NOT EDIT.
```
lumtis, line 1: we should look into excluding it from code coverage

Right now it seems it is included as there are the codecov comments

precompiles/regular/regular.go
```diff
@@ -0,0 +1,262 @@
+package regular
```
lumtis, line 1: Just for my understanding on this file.
Has it been fully written from scratch, written from a template, or generated then customized?

We should have a better unit test coverage on it (though we might make exception for this one as a test contract) then we should find some way to exclude test coverage from the generated/generic part but not the customized part

precompiles/regular/regular.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+	callingContract, ok := args[1].(common.Address)
+	if !ok {
+		return nil, ptypes.ErrInvalidAddr{
+			Got: callingContract.String(),
+		}
+	}
+
+	res, err := rc.FungibleKeeper.CallEVM(
```
lumtis, line 209: Even if it is just for test purpose, I would consider this example a bit dangerous,  the fungible module can mint ZRC20, meaning this precompiled could be used to mint infinite amount of ZRC20s

Even disable I would avoid having it defined in our codebase 😄

I would change it to a more scope and simple use case:

One thing I have in mind is interacting with the gas stability pool. In the fungible module `gas_stability_pool.go` provides pools autofilled from cctx fees that are used to fund cctx when gas price hikes. Eventually they can be manually funded by users.
You have a method to get the address of the gas stability pool from a chain ID:
```
func (k Keeper) GetGasStabilityPoolBalance(
	ctx sdk.Context,
	chainID int64,
)
```
We could have a precompile allow to get this address from a smart contract call.

e2e/e2etests/test_precompiles_regular.go
```diff
@@ -0,0 +1,29 @@
...
+	require.Equal(r, "zeta1h8duy2dltz9xz0qqhm5wvcnj02upy887fyn43u", res, "Failed to validate Bech32ify result")
+
+	addr, err := caller.Bech32ToHexAddr(nil, "zeta1h8duy2dltz9xz0qqhm5wvcnj02upy887fyn43u")
+	require.NoError(r, err, "Error calling Bech32ToHexAddr")
+	require.Equal(
+		r,
+		"0xB9Dbc229Bf588A613C00BEE8e662727AB8121cfE",
+		addr.String(),
+		"Failed to validate Bech32ToHexAddr result",
+	)
```
lumtis, line 28: We should add a test for the function using the fungible module (regularCall)

Although I suggest to define another function than regularCall

precompiles/types/errors.go
```diff
@@ -0,0 +1,27 @@
...
+func (e ErrInvalidNumberOfArgs) Error() string {
+	return fmt.Sprintf("invalid number of arguments; expected %d; got: %d", e.Expect, e.Got)
+}
+
+type ErrInvalidMethod struct {
+	Method string
+}
+
+func (e ErrInvalidMethod) Error() string {
+	return fmt.Sprintf("unknown method: %ss", e.Method)
```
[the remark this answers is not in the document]
fbac: Typo, fixing it.

precompiles/regular/regular.go
fbac: Makes sense, I'll modify the contract.

precompiles/regular/regular.go
fbac: This contract is part of bharvest work, I'm working currently to introduce unit testing and also switching from regularCall to a call that get the gas.

e2e/e2etests/e2etests.go
```diff
@@ -15,6 +15,7 @@ const (
 	TestZetaDepositRestrictedName = "zeta_deposit_restricted"
 	TestZetaWithdrawName          = "zeta_withdraw"
 	TestZetaWithdrawBTCRevertName = "zeta_withdraw_btc_revert" // #nosec G101 - not a hardcoded password
+	TestZetaPrecompilesName       = "zeta_precompiles"
```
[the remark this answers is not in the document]
fbac: Nice idea, applying it.

e2e/e2etests/test_precompiles_regular.go
fbac: Deleted regularCall and added a function to getGasStabilityBalancePool.

app/app.go
```diff
@@ -1049,10 +1054,18 @@ func (app *App) SimulationManager() *module.SimulationManager {
...
 	blockList := make(map[string]bool)
+
 	for k, v := range blockedReceivingModAcc {
 		addr := authtypes.NewModuleAddress(k)
 		blockList[addr.String()] = v
 	}
+
+	for addr, enabled := range precompiles.EnabledStatefulContracts {
+		if enabled {
+			blockList[addr.String()] = enabled
```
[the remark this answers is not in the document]
fbac: The logic is if a precompile contract is enabled, its address has to be included in BlockedAddresses so they're marked as non payable by the bank keeper.

precompiles/testutil/events.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+// validateEvents checks if the provided event names are included as keys in the contract events.
+func validateEvents(contractEvents map[string]abi.Event, events []string) ([]abi.Event, error) {
+	expEvents := make([]abi.Event, 0, len(events))
+	for _, eventStr := range events {
+		event, found := contractEvents[eventStr]
+		if !found {
+			availableABIEvents := make([]string, 0, len(contractEvents))
```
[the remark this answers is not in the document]
fbac: File was deleted.

cmd/zetae2e/local/precompiles.go
```diff
@@ -0,0 +1,56 @@
...
+
+// erc20TestRoutine runs erc20 related e2e tests
+func statefulPrecompilesTestRoutine(
+	conf config.Config,
+	deployerRunner *runner.E2ERunner,
+	verbose bool,
+	testNames ...string,
+) func() error {
+	return func() (err error) {
+		account := conf.AdditionalAccounts.UserERC20
```
lumtis, line 22: Reason we define new account is because the tests run in parallel, if same account send tx at the same time there might be issue with nonce, etc..
I think we can create a new account for Precompile interactions

e2e/e2etests/e2etests.go
```diff
@@ -116,6 +116,11 @@ const (
 	 Not used to test functionalities but do various interactions with the netwoks
 	*/
 	TestDeploy = "deploy"
+
+	/*
+	 Stateful precompiled contracts tests
+	*/
+	TestZetaPrecompilesPrototypeName = "precompiles_contract_prototype"
```
lumtis, line 123:
```suggestion
	TestZetaPrecompilesPrototypeName = "precompile_contracts_prototype"
```

precompiles/precompiles.go
```diff
@@ -0,0 +1,43 @@
+package precompiles
```
lumtis, line 1: I think this file can be excluded from test coverage

precompiles/prototype/IPrototype.go
```diff
@@ -0,0 +1,274 @@
+// Code generated - DO NOT EDIT.
```
lumtis, line 1: Can we exclude from code coverage?

precompiles/regular/regular.go
lumtis: So does it means this entire file is written from scratch?

precompiles/precompiles.go
fbac: Actually I just included tests in commit [e8c0a5de38ff9458d731f0599d51126471d43a4b](https://github.com/zeta-chain/node/pull/2703/commits/e8c0a5de38ff9458d731f0599d51126471d43a4b)

precompiles/regular/regular.go
fbac: Correct, the precompiled contracts are written from scratch and mostly following the same structure that you can see in prototype.go

precompiles/prototype/IPrototype.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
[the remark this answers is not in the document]
fbac: Removing it, as this contract is made by us.

precompiles/prototype/prototype.go
```diff
@@ -0,0 +1,260 @@
...
+func (c *Contract) Bech32ify(method *abi.Method, args []interface{}) ([]byte, error) {
+	if len(args) != 2 {
+		return nil, &ptypes.ErrInvalidNumberOfArgs{
+			Got:    len(args),
+			Expect: 2,
+		}
+	}
+
+	cfg := sdk.GetConfig()
+	prefix, _ := args[0].(string)
```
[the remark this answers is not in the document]
fbac: Adding it.

scripts/bindings-stateful-precompiles.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Generic function to generate bindings
+function bindings() {
+    cd $1
+    go generate > /dev/null 2>&1
```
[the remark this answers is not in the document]
fbac: This is similar as what we do with the TestDApp, the whole generate script is:
```go
//go:generate sh -c "solc IPrototype.sol --combined-json abi | jq '.contracts.\"IPrototype.sol:IPrototype\"'  > IPrototype.json"
//go:generate sh -c "cat IPrototype.json | jq .abi > IPrototype.abi"
//go:generate sh -c "abigen --abi IPrototype.abi  --pkg prototype --type IPrototype --out IPrototype.go"
```

IMHO it's convenient to have this hardcoded for every precompile, so the bindings are automatically generated always.

precompiles/prototype/IPrototype.sol
```diff
@@ -0,0 +1,35 @@
...
+pragma solidity ^0.8.26;
+
+/// @dev The IRegular contract's address.
+address constant IPROTOTYPE_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000000065; // 101
```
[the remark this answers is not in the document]
fbac: good catch, fixing!

precompiles/prototype/prototype.go
```diff
@@ -0,0 +1,260 @@
...
+) ([]byte, error) {
+	if len(args) != 1 {
+		return nil, &(ptypes.ErrInvalidNumberOfArgs{
+			Got:    len(args),
+			Expect: 1,
+		})
+	}
+
+	// Unwrap arguments. The chainID is the first and unique argument.
+	chainID, ok := args[0].(int64)
```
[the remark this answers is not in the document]
fbac: Initially I thought so, but changed it because of GetGasStabilityPoolBalance, which signature is:
`GetGasStabilityPoolBalance(ctx sdk.Context, chainID int64)`

This way seems to have less friction as we expect an int64 and it's passed directly to the function, without conversion. or possibility of overflow.

precompiles/prototype/prototype.go
```diff
@@ -0,0 +1,260 @@
...
+		default:
+			GasRequiredByMethod[methodID] = 0
+		}
+	}
+}
+
+type Contract struct {
+	ptypes.BaseContract
+
+	Ctx            sdk.Context
```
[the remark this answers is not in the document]
fbac: You're right, the ctx is actually passed by the function at ExecuteNative. Removed it.

precompiles/prototype/prototype.go
```diff
@@ -0,0 +1,260 @@
...
+	if err := sdk.VerifyAddressFormat(address.Bytes()); err != nil {
+		return nil, err
+	}
+
+	bech32Str, err := sdk.Bech32ifyAddressBytes(prefix, address.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	addressBz, err := sdk.GetFromBech32(bech32Str, "zeta")
```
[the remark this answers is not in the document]
fbac: Same as before.

cmd/zetae2e/config/localnet.yml
gartnera, line 1: Please keep these as double quotes for now to avoid merge conflicts.

You'd need to switch the  `DoubleQuotedString` type to `SingleQuotedString` for this to be consistent

fbac: Done!

cmd/zetae2e/local/precompiles.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"github.com/fatih/color"
+
+	"github.com/zeta-chain/zetacore/e2e/config"
+	"github.com/zeta-chain/zetacore/e2e/e2etests"
+	"github.com/zeta-chain/zetacore/e2e/runner"
+)
+
+// erc20TestRoutine runs erc20 related e2e tests
```
lumtis, line 14:
```suggestion
// statefulPrecompilesTestRoutine runs steateful precompiles related e2e tests
```

e2e/e2etests/test_precompiles_prototype.go
```diff
@@ -0,0 +1,33 @@
...
+	addr, err := iPrototype.Bech32ToHexAddr(nil, "zeta1h8duy2dltz9xz0qqhm5wvcnj02upy887fyn43u")
+	require.NoError(r, err, "Error calling Bech32ToHexAddr")
+	require.Equal(
+		r,
+		"0xB9Dbc229Bf588A613C00BEE8e662727AB8121cfE",
+		addr.String(),
+		"Failed to validate Bech32ToHexAddr result",
+	)
+
+	balance, err := iPrototype.GetGasStabilityPoolBalance(nil, int64(1337))
```
lumtis, line 30: For the chain ID we can do
```
r.EVMClient.ChainID()
```
(a function like that)
So it is compatible with any network

lumtis, line 30: Shouldn't we provide a context? We can use `r.Ctx`

fbac: The context is automatically provided in prototype.go, when the EVM executes the Run() function:
```go
		execErr := stateDB.ExecuteNativeAction(contract.Address(), nil, func(ctx sdk.Context) error {
			res, err = c.GetGasStabilityPoolBalance(ctx, method, args)
			return err
		})
```

cmd/zetae2e/local/precompiles.go
fbac: Applied!

e2e/e2etests/test_precompiles_prototype.go
fbac: Applied.

lumtis (approved):

precompiles/prototype/prototype.go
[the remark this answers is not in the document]
fbac: This was actually a bug. It has zeta hardcoded but it should use the variable prefix send by the user as arg[0].

Fixing it.

precompiles/prototype/prototype.go
```diff
@@ -0,0 +1,260 @@
...
+			Expect: 1,
+		}
+	}
+
+	address, ok := args[0].(string)
+	if !ok || address == "" {
+		return nil, fmt.Errorf("invalid bech32 address: %v", args[0])
+	}
+
+	bech32Prefix := strings.SplitN(address, "1", 2)[0]
```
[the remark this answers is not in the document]
fbac: The prefix could be "cosmos" for example. Let me revisit this function.

PR #2628, opened 2024-08-02
fix: adjust evm outbound tracker reporter to avoid submitting invalid hashes

ws4charlie:
# Description

- [x] move method `reportToOutboundTracker` from `chains/evm/signer/signer.go` to a new file `outbound_tracker_reporter.go`.
- [x] 1st change in   `reportToOutboundTracker`: DON'T report outbound hash any more after 20 mins timeout.
- [x] 2nd change in `reportToOutboundTracker`: simplified implementation by create sub method `IsTxConfirmed`.
- [x] 3rd change in `reportToOutboundTracker`: enforce `2 block` confirmation on tx receipt to prevent chain reorg (suggested by Panruo).
- [x] make `WatchOutbound` shorter by taking out inner tracker iteration loop and creating sub method `ProcessOutboundTrackers`.
- [x] Print log when zetaclient detects outbound trackers that are full of invalid hashes.
        The log print will contain pattern `outbound tracker is full of invalid hashes`
- [x] unified logger format with pre-defined field names, for instance `module, method, chain, nonce, tx, cctx` etc.

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Tested in development environment
- [x] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions

## Summary by CodeRabbit

- **New Features**
  - Enhanced outbound tracking mechanism to prevent submission of invalid hashes.
  - Added new constant `ZetaBlockTime` to represent estimated block timing.
  - Introduced `ReorgProtectBlockCount` to manage blockchain reorganizations.
  - Implemented `IsTxConfirmed` function for verifying transaction confirmations on Ethereum-like chains.
  - New function `reportToOutboundTracker` for managing outbound transaction tracking.
  
- **Bug Fixes**
  - Improved error handling in the logging functionality for outbound transactions.

- **Documentation**
  - Updated changelog to reflect new features and improvements.

- **Refactor**
  - Restructured outbound tracking logic into a dedicated method for better code organization.

PR #2706, opened 2024-08-14
fix: do not submit pending hash to tracker to avoid invalid hash

ws4charlie:
# Description

Cherry picked minimum fix from [the PR](https://github.com/zeta-chain/node/pull/2628):

- [x] on 20-min timeout, do not submit pending outbound tx hash to tracker to avoid invalid hash.
- [x] print log if outbound tracker is detected full of hashes.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/evm/evm_client.go
+++ b/zetaclient/evm/evm_client.go
@@ -374,7 +374,7 @@ func (ob *ChainClient) WatchOutTx() {
 					ob.logger.OutTx.Error().Msgf("WatchOutTx: confirmed multiple (%d) outTx for chain %d nonce %d", txCount, ob.chain.ChainId, nonceInt)
 				} else {
 					if len(tracker.HashList) == crosschainkeeper.MaxOutTxTrackerHashes {
-						ob.logger.OutTx.Error().Msgf("WatchOutTx: outbound tracker is full of invalid hashes for chain %d nonce %d", ob.chain.ChainId, nonceInt)
+						ob.logger.OutTx.Error().Msgf("WatchOutTx: outbound tracker is full of hashes for chain %d nonce %d", ob.chain.ChainId, nonceInt)
 					}
 				}
 			}
--- a/zetaclient/evm/evm_client.go
+++ b/zetaclient/evm/evm_client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/zeta-chain/zetacore/pkg/chains"
 	"github.com/zeta-chain/zetacore/pkg/coin"
 	"github.com/zeta-chain/zetacore/pkg/proofs"
+	crosschainkeeper "github.com/zeta-chain/zetacore/x/crosschain/keeper"
 	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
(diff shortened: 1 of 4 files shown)


zetaclient/evm/evm_client.go
```diff
@@ -371,6 +372,10 @@ func (ob *ChainClient) WatchOutTx() {
 					ob.SetTxNReceipt(nonceInt, outtxReceipt, outtx)
 				} else if txCount > 1 { // should not happen. We can't tell which txHash is true. It might happen (e.g. glitchy/hacked endpoint)
 					ob.logger.OutTx.Error().Msgf("WatchOutTx: confirmed multiple (%d) outTx for chain %d nonce %d", txCount, ob.chain.ChainId, nonceInt)
+				} else {
+					if len(tracker.HashList) == crosschainkeeper.MaxOutTxTrackerHashes {
+						ob.logger.OutTx.Error().Msgf("WatchOutTx: outbound tracker is full of invalid hashes for chain %d nonce %d", ob.chain.ChainId, nonceInt)
```
brewmaster012, line 377: we are not sure of the "invalid" part right?

brewmaster012 (approved):

zetaclient/evm/evm_client.go
ws4charlie: Yeah, we're not 100% sure, potentially.

CharlieMc0: To make this easier to parse the chainId and nonce should be attributes of the log message instead of embedded in the log message. These can be resued by other logs to so if we want to find all errors associated with a specific nonce it is very easy.

lumtis (commented):
The tracker is not added at all, we'll need to add it manually?

lumtis (approved):
Approving as I don't see anything concerning

brewmaster012:
> The tracker is not added at all, we'll need to add it manually?

a separate program is monitoring pending cctx and fill in missing tracker using etherscan API. 
Longer term, this functionality can be merged into zetaclient if an etherscan API is provided.

morde08:
> > The tracker is not added at all, we'll need to add it manually?
> 
> a separate program is monitoring pending cctx and fill in missing tracker using etherscan API. Longer term, this functionality can be merged into zetaclient if an etherscan API is provided.

It's not just etherscan though right? We'd also need Polygonscan, BSCscan, Basescan, etc. We would need one for every chain we support.

brewmaster012:
> > > The tracker is not added at all, we'll need to add it manually?
> > 
> > 
> > a separate program is monitoring pending cctx and fill in missing tracker using etherscan API. Longer term, this functionality can be merged into zetaclient if an etherscan API is provided.
> 
> It's not just etherscan though right? We'd also need Polygonscan, BSCscan, Basescan, etc. We would need one for every chain we support.

yes, etherscan API is one possible solution, but is just an example of a "indexer" that provides missing RPC calls from Eth JSON-RPC: namely the query txhash by from address & nonce. Alternatively, each zetaclient can index by themselves.

Issue #2696, opened 2024-08-13
`zetacore` : Invalid address fields may be truncated and processed in error


kingpinXD:
![image](https://github.com/user-attachments/assets/68d608c9-dd09-46c4-a0e1-63233e852cdf)

kingpinXD:
The validations for address should be placed in the messages themselves and not the ValidateBasic.
We use `AdditionalChains` which is set to the store and is required to validate the address format based on the chain

kingpinXD:
Related 
https://github.com/zeta-chain/node/issues/2234
https://github.com/zeta-chain/node/issues/2385

We should however limit the checks to address formats for now

kingpinXD:
The three messages 
MsgUpdateSystemContract : We can modify the validate basic to add the required validation 
MsgWhitelistErc20 : We can modify the validate basic to add the required validation 
MsgVoteInbound: This is more complicated as rejecting outright in the validate basic might be too harsh , for the end user as this does not give up any opportunity to refund back to the sender address .

kingpinXD:
With regards to evm_deposit , I am considering validating the receiver address based on coin-type, 

https://github.com/zeta-chain/zeta-node/blob/8da507f0cdfb5affbc5121a712670fbd1feaab11/x/crosschain/types/cctx.go#L175-L218

Which can be called at the beginning of the function `HandleEvmDeposit`
https://github.com/zeta-chain/zeta-node/blob/8da507f0cdfb5affbc5121a712670fbd1feaab11/x/crosschain/keeper/evm_deposit.go#L27-L30

This Validate is for ZEVM gateway only, and I considered placing this validation in the validate_inbound.go as well 
https://github.com/zeta-chain/zeta-node/blob/8da507f0cdfb5affbc5121a712670fbd1feaab11/x/crosschain/keeper/cctx_orchestrator_validate_inbound.go#L14-L57
And use gateway+cointype for conditions. 
This would have made more logical sense; however, due to the current structure of the code, this would not allow us to create a revert back to the sender address.

kingpinXD:
Draft pr for reference https://github.com/zeta-chain/node/pull/2707

PR #2707, opened 2024-08-14
refactor: add validations for evm address 

kingpinXD:
# Description

Closes https://github.com/zeta-chain/node/issues/2696

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Added a new test case for handling invalid address deposits in the Zeta deposit functionality.
	- Introduced a new function for validating cryptocurrency addresses across multiple blockchain networks.
	- Added comprehensive unit tests for address validation across various blockchain networks.
	- Enhanced error handling for invalid receiver addresses in cross-chain transactions.

- **Bug Fixes**
	- Improved error handling for invalid receiver addresses in the EVM deposit process.

- **Documentation**
	- Updated error messages for clarity in address validation processes.

- **Chores**
	- Removed obsolete address validation functions and tests to streamline codebase.


x/crosschain/types/cctx.go
```diff
@@ -203,6 +205,51 @@ func (m CrossChainTx) SetOutboundBallotIndex(index string) {
 }
 
 // GetCctxIndexFromBytes returns the CCTX index from a byte array.
+func (m CrossChainTx) GetValidReceiverAddress() (ethcommon.Address, error) {
```
kingpinXD, line 208: Will be adding more unit tests , and wanted to open up this pr for discussions wether this sort of validation based on coin type is the direction we want to go with

pkg/address/validate_address.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/zeta-chain/zetacore/pkg/chains"
+)
+
+// TODO : Use this function to validate Sender and Receiver address for CCTX
+// https://github.com/zeta-chain/node/issues/2697
+func ValidateAddressForChain(address string, chainID int64, additionalChains []chains.Chain) error {
```
kingpinXD, line 12: Same for this , Not using this function right now , but if used eventually it would make make the address validation a bit more strict . Wanted to get some thoughts before adding more unit tests and using this function

x/crosschain/types/cctx.go
```diff
@@ -203,6 +205,51 @@ func (m CrossChainTx) SetOutboundBallotIndex(index string) {
...
+			if err != nil {
+				return ethcommon.Address{}, cosmoserrors.Wrap(ErrInvalidReceiverAddress, err.Error())
+			}
+			if parsedAddress != (ethcommon.Address{}) {
+				to = parsedAddress.String()
+			}
+			err = address.ValidateEthereumAddress(to)
+			if err != nil {
+				return ethcommon.Address{}, cosmoserrors.Wrap(ErrInvalidReceiverAddress, err.Error())
+			}
```
fbac, line 236: Here we could expand `ValidateEthereumAddress` to also include the check `parsedAddress != (ethcommon.Address{})`, so all the logic behind checking the ETH address is contained in the same function.

x/crosschain/types/cctx.go
```diff
@@ -203,6 +205,51 @@ func (m CrossChainTx) SetOutboundBallotIndex(index string) {
...
+			return ethcommon.HexToAddress(m.GetCurrentOutboundParam().Receiver), nil
```
fbac, line 237:
```suggestion
			return ethcommon.HexToAddress(to), nil
```

x/crosschain/types/cctx.go
```diff
@@ -203,6 +205,51 @@ func (m CrossChainTx) SetOutboundBallotIndex(index string) {
...
+		}
+	// For coin type ERC20 we can use the receiver field of the inbound vote
+	case coin.CoinType_ERC20:
```
fbac, line 240: This case gets the same treatment as `case coin.CoinType_Zeta`, so we could do:
```suggestion
	case coin.CoinType_ERC20, coin.CoinType_Zeta:
```

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+	case chains.Network_base:
+		return nil
+	case chains.Network_solana:
+		return nil
+	default:
+		return fmt.Errorf("invalid network %d", chain.Network)
+	}
+}
+
+func ValidateEthereumAddress(address string) error {
```
swift1337, line 54:
```suggestion
func ValidateEVMAddress(address string) error {
```

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+	if !ethcommon.IsHexAddress(address) {
+		return fmt.Errorf("invalid address %s ", address)
```
swift1337, line 56:
```suggestion
		return fmt.Errorf("invalid evm address  %q", address)
```

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+	switch chain.Network {
+	case chains.Network_eth:
+		return ValidateEthereumAddress(address)
+	case chains.Network_zeta:
+		return nil
+	case chains.Network_btc:
+		{
+			addr, err := chains.DecodeBtcAddress(address, chainID)
+			if err != nil {
+				return fmt.Errorf("invalid address %s , chain %d: %s", address, chainID, err)
```
swift1337, line 32:
```suggestion
				return fmt.Errorf("invalid btc address %q, chain %d: %w", address, chainID, err)
```

e2e/runner/zeta.go
```diff
@@ -95,11 +95,11 @@ func (r *E2ERunner) DepositZeta() ethcommon.Hash {
 	amount := big.NewInt(1e18)
 	amount = amount.Mul(amount, big.NewInt(100)) // 100 Zeta
 
-	return r.DepositZetaWithAmount(r.EVMAddress(), amount)
+	return r.DepositZetaWithAmount(r.EVMAddress().Bytes(), amount)
 }
 
 // DepositZetaWithAmount deposits ZETA on ZetaChain from the ZETA smart contract on EVM with the specified amount
-func (r *E2ERunner) DepositZetaWithAmount(to ethcommon.Address, amount *big.Int) ethcommon.Hash {
+func (r *E2ERunner) DepositZetaWithAmount(to []byte, amount *big.Int) ethcommon.Hash {
```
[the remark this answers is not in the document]
swift1337: Let’s keep the E2E with `ethcommon.Address`, rather than refactoring to `[]byte` solely to test an invalid EVM address. I suggest we handle this case change separately.

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+func ValidateAddressForChain(address string, chainID int64, additionalChains []chains.Chain) error {
+	chain, found := chains.GetChainFromChainID(chainID, additionalChains)
+	if !found {
+		return fmt.Errorf("chain id %d not supported", chainID)
+	}
+	switch chain.Network {
+	case chains.Network_eth:
+		return ValidateEthereumAddress(address)
+	case chains.Network_zeta:
+		return nil
```
swift1337, line 27: This should also validate for evm

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+	case chains.Network_btc:
+		{
+			addr, err := chains.DecodeBtcAddress(address, chainID)
```
swift1337, line 30: let's move this to a separate `ValidateBTCAddres(...)`

pkg/address/validate_address.go
```diff
@@ -0,0 +1,59 @@
...
+				return fmt.Errorf("unsupported address %s", address)
+			}
+			return nil
+		}
+	case chains.Network_polygon:
+		return ValidateEthereumAddress(address)
+	case chains.Network_bsc:
+		return ValidateEthereumAddress(address)
+	case chains.Network_optimism:
+		return nil
```
swift1337, line 44: I think we need to validate by VM type, not by chain.network, all EVM's have the same addr

x/crosschain/types/cctx.go
swift1337, line 208:
```suggestion
func (m CrossChainTx) GetReceiver() (ethcommon.Address, error) {
```

pkg/address/validate_address.go
lumtis: Yes, I agree, to add all new check here we would need adding a new VM for Bitcoin (which can make sense as Bitcoin has some scripting functionality) but this would be higher scope than the PR.
I suggest to use VM for check with an exception for Bitcoin networks

x/fungible/types/message_update_system_contract.go
```diff
@@ -45,12 +45,9 @@ func (msg *MsgUpdateSystemContract) ValidateBasic() error {
...
 	}
 	// check if the system contract address is valid
-	if ethcommon.HexToAddress(msg.NewSystemContractAddress) == (ethcommon.Address{}) {
-		return cosmoserrors.Wrapf(
-			sdkerrors.ErrInvalidAddress,
-			"invalid system contract address (%s)",
-			msg.NewSystemContractAddress,
-		)
+	err = address.ValidateEthereumAddress(msg.NewSystemContractAddress)
+	if err != nil {
```
lumtis, line 49:
```suggestion
	if err := address.ValidateEthereumAddress(msg.NewSystemContractAddress);  err != nil {
```

x/crosschain/types/message_whitelist_erc20.go
```diff
@@ -54,8 +54,9 @@ func (msg *MsgWhitelistERC20) ValidateBasic() error {
 		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
 	}
 	// check if the system contract address is valid
-	if ethcommon.HexToAddress(msg.Erc20Address) == (ethcommon.Address{}) {
-		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid ERC20 contract address (%s)", msg.Erc20Address)
+	err = address.ValidateEthereumAddress(msg.Erc20Address)
+	if err != nil {
```
lumtis, line 58:
```suggestion
	
	if err := address.ValidateEthereumAddress(msg.Erc20Address); err != nil {
```

x/crosschain/types/cctx.go
```diff
@@ -203,6 +205,51 @@ func (m CrossChainTx) SetOutboundBallotIndex(index string) {
...
+		{
+			err := address.ValidateEthereumAddress(m.GetCurrentOutboundParam().Receiver)
+			if err != nil {
+				return ethcommon.Address{}, cosmoserrors.Wrap(ErrInvalidReceiverAddress, err.Error())
+			}
+			return ethcommon.HexToAddress(m.GetCurrentOutboundParam().Receiver), nil
+		}
+	// For coin type gas we create the message by composing the receiver and the call data so we need to parse the address from the RelayedMessage
+	// The receiver in this case is set to me the sender of the message
+	// For cases where this is not the case we should try and use the receiver field of the inbound vote
```
lumtis, line 222: This logic will be removed for V2 contracts, receiver will always contain the receiver address (which should have always been the case)

x/crosschain/keeper/evm_deposit.go
```diff
@@ -24,7 +24,10 @@ const InCCTXIndexKey = "inCctxIndex"
 // returns (isContractReverted, err)
 // (true, non-nil) means CallEVM() reverted
 func (k Keeper) HandleEVMDeposit(ctx sdk.Context, cctx *types.CrossChainTx) (bool, error) {
-	to := ethcommon.HexToAddress(cctx.GetCurrentOutboundParam().Receiver)
+	to, err := cctx.GetValidReceiverAddress()
```
lumtis, line 27: ctx.GetCurrentOutboundParam().Receiver will always contains the receiver moving forward, and since it is on ZetaChain, the address should always be EVM format

I think we could implement a function such as https://github.com/zeta-chain/node/issues/2789 and use it here with receiver address

e2e/runner/zeta.go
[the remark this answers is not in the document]
kingpinXD: I think using []byte or [20]byte here is a good idea; it keeps the function more generic

pkg/address/validate_address.go
kingpinXD: I agree with the fact the function should check based on the VM, which is in fact what this function does internally . 
All EVM based chains can use the same `ValidateEthereumAddress` function. 

Adding individual cases for each chain lets us provide a more granular check for each chain , and choose to ignore checks for some chains if we want to. 
IMO for checks like these it would be good be more cautious , and keep it at chain level

pkg/address/validate_address.go
kingpinXD: This function is not used right now
It would depend on the v2 protocol refactor on how this value is utilised .
Presently for btc inbound ,the receiver is set as `FromAddress` which would cause this check to fail and disable btc deposits .
Overall we should test this via end to end tests when using this function

x/crosschain/keeper/evm_deposit.go
kingpinXD: THe function is a good idea .
This is really the only change in logic that the PR implements, IMO if we want to change this , it would make sense to discard this pr for now , and revisit this once we have everything related to v2 functional

PR #2722, opened 2024-08-15
ci: Set Docker Workflow to use Go 1.22

CharlieMc0:
# Description

The workflow to push images to docker is not working and needs to be set to use Go 1.22. 
https://github.com/zeta-chain/node/actions/runs/10408541357

# How Has This Been Tested?

## Summary by CodeRabbit

- **New Features**
	- Enhanced GitHub Actions workflow to support Go programming environment with version 1.22.2.
	- Added a verification step to check the installed Go version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build Stage
-FROM ghcr.io/zeta-chain/golang:1.22.5-bookworm AS builder
+FROM golang:1.22-alpine3.18 AS builder
 
 ENV GOPATH /go
 ENV GOOS=linux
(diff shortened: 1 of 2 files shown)


julianrubino (approved):

gartnera (changes requested):
This is the go version outside the docker build. Per the errors you need to set the go version inside the docker build.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build Stage
-FROM golang:1.20-alpine3.18 AS builder
+FROM ghcr.io/zeta-chain/golang:1.22.5-bookworm AS builder
```
gartnera, line 2: This switches from alpine to debian which is a more involved change

gartnera (approved):

swift1337 (approved):

Issue #2719, opened 2024-08-15
`zetacore` : Unhandled error may result in incorrect query results for rate limiting and queries


kingpinXD:
![image](https://github.com/user-attachments/assets/d000265a-c7f0-4e9b-b4f8-2d3422d6918c)

PR #2725, opened 2024-08-15
refactor:  SetCctxAndNonceToCctxAndInboundHashToCctx receive tsspubkey as an argument

kingpinXD:
# Description
- This pr refactors the SetCctxAndNonceToCctxAndInboundHashToCctx to accept a tssPubkey instead of trying to fetch it. 
It does not modify any logic.

Closes https://github.com/zeta-chain/node/issues/2719
# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


x/crosschain/genesis.go
```diff
@@ -43,9 +42,14 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
...
-		if elem != nil {
-			k.SetCctxAndNonceToCctxAndInboundHashToCctx(ctx, *elem)
+	tss, found := k.GetObserverKeeper().GetTSS(ctx)
+	if found {
+		for _, elem := range genState.CrossChainTxs {
+			if elem != nil {
+				k.SetCctxAndNonceToCctxAndInboundHashToCctx(ctx, *elem, func(ctx sdk.Context) string {
+					return tss.TssPubkey
+				})
+			}
```
kingpinXD, line 52: Trying to retain older behaviour , but I feel we should use 
`elem.GetCurrentOutboundParam().TssPubkey` instead of trying to fetch the current tss.

the cctxs in the genstate would already have been created using the existing tss

PR #2735, opened 2024-08-16
fix(`zetaclient`): add zetaclient evm outbound tx index by nonce to supplement outtx tracker

### LINKED ISSUE #2846: missing or invalid outbound tx tracker blocks outbound processing
**Describe the Bug**
After TSS signers sign an outbound transaction and broadcasted to the external network, the 
zetaclient that bcast-ed successfully should report the outbound tx hash to the outtx tracker using
the `MsgAddOutboundTracker` msg. The reason that a outbound tx hash tracker is needed is because Ethereum JSON-RPC lack query tx/receipt by from_address+nonce. Zetaclients will need the txhash in order to vote on the confirmation of outbound tx, but they need to know the tx hash. The solution is to have one of the zetaclients report the txhash if bcast succeeded and receipts
become available. 

However this mechanism is not reliable. 
1. The zetaclient can report invalid txhash occasionally due to erroneous bcast RPC success returns; timeouts (this is no longer the case after a recent fix as timeout reports are usually false). 
2. the zetaclents can fail to report the txhash due to 
   - Bcast RPC erroneously report failure where in fact the bcast succeeded
   - Timeout: in the gas price hike cycle the tx could be pending for too long and zetaclient will give up waiting for receipt

Ideally, each zetaclient should locally index all outbound transaction so it would not rely completely on the outbound tx tracker
to discover outbound transaction hash and query its receipts. 

**To Reproduce**
Steps to reproduce the behavior.

**Expected Behavior**

**Screenshots**

**Environment (please complete the following information):**
- OS
- Version used

brewmaster012:
# Description

Currently outbound tracker is susceptible to being filled by invalid hashes and missing entries. 
Invalid hashes require manual cleanup by group proposal and missing entries require constant monitoring and filling in. 

This PoC PR adds simple logic in zetaclient to index outtx from TSS address by nonce, therefore avoid relying solely on outtx tracker for discovery of outbound tx hashes on Ethereum. 

Overhead (both memory and computation) added is minimal as it piggy backs on existing block by block scanner in inbound tx processing. 

- [x] Created a separate method `FilterTSSOutbound` in `outbound.go` to filter outbounds in blocks.
- [x] Unit tests thru evm rpc mocks.
- [x] Manually tested in localnet E2E by disabling outbound tracker in Ethereum client.

Closes #2846 

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced outbound tracking for EVM chains, improving transaction confirmation and processing.
	- Improved observation capabilities for TSS-related transactions, now handling both incoming and outbound processes.

- **Bug Fixes**
	- Resolved an issue with the outbound tracker that was blocking transaction confirmations on EVM chains.

- **Tests**
	- Updated test cases to reflect changes in transaction observation methods, ensuring comprehensive coverage of outbound operations.


zetaclient/chains/evm/observer/inbound.go
```diff
@@ -770,8 +770,8 @@ func (ob *Observer) BuildInboundVoteMsgForTokenSentToTSS(
 	)
 }
 
-// ObserveTSSReceiveInBlock queries the incoming gas asset to TSS address in a single block and posts votes
-func (ob *Observer) ObserveTSSReceiveInBlock(ctx context.Context, blockNumber uint64) error {
+// ObserveTSSReceiveInBlockAndOutTx queries the incoming gas asset to TSS address in a single block and posts votes
+func (ob *Observer) ObserveTSSReceiveInBlockAndOutTx(ctx context.Context, blockNumber uint64) error {
```
lumtis, line 774:
```suggestion
func (ob *Observer) ObserveTSSReceiveInBlockAndOutbound(ctx context.Context, blockNumber uint64) error {
```
We renamed all occurrence of outTx to outbound for consistency

zetaclient/chains/evm/observer/inbound.go
```diff
@@ -794,6 +794,12 @@
 				)
 			}
 		}
+		if ethcommon.HexToAddress(tx.From) == ob.TSS().EVMAddress() {
+			nonce := uint64(tx.Nonce)
+			if receipt, txx, ok := ob.checkConfirmedTx(ctx, tx.Hash, nonce); ok {
+				ob.SetTxNReceipt(nonce, receipt, txx)
```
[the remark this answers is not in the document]
brewmaster012: wut

zetaclient/chains/evm/observer/inbound.go
```diff
@@ -446,13 +446,13 @@ func (ob *Observer) ObserverTSSReceive(ctx context.Context, startBlock, toBlock
 	// query incoming gas asset
 	for bn := startBlock; bn <= toBlock; bn++ {
 		// observe TSS received gas token in block 'bn'
-		err := ob.ObserveTSSReceiveInBlock(ctx, bn)
+		err := ob.ObserveTSSReceiveInBlockAndOutbound(ctx, bn)
```
ws4charlie, line 449: maybe we could have the comment updated as `// observe TSS received gas token and outbounds txns signed by TSS in block 'bn'` ?

zetaclient/chains/evm/observer/inbound.go
```diff
@@ -770,8 +770,8 @@ func (ob *Observer) BuildInboundVoteMsgForTokenSentToTSS(
 	)
 }
 
-// ObserveTSSReceiveInBlock queries the incoming gas asset to TSS address in a single block and posts votes
-func (ob *Observer) ObserveTSSReceiveInBlock(ctx context.Context, blockNumber uint64) error {
+// ObserveTSSReceiveInBlockAndOutbound queries the incoming gas asset to TSS address in a single block and posts votes
```
ws4charlie, line 773:
```suggestion
// ObserveTSSReceiveInBlockAndOutbound queries the incoming gas asset to TSS address in a single block and posts votes.
// It also filters TSS outbounds indexed by nonce to supplement outtx tracker.
```

zetaclient/chains/evm/observer/inbound.go
```diff
@@ -794,6 +794,12 @@ func (ob *Observer) ObserveTSSReceiveInBlock(ctx context.Context, blockNumber ui
 				)
 			}
 		}
+		if ethcommon.HexToAddress(tx.From) == ob.TSS().EVMAddress() {
+			nonce := uint64(tx.Nonce)
```
ws4charlie, line 798: will need `// #nosec G115 always positive` to suppress nosec warning

brewmaster012: are they suppressed already in the CI gosec? I don't see complaints in CI.

ws4charlie: Not quite sure, maybe the dev ops team suppressed at some point. I see lots of `// $nosec G115` comments out there.

zetaclient/chains/evm/observer/inbound.go
```diff
@@ -794,6 +795,12 @@ func (ob *Observer) ObserveTSSReceiveInBlock(ctx context.Context, blockNumber ui
...
+			if receipt, txx, ok := ob.checkConfirmedTx(ctx, tx.Hash, nonce); ok {
```
ws4charlie, line 800: We could use the similar protection logic used in `ProcessOutboundTrackers` to avoid redundant calls on `checkConfirmedTx` (which involves a group of further RPCs)

```
if !ob.IsTxConfirmed(nonce) {
    // add this tx to supplement outtx tracker
    if receipt, txx, ok := ob.checkConfirmedTx(ctx, tx.Hash, nonce); ok {
	ob.SetTxNReceipt(nonce, receipt, txx)
    }
}
```

brewmaster012: done https://github.com/zeta-chain/node/pull/2735/commits/e787e70208e2193009e8b9431a77672f512900ed

ws4charlie (approved):

lumtis (commented):
I'm. sure to understand the change, we add new logic in TSS inbound observation. How does it concern outbound trackers?

lumtis (changes requested):
I see know the logic and idea. It would still be proper to define a new method like `ObserveTSSOutbound`, and calling this function in `ObserveInbound`. Ideally we should have this logic in `outbound.go` as it pertains to outbounds but the current state of the code doesn't allow it. ObserveInbound is fine as we iterate the blocks there.

I think the `ObserveTSSReceiveInBlockAndOutbound` naming is a bit confusing and the method is still called in `ObserverTSSReceive` which makes the control flow a bit more implicit.

Also `ObserverTSSReceive` will be removed in the future once we fully migrate to V2 contract and wind down v1 compatibility

>Tested in localnet E2E by disabling outbound tracker in Ethereum client. 

Is it possible to share the line you removed to disable the outbound trackers? Just wanted to do some checks as in the current code it seems we still rely on outbound trackers to watch the outbounds

ws4charlie:
> I see know the logic and idea. It would still be proper to define a new method like `ObserveTSSOutbound`, and calling this function in `ObserveInbound`. Ideally we should have this logic in `outbound.go` as it pertains to outbounds but the current state of the code doesn't allow it. ObserveInbound is fine as we iterate the blocks there.
> 
> I think the `ObserveTSSReceiveInBlockAndOutbound` naming is a bit confusing and the method is still called in `ObserverTSSReceive` which makes the control flow a bit more implicit.
> 
> Also `ObserverTSSReceive` will be removed in the future once we fully migrate to V2 contract and wind down v1 compatibility
> 
> > Tested in localnet E2E by disabling outbound tracker in Ethereum client.
> 
> Is it possible to share the line you removed to disable the outbound trackers? Just wanted to do some checks as in the current code it seems we still rely on outbound trackers to watch the outbounds

I think this PR was intended to be a supplement to (not to disable) outbound trackers. I think it is possible to minimize our reliance on outbound trackers in future and restrict its usage to only missed outbound (similar to inbound tracker). The new observation might look like:

- New impl will scan block and filter outbound transactions and won't solely relay on trackers, this could completely remove the complexity of tracker reporters.
- New impl will scan block by block sequentially and post vote sequentially. The current impl doesn't care the order by which the votes are posted.
- New impl will need to scan quick enough to catch up external chain in order to not cause delay. The current impl doesn't need to catch up on blocks.

overall: we could switch to a block-scanning model to minimize reliance on outbound trackers while still keep them for scenarios similar to inbound trackers.

brewmaster012:
> Is it possible to share the line you removed to disable the outbound trackers? Just wanted to do some checks as in the current code it seems we still rely on outbound trackers to watch the outbounds

image

i tested in localnet by disabling reporting to outbound tracker.

brewmaster012:
> Also `ObserverTSSReceive` will be removed in the future once we fully migrate to V2 contract and wind down v1 compatibility

then this function will just become ObserveOutbound
essentially a local indexer for all outbound txs. We can find a place for it that is more logical.

fbac (approved):

lumtis (approved):
Looks good

Issue #2786, opened 2024-08-28
Allow running e2e tests against tendermint/cometbft rpc


gartnera:
We currently require a gRPC endpoint to run the e2e tests. This makes it hard to debug them locally against real networks.

We should be able to proxy all gRPC requests via the cometbft `/abci_query`

lumtis:
Chaning label FYI, feature:idea is more for user/dev facing new features

Issue #959, opened 2023-08-14
Update the package name to `github.com/zeta-chain/node` to match the repo name


fadeev:
Replace `github.com/zeta-chain/zetacore` with `github.com/zeta-chain/node` in `go.mod` and all the imports

lumtis:
Hey @tantodefi , thanks for your interest in contributing.
Sounds good, assigning the issue to you, let us know if you have any questions.

tantodefi:
So i updated the package name everywhere and i saw that on PR template i should run smoketest  but i got the following error:
 `panic: failed to query rpc: post failed: Post "http://zetacore0:26657": dial tcp x.x.x.x:26657: connect: connection refused` smoketest failing on macOS after updating package name in go.mod and all imports... any ideas?
i can test again later on ubuntu vm but curious if im missing something obvious like maybe its macos firewall issue?
```cctxClient.TSS error rpc error: code = NotFound desc = not found TSS not ready yet, waiting for TSS to be appear in zetacore netowrk...``` found in orchestrator log while running smoketest
Any feedback would be appreciated..

**Comment by lumtis** [comment]
Hi @tantodefi, the error might be related to a Dockerfile that depends on the name of a package but it is curious. It shouldn't happen because of the firewall because smoketest is fully local.
Maybe the logs of `zetacore0` container can help:
```
cd contrib/localnet
docker-compose logs zetacore0
```
If `zetacore0` is not started, it is definitely an issue with the Docker image

PR #2788, opened 2024-08-28
feat: add common rpc package

gartnera:
We currently have logic to create zetacored rpc clients in both e2e and zetaclient/zetacore. We should deduplicate this logic and export it publicly so that we can easily make type safe RPCs in other codebases.

Also add a cometbft client so that it's easier to run e2e tests against anyone who exposes a cometbft/tendermint rpc interface. gRPC is not always exposed publicly.

Closes https://github.com/zeta-chain/node/issues/2786

External imports degraded by https://github.com/zeta-chain/node/issues/959. Fixed by https://github.com/zeta-chain/node/pull/2796

TODO:

- [x] initial client implementation
- [x] replace `zetaclient/zetacore.clients` struct. Update: embed the `rpc.Clients` struct rather than replacing as it has some other items. `rpc.Clients` is currently limited to read only operations.
- [x] replace `cmd/zetace2/config.zetaChainClients` struct
- [x] use `zetacore_rpc` rather than `zetacore_grpc` if `zetacore_grpc` is unset in e2e config

## Summary by CodeRabbit

- **New Features**
	- Introduced a unified `Clients` interface for easier interaction with various services within the ZetaChain ecosystem.
	- Added support for creating clients using both CometBFT and gRPC transport mechanisms, enhancing flexibility in service communication.


pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+
+// NewGRPCClient creates a Clients which uses gRPC as the transport
+func NewGRPClients(url string, opts ...grpc.DialOption) (Clients, error) {
+	grpcConn, err := grpc.Dial(url, opts...)
+	if err != nil {
+		return Clients{}, err
+	}
+	clientCtx := client.Context{}.WithGRPCClient(grpcConn)
+	return newClients(clientCtx)
+}
```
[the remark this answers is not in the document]
lumtis: Looks legit comment

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+	"google.golang.org/grpc"
+
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+// Clients contains RPC client interfaces to interact with zetacored
```
lumtis, line 19:
```suggestion
// Clients contains RPC client interfaces to interact with ZetaCore
```

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+type Clients struct {
+	AuthorityClient authoritytypes.QueryClient
+	CctxClient      crosschaintypes.QueryClient
+	FungibleClient  fungibletypes.QueryClient
+	AuthClient      authtypes.QueryClient
+	BankClient      banktypes.QueryClient
+	ObserverClient  observertypes.QueryClient
+	LightClient     lightclienttypes.QueryClient
```
lumtis, line 27:
```suggestion
	LightclientClient     lightclienttypes.QueryClient
```

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+// Clients contains RPC client interfaces to interact with zetacored
+type Clients struct {
+	AuthorityClient authoritytypes.QueryClient
+	CctxClient      crosschaintypes.QueryClient
```
lumtis, line 22:
```suggestion
	CrosschainClient      crosschaintypes.QueryClient
```
This is actually a wrong naming we have in the E2E suite

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+	FungibleClient  fungibletypes.QueryClient
+	AuthClient      authtypes.QueryClient
+	BankClient      banktypes.QueryClient
```
lumtis, line 25: Would be cleaner to group SDK clients and ZetaClient, and put those two above

lumtis (commented):
I think this is a direction that makes sense

Using grpc was more a shortcut and we should indeed remove it

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+		AuthClient:      authtypes.NewQueryClient(ctx),
+		BankClient:      banktypes.NewQueryClient(ctx),
+		ObserverClient:  observertypes.NewQueryClient(ctx),
+		LightClient:     lightclienttypes.NewQueryClient(ctx),
+	}, nil
+}
+
+// NewCometBFTClients creates a Clients which uses cometbft abci_query as the transport
+func NewCometBFTClients(url string) (Clients, error) {
+	cometRPCClient, err := rpcclient.New(url, "/websocket")
```
swift1337, line 44: Should be use HTTP instead?

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+	authoritytypes "github.com/zeta-chain/zetacore/x/authority/types"
+	crosschaintypes "github.com/zeta-chain/zetacore/x/crosschain/types"
+	fungibletypes "github.com/zeta-chain/zetacore/x/fungible/types"
+	lightclienttypes "github.com/zeta-chain/zetacore/x/lightclient/types"
+	observertypes "github.com/zeta-chain/zetacore/x/observer/types"
+)
+
+// Clients contains RPC client interfaces to interact with zetacored
+type Clients struct {
+	AuthorityClient authoritytypes.QueryClient
```
swift1337, line 21: Can we use embedding instead? (if these clients don't share same the method names)
```suggestion
	authoritytypes.QueryClient
```

This will simplify duck typing in cases when we want to pass a client with a subset of methods

pkg/rpc/rpc.go
gartnera: The transport is HTTP. It's using the cometbft rpc wrapper though.

pkg/rpc/rpc.go
gartnera: Interesting let me try it.

gartnera: Since they're all named `QueryClient` it seems you can't embed them. I guess I could type alias them but I'm not sure that would be great.

I think it might actually be a bit better to use the name so you know what types you need to import anyway.

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,61 @@
...
+	return Clients{
+		AuthorityClient: authoritytypes.NewQueryClient(ctx),
+		CctxClient:      crosschaintypes.NewQueryClient(ctx),
+		FungibleClient:  fungibletypes.NewQueryClient(ctx),
+		AuthClient:      authtypes.NewQueryClient(ctx),
+		BankClient:      banktypes.NewQueryClient(ctx),
+		ObserverClient:  observertypes.NewQueryClient(ctx),
+		LightClient:     lightclienttypes.NewQueryClient(ctx),
+	}, nil
+}
```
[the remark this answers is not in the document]
gartnera: Should be resolved

pkg/rpc/rpc.go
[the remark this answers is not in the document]
gartnera: Should be resolved

pkg/rpc/rpc.go
swift1337: Okay, makes sense. Then let's keep it as is but remove `*Client` suffix.
```go
type Clients struct {
	Authority authoritytypes.QueryClient
        // ...
}
```

gartnera: Yep already done :)

swift1337 (approved):

pkg/rpc/rpc.go
gartnera: Yeah this is the only client that satisfies the client parameter that is required: `func (ctx client.Context) WithClient(client client.TendermintRPC) client.Context`. Reviewing the imported code, the websocket is not opened unless you do an operation that requires it. Which should never happen since we do not expose the client directly.

lumtis (approved):

Issue #2752, opened 2024-08-20
`IsEVMChain` should check that the VM is EVM


lumtis:
**Describe the Issue**
Currently `IsEVMChain` is used to check the consensus while it should be used to check the VM (consensus is used for consensus related functionalities like block header)

PR #2801, opened 2024-08-30
fix(`pkg/chains`): check if VM is EVM for `IsEVMChain` check

lumtis:
# Description

`IsEVMChain` used to check if the chain consensus is Ethereum, which is invalid, the components using `IsEVMChain` don't perform logic that relies on consensus used by Ethereum.

Example: Base doesn't use Ethereum consensus as a L2 but comply with logic using isEVMChain.

This PR change the check.

All EVM chains so far had the value `ConsensusEthereum` except Base where we made a patch in the chain info. Therefore changing this check in theory doesn't bring any change in the workflow

Close: https://github.com/zeta-chain/node/issues/2752

## Summary by CodeRabbit

- **New Features**
	- Enhanced identification of EVM chains by broadening criteria from consensus mechanism to virtual machine type.
	- Improved error handling in the `IsEVMChain` function for better reliability.

- **Bug Fixes**
	- Updated test cases to correctly classify "Zeta Mainnet" as an EVM chain.

- **Tests**
	- Modified tests to use dynamic values for `SenderChainId` in cross-chain transactions, improving accuracy.

### DIFF
--- a/pkg/chains/chain.go
+++ b/pkg/chains/chain.go
@@ -86,7 +86,7 @@ func (chain Chain) EncodeAddress(b []byte) (string, error) {
 }
 
 func (chain Chain) IsEVMChain() bool {
-	return chain.Consensus == Consensus_ethereum
+	return chain.Vm == Vm_evm
 }
 
 func (chain Chain) IsBitcoinChain() bool {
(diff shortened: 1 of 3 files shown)


swift1337 (approved):
Can you explain why we need to distinguish EVM from let's say OP Stack? Any difference from Zeta's point of view?

kingpinXD (approved):
lgtm , 
While reviewing the usages of IsEVMChain function , I released that some logic in would be refactored to use network instead . Created an issue and will work on it 
https://github.com/zeta-chain/node/issues/2803.

Does not need any changes on this pr though

Issue #2827, opened 2024-09-04
V19 mainnet state export fails


gartnera:
```
panic: no concrete type registered for type URL /cosmos.auth.v1beta1.ModuleCredential against interface *types.PubKey
goroutine 1 [running]:
github.com/cosmos/cosmos-sdk/x/auth/keeper.AccountKeeper.decodeAccount({{0x3dc5148, 0xc0018196a0}, {0x3df83e0, 0xc000f39930}, {{0x3df83e0, 0xc000f39930}, 0xc000525c40, {0x3dc5148, 0xc001819790}, {0x3dc5198, ...}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.46.13/x/auth/keeper/keeper.go:224 +0xf4
github.com/cosmos/cosmos-sdk/x/auth/keeper.AccountKeeper.IterateAccounts({{0x3dc5148, 0xc0018196a0}, {0x3df83e0, 0xc000f39930}, {{0x3df83e0, 0xc000f39930}, 0xc000525c40, {0x3dc5148, 0xc001819790}, {0x3dc5198, ...}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.46.13/x/auth/keeper/account.go:102 +0x178
github.com/cosmos/cosmos-sdk/x/auth/keeper.AccountKeeper.ExportGenesis({{0x3dc5148, 0xc0018196a0}, {0x3df83e0, 0xc000f39930}, {{0x3df83e0, 0xc000f39930}, 0xc000525c40, {0x3dc5148, 0xc001819790}, {0x3dc5198, ...}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.46.13/x/auth/keeper/genesis.go:34 +0x1c5
github.com/cosmos/cosmos-sdk/x/auth.AppModule.ExportGenesis({{}, {{0x3dc5148, 0xc0018196a0}, {0x3df83e0, 0xc000f39930}, {{0x3df83e0, 0xc000f39930}, 0xc000525c40, {0x3dc5148, 0xc001819790}, ...}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.46.13/x/auth/module.go:151 +0x9b
[...]

PR #2828, opened 2024-09-04
fix: register all interfaces in encoding config

gartnera:
Fix https://github.com/zeta-chain/node/issues/2827

We have these `RegisterInterfaces()` calls copy pasted everywhere. Someone should probably clean theses up, but it would be nice to get state exports unblocked in the short term.

Will need to backport to `release/v19` and cut a 19.2.1

## Summary by CodeRabbit

- **New Features**
	- Enhanced encoding configuration by registering multiple interfaces, improving module interoperability and clarity.
  
- **Bug Fixes**
	- Resolved issues related to interface registration that could affect functionality across various modules.

### DIFF
--- a/app/encoding.go
+++ b/app/encoding.go
@@ -1,17 +1,47 @@
 package app
 
 import (
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/cosmos/cosmos-sdk/x/authz"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	crisistypes "github.com/cosmos/cosmos-sdk/x/crisis/types"
+	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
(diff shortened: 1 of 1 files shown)


app/encoding.go
```diff
@@ -1,17 +1,47 @@
...
+	upgradetypes.RegisterInterfaces(registry)
+	distrtypes.RegisterInterfaces(registry)
+	evidencetypes.RegisterInterfaces(registry)
+	crisistypes.RegisterInterfaces(registry)
+	evmtypes.RegisterInterfaces(registry)
+	ethermint.RegisterInterfaces(registry)
+	crosschaintypes.RegisterInterfaces(registry)
+	emissionstypes.RegisterInterfaces(registry)
+	fungibletypes.RegisterInterfaces(registry)
+	observertypes.RegisterInterfaces(registry)
```
lumtis, line 44: We might want to add the one from `authority` module? We have `lightclient` module as well but it is not used

swift1337 (approved):

app/encoding.go
gartnera: Updated.

julianrubino (approved):

PR #2849, opened 2024-09-07
fix: update go-tss with known fixes (v19)

gartnera:
https://github.com/zeta-chain/go-tss/pull/27

https://github.com/zeta-chain/node/pull/2849


CharlieMc0 (approved):
Needs change log message added

PR #2853, opened 2024-09-09
fix: calling precompile through sc with sc state update

skosito:
# Description

If stake precompile function is called through smart contract with some state changes on smart contract side, bank module state is not updated properly, it is overwritten, and bank balance is not reduced. In that case state on stateDB should manually be updated, so it is propagated to bank module update.

This PR adds counter in TestStaking.sol which is updated before and after stake call, and extends e2e test to verify that bank balance is reduced after stake method is executed. Also, adds one more function that stakes and reverts, to check that state on cosmos side is not updated.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Introduced new functions for staking operations: `stakeAndRevert` and `stakeWithStateUpdate`.
	- Added a view function `counter` to retrieve the current counter value.
	- Enhanced state tracking capabilities within the staking contract.

- **Bug Fixes**
	- Improved test validations for staking scenarios, ensuring accurate state management during reverts.

- **Documentation**
	- Updated contract interface documentation to reflect new functions and their parameters.


precompiles/staking/staking.go
```diff
@@ -231,6 +232,13 @@ func (c *Contract) Stake(
 		return nil, err
 	}
 
+	// if caller is not the same as origin it means call is coming through smart contract
+	// in that case, manually reduce amount in stateDB so it is properly reflected in bank module
```
lumtis, line 236: Maybe we can give more details in the comment why is it necessary when the caller is not origin

lumtis (approved):

fbac (approved):

Issue #2811, opened 2024-09-02
Extend precompile contracts with emitted events


skosito:
**Describe the Issue**
Add generic method to emit events in precompiles and add it to staking/prototype.

PR #2861, opened 2024-09-10
feat: emit events from staking precompile

skosito:
# Description

Emit events in stake, unstake and moveStake functions of staking precompile contract. Extend e2e tests to check if events are emitted.

closes: #2811 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Introduced new events for staking operations: `MoveStake`, `Stake`, and `Unstake` to enhance tracking and transparency.
	- Added logging functionalities for staking actions to improve event tracking within the blockchain.

- **Bug Fixes**
	- Updated tests to validate emitted staking events, ensuring correct handling of stake and unstake operations.

- **Documentation**
	- Enhanced documentation for new event declarations and logging functionalities related to staking.

### DIFF
--- a/precompiles/staking/logs.go
+++ b/precompiles/staking/logs.go
@@ -10,6 +10,12 @@ import (
 	"github.com/zeta-chain/node/precompiles/logs"
 )
 
+const (
+	StakeEventName     = "Stake"
+	UnstakeEventName   = "Unstake"
+	MoveStakeEventName = "MoveStake"
+)
+
 func (c *Contract) AddStakeLog(
 	ctx sdk.Context,
 	stateDB vm.StateDB,
@@ -17,7 +23,7 @@ func (c *Contract) AddStakeLog(
 	validator string,
 	amount *big.Int,
 ) error {
-	event := c.Abi().Events["Stake"]
+	event := c.Abi().Events[StakeEventName]
 
 	valAddr, err := sdk.ValAddressFromBech32(validator)
 	if err != nil {
@@ -48,7 +54,7 @@ func (c *Contract) AddUnstakeLog(
 	validator string,
 	amount *big.Int,
 ) error {
-	event := c.Abi().Events["Unstake"]
+	event := c.Abi().Events[UnstakeEventName]
 	valAddr, err := sdk.ValAddressFromBech32(validator)
 	if err != nil {
 		return err
@@ -79,7 +85,7 @@ func (c *Contract) AddMoveStakeLog(
 	validatorDst string,
 	amount *big.Int,
 ) error {
-	event := c.Abi().Events["MoveStake"]
+	event := c.Abi().Events[MoveStakeEventName]
 	validatorSrcAddr, err := sdk.ValAddressFromBech32(validatorSrc)
 	if err != nil {
 		return err


precompiles/staking/logs.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (c *Contract) addLog(ctx sdk.Context, stateDB vm.StateDB, topics []common.Hash, data []byte) {
+	stateDB.AddLog(&types.Log{
+		Address:     c.Address(),
+		Topics:      topics,
+		Data:        data,
+		BlockNumber: uint64(ctx.BlockHeight()),
+	})
+}
```
fbac, line 122: This function can be moved to precompiles package and be used from any contract.

precompiles/staking/logs.go
```diff
@@ -0,0 +1,154 @@
...
+
+	c.addLog(ctx, stateDB, topics, data)
+
+	return nil
+}
+
+func (c *Contract) AddMoveStakeLog(
+	ctx sdk.Context,
+	stateDB vm.StateDB,
+	staker common.Address,
```
kingpinXD, line 77: lets use delegator , to make it uniform with cosmos terminology

precompiles/staking/logs.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+)
+
+func (c *Contract) AddStakeLog(
+	ctx sdk.Context,
+	stateDB vm.StateDB,
+	staker common.Address,
```
kingpinXD, line 16: Lets use delgator, same for everywhere else in this package

precompiles/staking/logs.go
```diff
@@ -0,0 +1,154 @@
...
+	validator string,
+	amount *big.Int,
+) error {
+	event := c.Abi().Events["Stake"]
```
kingpinXD, line 20: Can we use delegate ? I think cosmos referees to stake as the creation of a validator

precompiles/staking/IStaking.sol
```diff
@@ -26,6 +26,38 @@ struct Validator {
...
 interface IStaking {
+    /// @notice Stake event is emitted when stake function is called
+    /// @param staker Staker address
+    /// @param validator Validator address
+    /// @param amount Coins amount
+    event Stake(
+        address indexed staker,
+        address indexed validator,
+        uint256 amount
+    );
```
kingpinXD, line 37: Does the calling of msgServer.Delegate , 
automatically create the cosmos event ? 

```
	ctx.EventManager().EmitEvents(sdk.Events{
		sdk.NewEvent(
			types.EventTypeDelegate,
			sdk.NewAttribute(types.AttributeKeyValidator, msg.ValidatorAddress),
			sdk.NewAttribute(types.AttributeKeyDelegator, msg.DelegatorAddress),
			sdk.NewAttribute(sdk.AttributeKeyAmount, msg.Amount.String()),
			sdk.NewAttribute(types.AttributeKeyNewShares, newShares.String()),
		),
	})

```

If so why do we need this event as well ? It seems to have duplicate information

precompiles/staking/staking.go
```diff
@@ -294,11 +300,18 @@ func (c *Contract) Unstake(
...
+	stateDB := evm.StateDB.(ptypes.ExtStateDB)
+	err = c.AddUnstakeLog(ctx, stateDB, stakerAddress, validatorAddress, amount)
+	if err != nil {
+		return nil, err
+	}
+
 	return method.Outputs.Pack(res.GetCompletionTime().UTC().Unix())
 }
 
 func (c *Contract) MoveStake(
```
kingpinXD, line 312: Why not use Redelagate?

kingpinXD (commented):
Had a quick look , maybe my understanding is wrong , so wanted to clarify before adding more comments. 

Should the staking precompiles functionality follow the naming / terminology of regular staking as much as possible , or is there a reason behind using different terms , because I see all three functions implemented have different names.

e2e/contracts/teststaking/TestStaking.sol
```diff
@@ -48,6 +48,25 @@ interface WZETA {
 
 // @dev Purpose of this contract is to call staking precompile
 contract TestStaking {
+    event Stake(
```
lumtis, line 51: Why are these defined here? Is it to generate parse function in the Go binding

precompiles/staking/logs.go
lumtis: We decieded to go with stake/unstake/moveStake terminology for the staking precompile

precompiles/staking/logs.go
lumtis: Maybe most function in this package could be made reusable, I think many contract will use packAmount as well

fbac (approved):
LGTM, I really like the shape this is taking.

precompiles/staking/logs.go
```diff
@@ -0,0 +1,113 @@
...
```
kingpinXD, line 20: Consider defining the event types and attributes as constants.

kingpinXD: Same for unstake and movestake

kingpinXD (commented):
Thanks for the clarification ,regarding the naming ,

I see the logs are added to the state DB. Would you be able to confirm what is the underlying database file which this gets added into ? `application.db` or `tx_index.db` 

Cosmos logs are stored as part of block result so they end up in tx_index.db which is fine because we can define the config for tx_index.db separately .
However if these logs end up in application.db , we should consider the implications on the state growth .

kingpinXD:
> Thanks for the clarification ,regarding the naming ,
> 
> I see the logs are added to the state DB. Would you be able to confirm what is the underlying database file which this gets added into ? `application.db` or `tx_index.db`
> 
> Cosmos logs are stored as part of block result so they end up in tx_index.db which is fine because we can define the config for tx_index.db separately . However if these logs end up in application.db , we should consider the implications on the state growth .

Actually it seems to be fine , the logs go the transient store , which is discarded , would be great if you could double check as well 
https://github.com/zeta-chain/ethermint/blob/2fad916e7179c604ae917b397e837b9f2b79125f/x/evm/keeper/state_transition.go#L288

lumtis (approved):

kingpinXD (approved):

Issue #942, opened 2023-08-09
`zetacore` : Delete Ballots after distribution


kingpinXD:
Ballots can be deleted from the state after they have matured to save space on the state.db

- Delete matured ballots
- Add an event when the ballot gets deleted, which can be indexed and used to query the list of voters if needed for debugging

ws4charlie:
Maybe we should not delete the ballots as the finalized ballots help us prevent double spending (by blocking the same ballot being created again)

kingpinXD:
We do have the finalized inbounds check to stop ballot creation 
https://github.com/zeta-chain/zeta-node/blob/db2c6f3ff7998c1f5f8500e853500a5d9d46adf9/x/crosschain/keeper/msg_server_vote_inbound_tx.go#L95-L97

ws4charlie:
> We do have the finalized inbounds check to stop ballot creation https://github.com/zeta-chain/zeta-node/blob/db2c6f3ff7998c1f5f8500e853500a5d9d46adf9/x/crosschain/keeper/msg_server_vote_inbound_tx.go#L95-L97

Sounds great. It will be safe then. We can go ahead to clean up ballots after distribution.

kingpinXD:
Deleting older ballots would remove the possibility of querying ballots older than 100 blocks (current maturity period ) 

This can affect current way of operations 
1. We wouldn't be able to query older ballots to see if directly if a particular observer is consistently missing votes 
2. We allow observers to vote on ballots to receive rewards , they would not be able to vote on ballots once deleted 



Proposed solutions 
When deleting a ballot, we can emit a new event containing the list of observers who have added their votes to the ballot. 
This should be enough, as we expect every observer to cast their vote within the maturity period (casting after that is useless as they would not get any rewards )

- The delete ballot event would be part of the BlockBeginner and not a tx_result.

- We can also consider removing the `Ballot_Created` even if it is not used by any consumer.


Changes required 
- The explorer/indexer should be able to index these deleted ballot events so that they can be easily searched. This should be very similar to existing emissions events such as 
`/zetachain.zetacore.emissions.internal.ObserverEmissions`
`/zetachain.zetacore.emissions.internal.BlockEmissions`


cc @zeta-chain/fullstack , @zeta-chain/devops

GMaiolo:
If these changes come within `/cosmos/tx/v1beta1/txs` when querying for a specific block event, then we don't need to update anything on the indexer side 👍

kingpinXD:
[...]

PR #2863, opened 2024-09-11
refactor(zetacore): delete ballots after maturity

kingpinXD:
# Description

Closes https://github.com/zeta-chain/node/issues/942

The PR does the following 
- Adds logic to delete matured ballots automatically 
- Adds a migration script to delete older matured ballots to clean up the state 
- Adds a new query to enable query of all ballots 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Added functionality to delete matured ballots, enhancing ballot lifecycle management.
  - Introduced a new `VoterList` message type to track individual voter details.
  - Added event handling for deleted ballots to improve system responsiveness.

- **Bug Fixes**
  - Enhanced logic for managing observer rewards and matured ballots.

- **Tests**
  - Expanded test coverage for ballot deletion and voter list generation, ensuring robustness in functionality.

- **Documentation**
  - Updated changelog to reflect the latest changes and features.


x/observer/keeper/ballot.go
```diff
@@ -65,3 +78,18 @@ func (k Keeper) AddBallotToList(ctx sdk.Context, ballot types.Ballot) {
 	list.BallotsIndexList = append(list.BallotsIndexList, ballot.BallotIdentifier)
 	k.SetBallotList(ctx, &list)
 }
+
+// ClearMaturedBallots deletes all matured ballots and the list of ballots for a given height.
```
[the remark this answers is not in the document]
kingpinXD: The rewards distribution loop already fetches the list once , and I wanted to reuse that without called `GetMaturedBallots` again 

https://github.com/zeta-chain/zeta-node/blob/1799d0942eb5baa51e82e486b931ace9aefdbe6f/x/emissions/abci.go#L96-L96

x/emissions/abci.go
```diff
@@ -102,11 +103,13 @@ func DistributeObserverRewards(
 	if len(ballotIdentifiers) == 0 {
 		return nil
 	}
+	ballots := make([]observertypes.Ballot, 0, len(ballotIdentifiers))
```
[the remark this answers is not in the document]
kingpinXD: We need the ballot , for the `EmitBallotDeleted` event .

I wanted to concentrate all the deletion logic into `ClearMaturedBallots` which is why I am emitting the event from inside the function instead of from emissons/abci.go directly

x/observer/types/ballot.go
```diff
@@ -115,3 +115,22 @@ func (m Ballot) BuildRewardsDistribution(rewardsMap map[string]int64) int64 {
 	}
 	return totalRewardUnits
 }
+
+// GenerateVoterList generates a list of voters from the `VoterList` and `Votes` fields
+// This is used for queries and events and not persisted to the store
+func (m Ballot) GenerateVoterList() ([]VoterList, error) {
+	if len(m.VoterList) != len(m.Votes) {
+		return nil, ErrInvalidVoterList
```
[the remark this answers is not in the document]
kingpinXD: I don't think it can happen, but I don't see any harm in checking.

The function is used for queries and events only for now

x/observer/keeper/ballot.go
```diff
@@ -58,10 +71,25 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
 	if !found {
 		list = types.BallotListForHeight{Height: ballot.BallotCreationHeight, BallotsIndexList: []string{}}
 	}
 	list.BallotsIndexList = append(list.BallotsIndexList, ballot.BallotIdentifier)
 	k.SetBallotList(ctx, &list)
 }
+
+// ClearMaturedBallots deletes all matured ballots and the list of ballots for a given height.
+// It also emits an event for each ballot deleted.
+func (k Keeper) ClearMaturedBallots(ctx sdk.Context, ballots []types.Ballot, maturityBlocks int64) {
```
swift1337, line 84: This method should not accept any input besides context. Current method signature allows deleting any `ballots []types.Ballot`

kingpinXD: We need the `maturityBlocks` parameter , as its from the emissions params 
`ballots` are passed as a param to reduce the number of reads to the store , as ballot are already fetched once here 
https://github.com/zeta-chain/zeta-node/blob/afaa312cdc815babff7ec59bc90cca36b45f0d0d/x/emissions/abci.go#L106-L114

The function name can be changed though ,cause it does more than just clear ballots , it deletes the ballot list as well

fbac (approved):

x/observer/keeper/ballot.go
```diff
@@ -58,10 +71,25 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
 	if !found {
 		list = types.BallotListForHeight{Height: ballot.BallotCreationHeight, BallotsIndexList: []string{}}
 	}
 	list.BallotsIndexList = append(list.BallotsIndexList, ballot.BallotIdentifier)
 	k.SetBallotList(ctx, &list)
 }
+
+// ClearMaturedBallotsAndBallotList deletes all matured ballots and the list of ballots for a given height.
+// It also emits an event for each ballot deleted.
+func (k Keeper) ClearMaturedBallotsAndBallotList(ctx sdk.Context, ballots []types.Ballot, maturityBlocks int64) {
```
lumtis, line 84: I agree with @swift1337 , we should add more encapsulation, the external module shouldn't need to provide a list of ballot

-> You provide an height
-> You can get all ballots from that height in the function
-> You delete all the ballots

x/observer/keeper/ballot.go
```diff
@@ -58,10 +71,25 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
+	for _, ballot := range ballots {
+		k.DeleteBallot(ctx, ballot.BallotIdentifier)
+		EmitEventBallotDeleted(ctx, ballot)
```
lumtis, line 87: I don't think we need an event here

x/observer/keeper/ballot.go
```diff
@@ -58,10 +71,25 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
+	}
+	k.DeleteBallotListForHeight(ctx, GetMaturedBallotHeight(ctx, maturityBlocks))
+	return
+}
+
+func GetMaturedBallotHeight(ctx sdk.Context, maturityBlocks int64) int64 {
```
lumtis, line 93:
```suggestion
func getMaturedBallotHeight(ctx sdk.Context, maturityBlocks int64) int64 {
```
need comment

x/observer/keeper/ballot.go
```diff
@@ -58,10 +71,25 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
 func (k Keeper) AddBallotToList(ctx sdk.Context, ballot types.Ballot) {
-	list, found := k.GetBallotList(ctx, ballot.BallotCreationHeight)
+	list, found := k.GetBallotListForHeight(ctx, ballot.BallotCreationHeight)
 	if !found {
 		list = types.BallotListForHeight{Height: ballot.BallotCreationHeight, BallotsIndexList: []string{}}
 	}
 	list.BallotsIndexList = append(list.BallotsIndexList, ballot.BallotIdentifier)
 	k.SetBallotList(ctx, &list)
 }
+
```
lumtis, line 81: We should add a function allowing to delete all matured ballots ever and run it in the migration. We should also check efficiency in the mainnet upgrade beforehand, might be long with all the ballots

lumtis (changes requested):

x/observer/keeper/ballot.go
kingpinXD: Once indexed, the event can be used to query details about deleted ballots without having to query an archive node with an older height. 
https://github.com/zeta-chain/node/issues/942#issuecomment-2315975640

x/observer/keeper/ballot.go
kingpinXD: I don't know if deleting the older ballots would help with the archive node's growth; non-archives would get pruned at some point. 
Happy to add that in and test if you think we still need it

lumtis: This will allow to have a cleaner state in general, we can then query all the existing ballots to get the active ones.

The state of the store would make less sense otherwise, you would have active ballots but still matured ballots for older blocks?

x/observer/keeper/ballot.go
lumtis: I don't think you need to reread the store for ballots? The ballot list from height give the ID list that can be directly used in the delete function?

This is small optimization in any case I would consider the encapsulation of the code more important

x/observer/keeper/ballot.go
kingpinXD: Hmm , I guess we could have a query for all ballots if we delete the older ones, 
Ill look into testing the upgrade handler then

x/observer/keeper/ballot.go
kingpinXD: That is true, but the delete ballot event uses other fields of the ballot. 

We could alternatively fire the event from abci.go, but I feel that firing it from the for loop is better

x/observer/keeper/ballot.go
kingpinXD: Tested with mainnet data , the delete time is less than 1s
The data is is ~700MB 

We are waiting for the test snapshot to be available before resolving this comment. The testnet data is too large to be able to parse easily , I am working with Julian to extract just the observer data , instead of the whole export

x/observer/keeper/ballot.go
lumtis: The current interface doesn't respect encapsulation principle. We delete ballots from a list and a `BallotListForHeight` value, while the list of ballots can be derived from this height. Both values should be deleted atomically from the same inputs, the current interface allows an external module to create an inconsistent state for the ballot

x/observer/migrations/v9/migrate.go
gartnera, line 1: Do we really need the migration logic too or can we just use the `BeginBlocker` logic? Couldn't we limit the total number of deleted ballots to amortize the cost?

kingpinXD: The begin blocker logic would delete a ballots from when it gets deployed. 
For anything before that we need the migration script. We want to remove the state so that we can create a query like `GetAllBallots` an `GetBallotsbyStatus` in the future, which would be very useful for debugging .

lumtis:
Small note: let's keep this one aside for now to not introduce migration scripts as we already heavy operations with gateway and Solana

lumtis:
> Hi @lumtis Should we revisit this for the next release?
> 
> We can consider breaking the pr down, into
> 
> * Change in logic
> * Migration script
>   And only add the former if in case a long-running migration script is not desired

We actually want to remove all old ballot so we can easily query all active ballots

x/observer/keeper/ballot.go
kingpinXD: I do agree though we should remove the event if its not useful .

x/observer/migrations/v9/migrate.go
```diff
@@ -0,0 +1,38 @@
...
+	for i := maturedHeight; i > 0; i-- {
+		ballotList, found := observerKeeper.GetBallotListForHeight(ctx, i)
+		if !found {
+			continue
+		}
+		for _, ballotIndex := range ballotList.BallotsIndexList {
+			observerKeeper.DeleteBallot(ctx, ballotIndex)
+		}
+		observerKeeper.DeleteBallotListForHeight(ctx, i)
+	}
```
gartnera, line 35: I seem to remember you saying this would take a long time right?

x/observer/keeper/ballot.go
```diff
@@ -30,18 +42,21 @@ func (k Keeper) GetBallot(ctx sdk.Context, index string) (val types.Ballot, foun
 	return val, true
 }
 
-func (k Keeper) GetBallotList(ctx sdk.Context, height int64) (val types.BallotListForHeight, found bool) {
+func (k Keeper) GetBallotListForHeight(ctx sdk.Context, height int64) (val types.BallotListForHeight, found bool) {
```
ws4charlie, line 45: Would just returning a `types.BallotListForHeight` be enough for the caller?

kingpinXD: I think it should be fine , but this pr just renames the function , 
We can consider changing the signature in a separate PR, as the function is used in a few different places.

x/observer/keeper/ballot.go
```diff
@@ -58,10 +73,41 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
+	if !found {
+		return
+	}
+	// Delete all the matured ballots and emit an event for each ballot deleted
+	for _, ballotIndex := range maturedBallots.BallotsIndexList {
+		ballot, found := k.GetBallot(ctx, ballotIndex)
+		if !found {
+			continue
+		}
+		k.DeleteBallot(ctx, ballotIndex)
```
ws4charlie, line 102: what happens if the ballot is still pending with minority votes (e.g., 4/9)?

kingpinXD: I think the chances of that are quite unlikely, that is why we have maturity blocks 
However, if that does happen, a new ballot would be created when the fifth vote comes in, and the four observers who already voted would need to revote. This can be triggered by adding trackers manually.

Overall though the maturity blocks should be adjusted so that it has enough time for a ballot to finalize as we do not want to keep the ballots in the system indefinitly

x/observer/migrations/v9/migrate.go
kingpinXD: yeah 
https://zetachain.slack.com/archives/C04GM8RV5E3/p1729172612290599

x/observer/keeper/ballot.go
kingpinXD: Events would go into the blockstore , and I think nodes have the option to not store them through configuration, events usually are not part of the application.db .
But yeah, we should avoid adding anything to the node that we don't need 
The primary use case that I can think of , where we use the data from these events is for debugging to see if there is a particular observer missing votes consistently fro more than maturityBlocks

x/observer/migrations/v9/migrate.go
gartnera: > its taking areound 12 secs on mainnet and 21 secs on testnet data to run the delete.

yeah I think that's fine

proto/zetachain/zetacore/observer/events.proto
```diff
@@ -15,6 +16,13 @@ message EventBallotCreated {
   string ballot_type = 5;
 }
 
+message EventBallotDeleted {
```
swift1337, line 19: Do we even need to have this event? In this case, each vote will have `EventBallotCreated` and `EventBallotDeleted`

x/observer/migrations/v9/migrate.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/zeta-chain/node/x/observer/types"
+)
+
+type observerKeeper interface {
+	GetBallotListForHeight(ctx sdk.Context, height int64) (val types.BallotListForHeight, found bool)
+	DeleteBallot(ctx sdk.Context, index string)
+	DeleteBallotListForHeight(ctx sdk.Context, height int64)
+}
+
+const MaturityBlocks = int64(100)
```
swift1337, line 15: Should we provide maturityBlocks as an arg here?

x/observer/migrations/v9/migrate.go
swift1337: We can try to spin up 4-8 goroutines for this migration and test whether will be faster

x/observer/keeper/ballot.go
```diff
@@ -58,10 +73,41 @@ func (k Keeper) GetAllBallots(ctx sdk.Context) (voters []*types.Ballot) {
...
 	if !found {
 		list = types.BallotListForHeight{Height: ballot.BallotCreationHeight, BallotsIndexList: []string{}}
 	}
 	list.BallotsIndexList = append(list.BallotsIndexList, ballot.BallotIdentifier)
 	k.SetBallotList(ctx, &list)
 }
+
+// ClearMaturedBallotsAndBallotList deletes all matured ballots and the list of ballots for a given height.
+// It also emits an event for each ballot deleted.
+func (k Keeper) ClearMaturedBallotsAndBallotList(ctx sdk.Context, maturityBlocksParam int64) {
```
swift1337, line 86: I think the naming is too verbose. Consider renaming this to `CleanupBallots(...)`

proto/zetachain/zetacore/observer/events.proto
kingpinXD: Same answer as this : https://github.com/zeta-chain/node/pull/2863#discussion_r1894339848

x/observer/migrations/v9/migrate.go
kingpinXD: The maturity blocks are part of the emissions params, which makes it challenging to use the parameter. 
I think using the constant here is fine for now

x/observer/keeper/ballot.go
kingpinXD: Imo , CleanupBallots might be less verbose than necessary, I do not see any harm in having a very verbose function name though

x/observer/migrations/v9/migrate.go
kingpinXD: I actually tried with some goroutines earlier , something like this 
```
func MigrateStore(ctx sdk.Context, observerKeeper observerKeeper) error {
	currentHeight := ctx.BlockHeight()
	if currentHeight < MaturityBlocks {
		return nil
	}
	maturedHeight := currentHeight - MaturityBlocks

	var wg sync.WaitGroup
	numGoroutines := 10
	blockRange := maturedHeight / int64(numGoroutines)

	for i := 0; i < numGoroutines; i++ {
		start := int64(i) * blockRange
		end := start + blockRange
		if i == numGoroutines-1 {
			end = maturedHeight + 1
		}

		wg.Add(1)
		go func(start, end int64) {
			defer wg.Done()
			for j := start; j < end; j++ {
				ballotList, found := observerKeeper.GetBallotListForHeight(ctx, j)
				if !found {
					continue
				}
				for _, ballotIndex := range ballotList.BallotsIndexList {
					observerKeeper.DeleteBallot(ctx, ballotIndex)
				}
				observerKeeper.DeleteBallotListForHeight(ctx, j)
			}
		}(start, end)
	}

	wg.Wait()
	return nil
}
```

although it does show an improvement , 
Without goroutine is 11.5 secs 
With goroutine is 8.6 secs 

I decided to go in favour of the simpler logic as the difference in time was not as much .

x/observer/keeper/ballot.go
swift1337: Yeah it's subjective. I prefer concise `DoStuff()` over `DoThisAndThenThis()`

x/observer/migrations/v9/migrate.go
swift1337: > The maturity blocks are part of the emissions params

Why emission and not observer / cross-chain ?

kingpinXD: Its legacy code ,and should be changed , 
But not in this PR

x/observer/keeper/ballot.go
ws4charlie: We can mark it as resolved. Deleting old ballots looks good, regardless of pending or not. We could always add missed inbound tracker back to let the zetaclients to vote again.

x/observer/keeper/events.go
```diff
@@ -36,6 +37,24 @@ func prettyPrintStruct(val interface{}) string {
 	return string(prettyStruct)
 }
 
+func EmitEventBallotDeleted(ctx sdk.Context, ballot types.Ballot) {
+	var voterList []types.VoterList
```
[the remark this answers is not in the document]
kingpinXD: It's not needed, but it provides a better output compared to two strings.
Is there any reason why adding an additional type could be a problem?

x/observer/keeper/grpc_query_ballot.go
```diff
@@ -40,20 +42,49 @@ func (k Keeper) BallotByIdentifier(
...
+	ballots := make([]types.Ballot, 0)
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	ballotStore := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.VoterKey))
+
+	if req.Pagination == nil {
+		req.Pagination = &query.PageRequest{}
+	}
+
+	if req.Pagination.Limit > 100 {
```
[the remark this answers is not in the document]
kingpinXD: Thats true, we dont. 
However I think it is safer to add a pagination limit on the query so that an end user cannot use a very high number for the limit.
I am planning to add this to other queries such as the ones for CCTX as well .

x/observer/keeper/events.go
```diff
@@ -36,6 +37,24 @@ func prettyPrintStruct(val interface{}) string {
...
+	voterList, err := ballot.GenerateVoterList()
+	if err != nil {
+		ctx.Logger().
+			Error(fmt.Sprintf("failed to generate voter list for ballot %s", ballot.BallotIdentifier), err.Error())
+	}
+	err = ctx.EventManager().EmitTypedEvent(&types.EventBallotDeleted{
+		MsgTypeUrl:       "zetachain.zetacore.observer.internal.BallotDeleted",
```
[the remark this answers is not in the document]
kingpinXD: Remove this field , This was initially added to maintain a similar structure with other events , 
I have added a issue to refactor other events emitted from block beginner as well , but would make that change in a separate pr after verifying that it does not affect any end users 
https://github.com/zeta-chain/node/issues/3335

proto/zetachain/zetacore/observer/ballot.proto
```diff
@@ -41,3 +41,8 @@ message BallotListForHeight {
   int64 height = 1;
   repeated string ballots_index_list = 2;
 }
+
+message VoterList {
```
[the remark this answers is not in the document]
kingpinXD: 0069905e84bef2e2d0b1147eceaed2a0762bc6b9

x/observer/keeper/events.go
```diff
@@ -36,6 +37,23 @@ func prettyPrintStruct(val interface{}) string {
 	return string(prettyStruct)
 }
 
+func EmitEventBallotDeleted(ctx sdk.Context, ballot types.Ballot) {
```
swift1337, line 40: I think we agreed that this should be removed, right? cc @lumtis

lumtis: I looked at the change today, I think it would be nice to have a trace of the removal for the time being for debugging. But on the other side, this events bring lot of lines in the codebase, in addition to defining a type we don't use. 
It might make more sense to just use a debug log for ballot removal.

gartnera: Just to be clear I am still opposed because this will just be one more thing that increases the state size.

Our archive snapshot size takes roughtly ~20 hours to download and restore which is pretty brutal.

> It might make more sense to just use a debug log for ballot removal.

Yes I prefer this way

x/observer/keeper/grpc_query_ballot.go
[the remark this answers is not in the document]
lumtis: >Are you saying having something like a config flag that each server operator could set?

We can't set a max size for the query length one the node directly?

lumtis:
The PR looks good to me, let's just wrap up v25 first before merging it

x/observer/keeper/grpc_query_ballot.go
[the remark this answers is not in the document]
kingpinXD: I have removed the check for now, but if we don't have a way to control the maximum limit, I think it's safer to have a limit on the protocol side

lumtis (approved):

testutil/sample/logger.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/cometbft/cometbft/libs/log"
+)
+
+type TestLogger struct {
+	buf bytes.Buffer
+	log.Logger
+}
+
+func NewTestLogger() *TestLogger {
```
swift1337, line 14: Please remove this and use `testlog` package

x/observer/keeper/export_private_functions_test.go
```diff
@@ -0,0 +1,6 @@
+package keeper
+
+// These functions are exported for testing purposes
+
+var GetMaturedBallotHeightFunc = getMaturedBallotHeight
+var LogBallotDeletionFunc = logBallotDeletion
```
swift1337, line 6: Why is this needed? Consider either exporting these functions or not using them in tests

kingpinXD: These functions are used only in the same file, so it makes sense to have them private.

testutil/sample/logger.go
kingpinXD: Minimizing dependencies on external packages is a good idea; I don't think we need additional features from a test logger other than what this file provides

x/observer/keeper/export_private_functions_test.go
swift1337: In this case you can create `ballot_internal_test.go` with `package keeper` instead of `keeper_test` and do the following:
```go

// ...

t.Run(tc.name, func(t *testing.T) {
	// ...
	require.Equal(t, tc.expectedHeight, getMaturedBallotHeightFunc(ctx, tc.maturityBlocks))
})
```

Otherwise, it breaks the design flow of Go packages

testutil/sample/logger.go
gartnera: I'm pretty sure this is a new internal package that Dmitry has introduced

x/observer/keeper/export_private_functions_test.go
gartnera: So does doing it this way prevent a non-test package from using `GetMaturedBallotHeightFunc`?

testutil/sample/logger.go
kingpinXD: Yeah , Dimitry posted the link in slack . 
here is a link to the converstation 
https://zetachain.slack.com/archives/C04GM8RV5E3/p1736791804185059?thread_ts=1736790665.346879&cid=C04GM8RV5E3

swift1337 (approved):

PR #2912, opened 2024-09-23
ci: Add Semgrep to CI

jkan2:
# Description

adds semgrep to github actions

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Introduced a new GitHub Actions workflow for continuous integration checks using Semgrep.
	- The workflow runs automatically on specified events and is scheduled to execute daily.

- **Bug Fixes**
	- The workflow is designed to skip execution for changes made by the dependabot bot.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -5,8 +5,6 @@ on:
   push:
     branches:
       - develop
-    paths:
-      - .github/workflows/semgrep.yml
   schedule:
     # random HH:MM to avoid a load spike on GitHub Actions at 00:00
     - cron: '56 22 * * *'


CharlieMc0 (approved):

lumtis (approved):

gartnera (changes requested):
It doesn't seem to be actually running? it exited instantly?

```
run `semgrep login` before using `semgrep ci` or use `semgrep scan` and set `--config`
```

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,25 @@
+name: Semgrep
+on:
+  workflow_dispatch: {}
+  pull_request: {}
+  push:
+    branches:
+      - main
+      - master
```
gartnera, line 8: develop is our default branch

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,25 @@
...
+    paths:
+      - .github/workflows/semgrep.yml
```
gartnera, line 10: This will only run on push to this exact file, is that the desired behavior?

CharlieMc0: No it's not. There should be no path set since we want to run it on everything except E2E tests but that'll be configured in the semgrep config

gartnera: We typically like to run all CI again on push to develop as the merged result may be different than the PR contents.

gartnera (approved):

PR #2928, opened 2024-09-26
chore: improve localnet build performance

gartnera:
# Description

- Don't use a timestamp if dirty build. Doing this forced unconditional rebuilds every run.
- Don't copy the git directory into the build. This requires that we also provide a way to push the version info into the build.

With this change you should be able to run `make start-e2e-test` and if nothing has changed `zetanode` should not be rebuilt.

I could also be convinced that the localnet build should always just be static. Like `v22.0.0-next` or something like that.

## Summary by CodeRabbit

- **New Features**
	- Enhanced versioning logic in the `version.sh` script, providing more detailed version information including short commit hashes.
  
- **Improvements**
	- Docker build process optimized with new arguments for version control.
	- Streamlined handling of legacy versions in Dockerfile.

- **Bug Fixes**
	- Adjusted output format for versioning in `version.sh` to include commit details for both dirty and non-dirty builds.

### DIFF
--- a/Dockerfile-localnet
+++ b/Dockerfile-localnet
@@ -22,8 +22,10 @@ COPY --exclude=*.sh --exclude=*.md --exclude=*.yml . .
 ARG NODE_VERSION
 ARG NODE_COMMIT
 
-RUN --mount=type=cache,target="/root/.cache/go-build" make install
-RUN --mount=type=cache,target="/root/.cache/go-build" make install-zetae2e
+RUN --mount=type=cache,target="/root/.cache/go-build" \
+    NODE_VERSION=${NODE_VERSION} \
+    NODE_COMMIT=${NODE_COMMIT} \
+    make install install-zetae2e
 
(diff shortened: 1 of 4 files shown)


Dockerfile-localnet
```diff
@@ -19,6 +19,8 @@ COPY go.sum .
 RUN go mod download
 COPY version.sh .
 COPY --exclude=*.sh --exclude=*.md --exclude=*.yml . .
+ARG NODE_VERSION
+ARG NODE_COMMIT
```
fbac, line 23: nit: is there are plan to use these vars inside Dockerfile?

fbac (approved):

swift1337 (approved):

Dockerfile-localnet
gartnera: Yeah updated

lumtis (approved):

PR #2958, opened 2024-10-02
docs(rfc): protocol contract arbitrary call interface proposal

lumtis:
# Description

List possible alternative for arbitrary call interface.


lumtis:
Closing it as the proposal didn't end up in a clear solution

PR #2999, opened 2024-10-11
fix(emissions): add nil checks for reward distributions

lumtis:
# Description

- Add nil check for the reward per block params, panic here would halt the chain
- Add nil checks in the params validation, originally it panicked at this stage as well
- Skip reward distribution logic if reward per block is 0

## Summary by CodeRabbit

- **New Features**
	- Introduced a new helper function for improved logging during reward distribution.
  
- **Bug Fixes**
	- Enhanced error handling and validation for reward distribution and parameter checks.
  
- **Documentation**
	- Updated test cases and error messages for better clarity and coverage.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -224,7 +224,7 @@ var (
 		fungibletypes.ModuleName:                        {authtypes.Minter, authtypes.Burner},
 		emissionstypes.ModuleName:                       nil,
 		emissionstypes.UndistributedObserverRewardsPool: nil,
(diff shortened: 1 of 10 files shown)


lumtis:
> > ... panic here would halt the chain
> 
> Do we have a wrapper with `defer` for panic recovery somewhere?

Nope, I'm not sure we should had a general wrapper, in some case a bug where a panic happen could reflect a deeper issue where it's not safe to let the chain continue producing blocks

kingpinXD (approved):

Issue #2880, opened 2024-09-14
Defining new standard for BTC inbound data


lumtis:
**Describe the Issue**
With inscription support we now have ability to provide inputs up to 4MB in the inbound, this gives flexibility for more complete inbound structure.

This issue is about to define a standard on how to parse the inbound data for a Bitcoin deposit.

Example:
- Explicitly defining if the function is a `deposit` or `depositAndCall`
- Allowing to provide `revertOptions`, choosing for example the address on ZetaChain that should receive funds in case of cctx abort

lumtis:
For Solana we are splitting the memo to represent receiver and message separately but the idea of using a standard among all chains adapted to EVM is interesting

bitSmiley:
Hi all, just from testing inscription based memo on testnet sharing a few thoughts here.

This feature is quite critical for inscription based memo, because(long story short) inscription happens in two btc transactions, the current revert txn will send funds back to the "tapscript" address instead of the correct sender. Funds are difficult to retrieve. I think this might happen, probably more edge cases, to `OP_RETURN` as well because there could be multiple inputs to the funding transaction, currently only the first one is picked as revert address.

From trying on BTC inscription, kind of feel it's ok to introduce some kind of envelop to memo, such as prepend the envelop to the memo:
```
[Envelop][Memo]
```
Actually the current structure of `memo` has the above, such as `[address][calldata]`. At least for inscription based envelop, the following envelop could be considered?
```
[Version][Envelop Details]
```
The `version` is just a single byte(u8) or two bytes(u16) indicating the version of the payload, for better upgradability. Then envelop details could be whatever bytes/protocol for the corresponding version.

For our immediate use case, it would be great maybe call it as version 0, with envelop details such as:
```
[Header][Optional Field 1][Optional Field 2][RecivingAddress]
```
The `header` is just a single byte(u8) and each bit represents if the optional fields are included for interpretation. Some optional fields includes:
- revert address

ws4charlie:
@bitSmiley Thanks for your providing the thoughts and feedback during the inscription integration. We're going to work on this and figure out a consistent (across supported chains) way to enable the options on inbound txs.

ws4charlie:
Here is a proposed [standard format of non-EVM chain inbound memo](https://www.notion.so/zetachain/Standard-non-EVM-Inbound-Memo-10e7b3fb904980ed8ae6d7330d79839e). Please let me know your thoughts @fadeev, cc @lumtis and @bitSmiley

[...]

Issue #2149, opened 2024-05-09
fix: the outbound receiver address is wrong for deposits from Bitcoin


ws4charlie:
**Describe the Issue**
Here is an example of inbound cctx: https://explorer.zetachain.com/cc/tx/0x000abb86711bd720ac2eb5cc00cdd602ad56a2c21f8f64e7a21bec781f13b241
The receiver shown in ZetaScan is an irrelevant address:`0xe9d6b4Db5EDFD9e6b779AEdA6b5dbc6dCEFCEFCe`
![image](https://github.com/zeta-chain/node/assets/34498985/51936b05-aef8-4ac2-b7b4-5972b1c0fe17)

The actual intended receiver is `0x6dA0217f2eA3ea7aA128bc7878507EdbD3847bBc` in the Blockscout: https://zetachain.blockscout.com/tx/0xd52937ea0ae881b82123e676a9b575f96286f1d7a938c6100f0ec4a3d967a730
![image](https://github.com/zeta-chain/node/assets/34498985/4b97ea7a-6bec-4923-a1c1-e4087d402bfc)

**Expected Outcome**
A description of what outcome you are seeking.

PR #3025, opened 2024-10-21
feat: integrate standard memo to Bitcoin inbound and E2E tests

ws4charlie:
# Description

Complete: 
- [x] add standard memo support for Bitcoin inbound.
- [x] add E2E tests for `deposit`, `deposit_and_call`, `deposit_and_call_revert`, etc.
- [x] fix the incorrect receiver address (which is currently a btc address) with actual receiver parsed from memo.
- [x] unit tests
- [x] feature is disabled for Bitcoin mainnet

Closes: https://github.com/zeta-chain/node/issues/2880
Closes: https://github.com/zeta-chain/node/issues/2149

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Added support for stateful precompiled contracts and a staking precompiled contract.
	- Introduced a common importable `zetacored` RPC package.
	- Enhanced Bitcoin inscription support on testnet.
	- New functionality for handling Bitcoin inbound events, including compliance checks and memo decoding.

- **Bug Fixes**
	- Resolved issues with operator voting on discarded keygen ballots and improved outbound tracker functionality.

- **Tests**
	- Expanded testing coverage with new end-to-end tests for Bitcoin deposits and stateful precompiled contracts.
	- Added tests for handling various Bitcoin transaction scenarios, including donations and standard memo deposits.

- **Documentation**
	- Updated changelog to reflect all new features, refactoring, and fixes.


e2e/e2etests/test_bitcoin_deposit_and_call_revert.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zeta-chain/node/e2e/runner"
+	"github.com/zeta-chain/node/e2e/utils"
+	"github.com/zeta-chain/node/testutil/sample"
+	zetabitcoin "github.com/zeta-chain/node/zetaclient/chains/bitcoin"
+)
+
+func TestBitcoinDepositAndCallRevert(r *runner.E2ERunner, args []string) {
```
lumtis, line 13: Should be rather called `TestBitcoinDepositAndRevert` as we don't call anything on Bitcoin

pkg/memo/memo.go
```diff
@@ -51,7 +51,7 @@ func (m *InboundMemo) EncodeToBytes() ([]byte, error) {
 
 // DecodeFromBytes decodes a InboundMemo struct from raw bytes
 //
-// Returns an error if given data is not a valid memo
+// Returns nil memo if given data can't be decoded as a memo.
```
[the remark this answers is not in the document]
lumtis: If this has to be processed differently we could also reutrn a parsing error to be handled by the calling function, not sure returning an inboundMemo as a pointer makes the API simpler

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+// DecodeEventMemoBytes decodes the memo bytes as either standard or legacy memo
+func (ob *Observer) DecodeEventMemoBytes(event *BTCInboundEvent) error {
+	var (
+		err      error
+		memoStd  *memo.InboundMemo
+		receiver ethcommon.Address
+	)
+
+	// skip decoding donation tx as it won't go through zetacore
+	if bytes.Equal(event.MemoBytes, []byte(constant.DonationMessage)) {
```
lumtis, line 100: We shouldn't need this call here if we already have this check in `Processability`

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil
+	}
+
+	// try to decode the standard memo as the preferred format
+	// the standard memo is NOT enabled for Bitcoin mainnet
+	if ob.Chain().ChainId != chains.BitcoinMainnet.ChainId {
+		memoStd, err = memo.DecodeFromBytes(event.MemoBytes)
```
lumtis, line 107: If it is mainnet, we here skip this variable initialization but still goes to the switch reading this variable.

This flow should be simplified. If it's mainnet, we should have some logic and return in the if body

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	switch {
+	case memoStd != nil:
+		// skip memo that carries improper data
+		if err != nil {
```
lumtis, line 113: Confusing, why not handling the error before

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+			return errors.Wrap(err, "standard memo contains improper data")
+		}
+
+		// NoAssetCall will be disabled for Bitcoin until full V2 support
```
lumtis, line 117:
```suggestion
		// NoAssetCall will be disabled for Bitcoin until full V2 support
		// https://github.com/zeta-chain/node/issues/2711
```

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	if crypto.IsEmptyAddress(receiver) {
+		return errors.New("got empty receiver address from memo")
+	}
+	event.ToAddress = receiver.Hex()
+
+	return nil
+}
+
+// CheckEventProcessability checks if the inbound event is processable
+func (ob *Observer) CheckEventProcessability(event *BTCInboundEvent) bool {
```
lumtis, line 154:
```suggestion
func (ob *Observer) CheckEventProcessability(event BTCInboundEvent) bool {
```
Let's stop using pointers when we don't need those

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	// check if the event is processable
+	result := event.CheckProcessability()
+	switch result {
```
lumtis, line 157:
```suggestion
	switch result := event.CheckProcessability(); result {
```

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		ob.Logger().Inbound.Info().Fields(logFields).Msgf("thank you rich folk for your donation!")
+		return false
+	case InboundProcessabilityComplianceViolation:
+		compliance.PrintComplianceLog(ob.logger.Inbound, ob.logger.Compliance,
+			false, ob.Chain().ChainId, event.TxHash, event.FromAddress, event.ToAddress, "BTC")
+		return false
+	}
+
+	return true
```
lumtis, line 173: Should return false if we have an unexpected value, should be covered by test as well

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// NewInboundVoteV1 creates a MsgVoteInbound message for inbound that uses legacy memo
+func (ob *Observer) NewInboundVoteV1(event *BTCInboundEvent, amountSats *big.Int) *crosschaintypes.MsgVoteInbound {
```
lumtis, line 177:
```suggestion
func (ob *Observer) NewInboundVoteFromLegacyMemo(event *BTCInboundEvent, amountSats *big.Int) *crosschaintypes.MsgVoteInbound {
```

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+		0,
+		crosschaintypes.ProtocolContractVersion_V1,
+		false, // not relevant for v1
+	)
+}
+
+// NewInboundVoteMemoStd creates a MsgVoteInbound message for inbound that uses standard memo
+// TODO: rename this function as 'EventToInboundVoteV2' to use ProtocolContractVersion_V2 and enable more options
+// https://github.com/zeta-chain/node/issues/2711
+func (ob *Observer) NewInboundVoteMemoStd(event *BTCInboundEvent, amountSats *big.Int) *crosschaintypes.MsgVoteInbound {
```
lumtis, line 203:
```suggestion
func (ob *Observer) NewInboundVoteFromStdMemo(event *BTCInboundEvent, amountSats *big.Int) *crosschaintypes.MsgVoteInbound {
```

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+		coin.CoinType_Gas,
+		"",
+		0,
+		crosschaintypes.ProtocolContractVersion_V1,
+		false, // not relevant for v1
+	)
+}
+
+// NewInboundVoteMemoStd creates a MsgVoteInbound message for inbound that uses standard memo
+// TODO: rename this function as 'EventToInboundVoteV2' to use ProtocolContractVersion_V2 and enable more options
```
lumtis, line 201: > TODO: rename this function as 'EventToInboundVoteV2'
No this would simply be the default logic

zetaclient/zetacore/constant.go
```diff
@@ -24,6 +24,9 @@ const (
 	// PostVoteInboundMessagePassingExecutionGasLimit is the gas limit for voting on, and executing ,observed inbound tx related to message passing (coin_type == zeta)
 	PostVoteInboundMessagePassingExecutionGasLimit = 4_000_000
 
+	// PostVoteInboundCallOptionsGasLimit is the gas limit for inbound call options
+	PostVoteInboundCallOptionsGasLimit uint64 = 1_500_000
```
lumtis, line 28: What is this, why is it needed?

lumtis (commented):
Need some improvements in `event.go`

e2e/utils/zetacore.go
```diff
@@ -204,6 +204,27 @@ type waitConfig struct {
...
 
+// WaitCctxRevertedByInboundHash waits until cctx is reverted by inbound hash.
+func WaitCctxRevertedByInboundHash(
+	ctx context.Context,
+	t require.TestingT,
+	hash string,
+	c CCTXClient,
+) crosschaintypes.CrossChainTx {
+	// criteria for reverted cctx
+	searchForCCTXReverted := Matches(func(tx crosschaintypes.CrossChainTx) bool {
```
swift1337, line 215: Can we use `MatchStatus()` ?

pkg/memo/memo.go
```diff
@@ -64,15 +64,20 @@ func DecodeFromBytes(data []byte) (*InboundMemo, error) {
...
+		err = memo.FieldsV0.Validate(memo.OpCode, memo.Header.DataFlags)
 	default:
 		return nil, fmt.Errorf("invalid memo version: %d", memo.Version)
 	}
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to unpack memo fields version: %d", memo.Version)
-	}
 
-	return memo, nil
+	return memo, err
```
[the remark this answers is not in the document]
ws4charlie: it's by design

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	if ob.Chain().ChainId != chains.BitcoinMainnet.ChainId {
+		memoStd, err = memo.DecodeFromBytes(event.MemoBytes)
+	}
+
+	switch {
+	case memoStd != nil:
+		// skip memo that carries improper data
+		if err != nil {
+			return errors.Wrap(err, "standard memo contains improper data")
+		}
```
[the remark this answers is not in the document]
ws4charlie: a non-Nil memoStd is to tell the caller that the bytes are still well formatted as a memo

e2e/e2etests/test_bitcoin_deposit_and_call_revert.go
ws4charlie: But remove the `Call` doesn't express the operation code of `deposit_and_call`. Or do you like `TestBitcoinDepositCallAndRevert`?

e2e/e2etests/test_bitcoin_donation.go
```diff
@@ -0,0 +1,44 @@
...
+	require.NotEmpty(r, utxos)
+
+	// ACT
+	// Send BTC to TSS address with donation message
+	memo := []byte(constant.DonationMessage)
+	txHash, err := r.SendToTSSFromDeployerWithMemo(amountTotal, utxos, memo)
+	require.NoError(r, err)
+
+	// ASSERT after 20 seconds
+	time.Sleep(time.Second * 20)
```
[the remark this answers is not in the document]
ws4charlie: The donation tx currently triggers nothing. Zetaclient only print a log on donation. This is a indirectly way to make sure nothing has happened. We have a group of Bitcoin E2E tests to run, what I can do is to move this donation as the 1st test so to avoid wasting time. wdyt?

pkg/memo/memo.go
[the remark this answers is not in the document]
ws4charlie: We'll have to support both legacy memo and standard memo at some point. Zetaclient has to try deserializing given bytes as a standard memo struct. If serialization fails, zetaclient knows that the given bytes is not a standard memo and continue processing the bytes as legacy memo.

**The problem:**
       If the given bytes is well formatted in the standard way but still carry improper values (e.g., invalid address), we SHOULD NOT continue and process it as a legacy memo. So we kinda need an extra information to tell: `are these bytes a standard memo or not ?`.
       
**The way to solve:**
      Before phasing out legacy memo. Let the API `DecodeFromBytes` to return a hint (nil or not nil) to tell `whether the data is a standard memo or not`, otherwise we'll need a separate API `IsStandardMemo(memoBytes []byte) bool` to do so.
     
   After phasing out legacy, let the API `DecodeFromBytes` to return `nil` on error.

pkg/memo/memo.go
```diff
@@ -64,15 +64,20 @@ func DecodeFromBytes(data []byte) (*InboundMemo, error) {
 	// decode fields based on version
 	switch memo.Version {
 	case 0:
```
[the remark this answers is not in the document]
ws4charlie: can do

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// BlockNumber is the block number of the inbound
+	BlockNumber uint64
+
+	// TxHash is the hash of the inbound
+	TxHash string
+}
+
+// IsProcessable checks if the inbound event is processable
+func (event *BTCInboundEvent) CheckProcessability() InboundProcessability {
```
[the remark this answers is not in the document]
ws4charlie: I have no preference except that the pointer receiver can avoid the copying when the method is called. wdyt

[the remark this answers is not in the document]
ws4charlie: makes sense

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	// donation check
+	if bytes.Equal(event.MemoBytes, []byte(constant.DonationMessage)) {
+		return InboundProcessabilityDonation
+	}
+
+	return InboundProcessabilityGood
+}
+
+// DecodeEventMemoBytes decodes the memo bytes as either standard or legacy memo
+func (ob *Observer) DecodeEventMemoBytes(event *BTCInboundEvent) error {
```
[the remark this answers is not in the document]
ws4charlie: Good catch for improvement. It should be more OOP. 
I'm going to make `DecodeEventMemoBytes` a method of the event struct self. We'll have to update two fields of the event after decoding: `event.MemoStd`, `event.ToAddress`.
`func (event *BTCInboundEvent) DecodeMemoBytes(chainID int64) error`

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: It's a good point and it's by intention. Here is the workflow:
step-1: call `DecodeMemoBytes` so that we know the `receiver`, `revert address`, `abort address` etc.
step-2: call `CheckEventProcessability` against decoded addresses to not violate the compliance.


If the early on `DonationMessage` gets removed, the followed logic will try decoding it as a legacy memo and then fails because the message `I am rich!` is too short to be valid(a receiver address required and memo must be` >= 20 bytes`

```
// decode event memo bytes
err := event.DecodeMemoBytes(ob.Chain().ChainId)
if err != nil {
	return nil
}

// check if the event is processable
if !ob.CheckEventProcessability(event) {
	return nil
}
```

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: The edge case:

A user serialize below memo (e.g. invalid receiver or revert address) and wrap it into his/her transaction. The memo deserialization will pass without issue. In this case, we should still `reject` it (if err != nil) and stop trying decode it as a legacy memo. The complication is caused by supporting both type of memos as I mentioned in one of above comments.

```
memo := &memo.InboundMemo{
    Header: memo.Header{
	Version:     0,
	EncodingFmt: memo.EncodingFmtCompactShort,
	OpCode:      memo.OpCodeDepositAndCall,
    },
    FieldsV0: memo.FieldsV0{
	Receiver: common.Address{}, // invalid address
	Payload:  []byte("a payload"),
    },
}
```

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: will add a default as an unreachable path

zetaclient/zetacore/constant.go
ws4charlie: I moved the magic number from `evm/observer/v2_inbound.go` to here as a constant, which can then be reused by Bitcoin as well (in V2). wdyt

e2e/utils/zetacore.go
ws4charlie: Good point. I'll wrap this into a `MatchReverted` to check reverted cctx strictly.

ws4charlie: https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: updated comment: https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: renamed https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: renamed https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

pkg/memo/memo.go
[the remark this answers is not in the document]
ws4charlie: added constant https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,233 @@
...
+	MemoStd *memo.InboundMemo
+
+	// BlockNumber is the block number of the inbound
+	BlockNumber uint64
+
+	// TxHash is the hash of the inbound
+	TxHash string
+}
+
+// IsProcessable checks if the inbound event is processable
```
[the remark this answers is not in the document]
ws4charlie: updated https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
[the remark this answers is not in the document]
ws4charlie: renamed in https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: updated in https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: updated in https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: updated in https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: default path added in https://github.com/zeta-chain/node/pull/3025/commits/7a2880bf21e47dfa10d0f3a9dd3756064237afbd

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: Those variables will be initialized as nil or empty value, and this is done intentionally.

e2e/e2etests/test_bitcoin_donation.go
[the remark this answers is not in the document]
ws4charlie: I moved as the 1st test in Bitcoin group in commit: https://github.com/zeta-chain/node/pull/3025/commits/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

pkg/memo/memo.go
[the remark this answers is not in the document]
ws4charlie: To make the logic clear, here is the refactor: https://github.com/zeta-chain/node/pull/3025/commits/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

zetaclient/chains/bitcoin/observer/event.go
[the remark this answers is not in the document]
ws4charlie: This is updated https://github.com/zeta-chain/node/pull/3025/commits/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

zetaclient/chains/bitcoin/observer/event.go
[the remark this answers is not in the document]
ws4charlie: in same refactor https://github.com/zeta-chain/node/commit/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: simplified in same refactor https://github.com/zeta-chain/node/commit/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: Introduced a variable `isStandardMemo` to made the flow clear in the same refactor https://github.com/zeta-chain/node/commit/b2c149c4124f27c83f2ef6f4a55f76f8c3534433

lumtis:
Tested by disabling the mainnet condition (like mainnet is used for E2E tests), this looks good
```
bitcoin    | ⏳running - deposit Bitcoin into ZEVM
bitcoin    | ✅ completed in 6.038452044s - deposit Bitcoin into ZEVM
bitcoin    | ⏳running - deposit Bitcoin into ZEVM and call a contract
bitcoin    | ✅ completed in 4.033241877s - deposit Bitcoin into ZEVM and call a contract
bitcoin    | ⏳running - deposit Bitcoin into ZEVM; expect refund
bitcoin    | ✅ completed in 20.782137884s - deposit Bitcoin into ZEVM; expect refund
```

e2e/runner/bitcoin.go
```diff
@@ -366,6 +391,25 @@ func (r *E2ERunner) GenerateToAddressIfLocalBitcoin(
 	return nil, nil
 }
 
+// QueryOutboundReceiverAndAmount queries the outbound receiver and amount (in satoshis) from the given txid
+func (r *E2ERunner) QueryOutboundReceiverAndAmount(txid string) (string, int64) {
```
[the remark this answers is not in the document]
ws4charlie: The wait is done by `utils.WaitCctx....`, so there is no wait in here any more.

e2e/runner/bitcoin.go
```diff
@@ -366,6 +391,25 @@ func (r *E2ERunner) GenerateToAddressIfLocalBitcoin(
...
+	txHash, err := chainhash.NewHashFromStr(txid)
+	require.NoError(r, err)
+
+	// query outbound raw transaction
+	revertTx, err := r.BtcRPCClient.GetRawTransaction(txHash)
+	require.NoError(r, err, revertTx)
+	require.True(r, len(revertTx.MsgTx().TxOut) >= 2, "bitcoin outbound must have at least two outputs")
+
+	// parse receiver address from pkScript
+	txOutput := revertTx.MsgTx().TxOut[1]
```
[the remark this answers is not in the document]
ws4charlie: no, the TxOut[0] is TSS

pkg/memo/fields_v0.go
```diff
@@ -54,18 +54,13 @@ func (f *FieldsV0) Pack(opCode OpCode, encodingFmt EncodingFormat, dataFlags uin
 }
 
 // Unpack decodes the memo fields
-func (f *FieldsV0) Unpack(opCode OpCode, encodingFmt EncodingFormat, dataFlags uint8, data []byte) error {
+func (f *FieldsV0) Unpack(encodingFmt EncodingFormat, dataFlags uint8, data []byte) error {
```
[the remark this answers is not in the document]
ws4charlie: This unpacking only works on new version. The old version is the `DecodeLegacyMemoHex`.

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,253 @@
...
+		err = ValidateStandardMemo(*memoStd, chainID)
+		if err != nil {
+			return errors.Wrap(err, "invalid standard memo for bitcoin")
+		}
+
+		event.MemoStd = memoStd
+		receiver = memoStd.Receiver
+	} else {
+		parsedAddress, _, err := memo.DecodeLegacyMemoHex(hex.EncodeToString(event.MemoBytes))
+		if err != nil { // unreachable code
```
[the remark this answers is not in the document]
ws4charlie: The `DecodeLegacyMemoHex` returns error only when the argument is a invalid hex string, but the argument passed here is guaranteed to be a valid hex string.

e2e/runner/bitcoin.go
```diff
@@ -168,13 +175,31 @@ func (r *E2ERunner) DepositBTC() {
 	require.Equal(r, 1, balance.Sign(), "balance should be positive")
 }
 
-func (r *E2ERunner) SendToTSSFromDeployerToDeposit(amount float64, inputUTXOs []btcjson.ListUnspentResult) (
-	*chainhash.Hash,
-	error,
-) {
+// DepositBTCWithLegacyMemo deposits BTC from the deployer address to the TSS using legacy memo
```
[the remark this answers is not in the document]
ws4charlie: added https://github.com/zeta-chain/node/pull/3025/commits/7e6368f3ed07f0a3afaa55f2270a65a0da79746a

e2e/runner/bitcoin.go
```diff
@@ -76,10 +77,8 @@ func (r *E2ERunner) GetTop20UTXOsForTssAddress() ([]btcjson.ListUnspentResult, e
 	return utxos, nil
 }
 
-// DepositBTCWithAmount deposits BTC on ZetaChain with a specific amount
-func (r *E2ERunner) DepositBTCWithAmount(amount float64) *chainhash.Hash {
-	r.Logger.Print("⏳ depositing BTC into ZEVM")
-
+// DepositBTCWithAmount deposits BTC into ZetaChain with a specific amount
```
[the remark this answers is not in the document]
ws4charlie: updated https://github.com/zeta-chain/node/pull/3025/commits/7e6368f3ed07f0a3afaa55f2270a65a0da79746a

pkg/memo/fields_v0.go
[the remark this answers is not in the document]
ws4charlie: I added a unit test to ensure this feature is disabled on mainnet: https://github.com/zeta-chain/node/pull/3025/commits/7e6368f3ed07f0a3afaa55f2270a65a0da79746a

lumtis (approved):

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// BlockNumber is the block number of the inbound
+	BlockNumber uint64
+
+	// TxHash is the hash of the inbound
+	TxHash string
+}
+
+// Processability returns the processability of the inbound event
+func (event BTCInboundEvent) Processability() InboundProcessability {
```
fbac, line 66: nit: even when we're not modifying anything in `event` I'd suggest to use it as a pointer receiver, to maintain consistency.

zetaclient/chains/bitcoin/observer/event.go
fbac: Why would `DonationMessage` gets removed, on which step?

zetaclient/chains/bitcoin/observer/event.go
fbac: Is there anycase where `isStandardMemo == false` and `err != nil`? I'm failing here to see why we shouldn't simply check for errors just after DecodeFromBytes.

zetaclient/chains/bitcoin/observer/event.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+
+		event.MemoStd = memoStd
+		receiver = memoStd.Receiver
+	} else {
+		parsedAddress, _, err := memo.DecodeLegacyMemoHex(hex.EncodeToString(event.MemoBytes))
+		if err != nil { // unreachable code
+			return errors.Wrap(err, "invalid legacy memo")
+		}
+		receiver = parsedAddress
```
fbac, line 132: If it's unreachable couldn't we simply do:
```suggestion
		receiver, _, _ = memo.DecodeLegacyMemoHex(hex.EncodeToString(event.MemoBytes))
```

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -169,27 +166,49 @@ func DecodeScriptP2PKH(scriptHex string, net *chaincfg.Params) (string, error) {
...
-		if int(memoSize) != (len(scriptHex)-4)/2 {
-			return nil, false, fmt.Errorf("memo size mismatch: %d != %d", memoSize, (len(scriptHex)-4)/2)
+		memoBytes = scriptBytes[2:]
+	case memoSize == txscript.OP_PUSHDATA1:
+		// when data size >= OP_PUSHDATA1 (76), Bitcoin uses 2 bytes to represent the length: [OP_PUSHDATA1 + dataLen]
+		// see: https://github.com/btcsuite/btcd/blob/master/txscript/scriptbuilder.go#L183
+		if len(scriptBytes) < 3 {
+			return nil, false, fmt.Errorf("script too short: %s", scriptHex)
 		}
+		memoSize = scriptBytes[2]
```
fbac, line 198: nit: for my own education, on this case shouldn't memoSize be scriptBytes[1:2], that is the bytes 1 and 2, since btc will use 2 bytes to represent the length.

pkg/memo/memo.go
```diff
@@ -51,28 +56,41 @@ func (m *InboundMemo) EncodeToBytes() ([]byte, error) {
 
 // DecodeFromBytes decodes a InboundMemo struct from raw bytes
 //
-// Returns an error if given data is not a valid memo
-func DecodeFromBytes(data []byte) (*InboundMemo, error) {
+// Returns:
+//   - [memo, true, nil] if given data is successfully decoded as a memo.
+//   - [nil,  true, err] if given data is successfully decoded as a memo but contains improper field values.
+//   - [nil, false, err] if given data can't be decoded as a memo.
```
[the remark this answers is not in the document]
ws4charlie: That will work as well. Here is a minimum change which will be reverted once legacy memo phases out.

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: It used to be a point receiver (I just want to avoid copy). Seems we three of us (Lol!) had different ideas on this line. I have no preference except avoid copying.

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: I see what you mean. I won't assume that the implementation of `DecodeLegacyMemoHex` will never change in future.

zetaclient/chains/bitcoin/tx_script.go
ws4charlie: This part is a trick. They fill in the 2 bytes in format `[OP_PUSHDATA1][dataLen]`, in which the 1st byte is served as a flag. When flag is set to `OP_PUSHDATA1`, we need to take the 2nd byte only. Their code is [here](https://github.com/btcsuite/btcd/blob/master/txscript/scriptbuilder.go#L183)

ws4charlie:
@fbac 

> Why would DonationMessage gets removed, on which step?
It's skipped in [here](https://github.com/zeta-chain/node/blob/feat-btc-revert-address/zetaclient/chains/bitcoin/observer/event.go#L173)

> Is there anycase where isStandardMemo == false and err != nil? I'm failing here to see why we shouldn't simply check for errors just after DecodeFromBytes.
No. When `isStandardMemo == false`, we simply overwrite the `err` by whatever returned from `DecodeLegacyMemoHex`

x/crosschain/types/message_vote_inbound.go
```diff
@@ -22,6 +22,13 @@ const MaxMessageLength = 10240
 // InboundVoteOption is a function that sets some option on the inbound vote message
 type InboundVoteOption func(*MsgVoteInbound)
 
+// WithMemoRevertOptions sets the revert options for inbound vote message
+func WithRevertOptions(revertOptions RevertOptions) InboundVoteOption {
+	return func(msg *MsgVoteInbound) {
+		msg.RevertOptions = revertOptions
+	}
```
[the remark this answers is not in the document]
ws4charlie: We can leave it as consistent for now and it will never be nill because the usage is very restricted. See [the usage](https://github.com/zeta-chain/node/blob/feat-btc-revert-address/x/crosschain/types/inbound_parsing.go#L146)

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: You and me vote on pointer, so here I reverted to use pointer: https://github.com/zeta-chain/node/pull/3025/commits/d0e7b08878d1c5a502a3ab487795538161205188

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: [removed on this step](https://github.com/zeta-chain/node/blob/feat-btc-revert-address/zetaclient/chains/bitcoin/observer/event.go#L173)

zetaclient/chains/bitcoin/observer/event.go
ws4charlie: No. When `isStandardMemo == false`, the `err` will simply be ignored and overwritten by whatever returned from `DecodeLegacyMemoHex`.

fbac (approved):

PR #3028, opened 2024-10-22
feat: whitelist connection gater

skosito:
# Description

go-tss PR https://github.com/zeta-chain/go-tss/pull/31

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Release Notes for Version 21.0.0

- **New Features**
  - Support for stateful precompiled contracts and a staking precompiled contract.
  - Integration of a common importable `zetacored` RPC package.
  - Enhanced Bitcoin chain support with static information for various testnets.
  - Added ability to derive Bitcoin TSS addresses by chain ID.
  - Introduced a `whitelist connection gater` for improved peer management.

- **Bug Fixes**
  - Resolved issues with outbound tracker blocking confirmations on EVM chains.
  - Allowed operators to vote on ballots associated with discarded keygen.

- **Tests**
  - Expanded test coverage for stateful precompiled contracts and staking precompiles.
  - Enhanced testing for the `AppContext` and orchestrator functionalities.

- **Documentation**
  - Updated changelog to reflect all changes and enhancements in this version.


cmd/zetaclientd/start.go
```diff
@@ -157,6 +160,25 @@ func start(_ *cobra.Command, _ []string) error {
 
 	go zetacoreClient.UpdateAppContextWorker(ctx, appContext)
 
+	keygen := appContext.GetKeygen()
```
gartnera, line 163: We should add `c.GetAllNodeAccounts()` to `UpdateAppContext()` and use that instead. That will also work during keygen.

cmd/zetaclientd/start.go
```diff
@@ -157,6 +160,25 @@ func start(_ *cobra.Command, _ []string) error {
...
+		bz := pk.Bytes()
+		k, err := crypto2.UnmarshalSecp256k1PublicKey(bz)
+		if err != nil {
+			return err
+		}
+		pid, err := peer.IDFromPublicKey(k)
+		if err != nil {
+			return err
+		}
+		cfg.WhitelistedPeers = append(cfg.WhitelistedPeers, pid.String())
```
gartnera, line 179: maybe store this on the go-tss side (under conversion) so we can use it in go-tss tests.

ws4charlie (approved):

kingpinXD (approved):

gartnera:
Let's make sure we use the final merged version of https://github.com/zeta-chain/go-tss/pull/31 in go.mod

gartnera (approved):

brewmaster012 (approved):
Looks good except I'm not 100% sure about how it works during TSS migration (new Keygen, NodeAccounts and the Keygen object, which one is more truthful? Would they diverge? ). 

Maybe we run the TSS migration test? Is there a new keygen there?

zetaclient/context/app.go
```diff
@@ -120,6 +125,27 @@ func (a *AppContext) GetKeygen() observertypes.Keygen {
 	}
 }
 
+func (a *AppContext) GetNodeAccounts() []observertypes.NodeAccount {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+
+	// deep copy node accounts
+	nodeAccounts := []observertypes.NodeAccount{}
```
swift1337, line 133: Can you elaborate on why a deep copy is required? 

Also, I think we need to implement `a.updateNodeAccounts()` that takes new accounts, sorts them, compares them with the current slice, and changes only if necessary (deterministically). In 99.99% of times it will be a `no-op`.

zetaclient/context/app.go
```diff
@@ -173,6 +202,7 @@ func (a *AppContext) Update(
 	a.crosschainFlags = crosschainFlags
 	a.keygen = keygen
 	a.currentTssPubKey = tssPubKey
+	a.nodeAccounts = nodeAccounts
```
swift1337, line 205:
```suggestion
	a.updateNodeAccounts(nodeAccounts)
```

kingpinXD:
> ew Keygen, NodeAccounts and the Keygen object, which one is more truthful? Would they

Yes, the TSS migration test does a new keygen , using the node accounts.
A keygen is set using the existing set of node accounts, so adding the label should be enough to test this imo

kingpinXD (approved):

cmd/zetaclientd/start.go
kingpinXD: I think we should still use the Keygen object, as the node account list can be modified when we add an observer.

The newly added observer should not be allowed into the whitelist unless they are part of a tss ceremony (i.e. the keygen object )

cmd/zetaclientd/start.go
```diff
@@ -204,8 +206,19 @@ func start(_ *cobra.Command, _ []string) error {
 	}
 
 	telemetryServer.SetIPAddress(cfg.PublicIP)
+
+	nodeAccounts := appContext.GetNodeAccounts()
+	whitelistedPeers := []peer.ID{}
+	for _, nodeAccount := range nodeAccounts {
+		pid, err := conversion.Bech32PubkeyToPeerID(nodeAccount.GranteePubkey.Secp256k1.String())
+		if err != nil {
+			return err
```
fbac, line 215: nit/question: should we return an error here or simply `continue`? Do we need all the `nodeAccounts` to start the chain?

fbac (approved):

cmd/zetaclientd/start.go
kingpinXD: I agree , and I think that list should be the keygen List ,and not the nodeAccount list .
Becuase of the fact that the nodeAccount list might contain accounts which are not part of TSS yet. 

We might need to track this in a separate issue if we need to modify the keysign list as well .

PR #3031, opened 2024-10-22
chore: upgrade go-ethereum

gartnera:
# Description

See https://github.com/zeta-chain/ethermint/pull/141 for a detailed description of all go-ethereum changes.

Changes only effecting node:
- tolerate `eth.Address{}.Hash()` removal ([ref](https://github.com/ethereum/go-ethereum/pull/28228))
- fix panic in `hasher.Reset()` (use `trie.NewEmpty(nil)`) instead

TODO:

- [x] merge https://github.com/zeta-chain/ethermint/pull/141

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Release Notes

- **New Features**
  - Enhanced logging capabilities with a new structured logging handler.
  - Updated transaction signing to utilize Ethermint's types for better integration.

- **Bug Fixes**
  - Improved error handling in contract interactions and tests, ensuring accurate responses to various scenarios.

- **Refactor**
  - Simplified message handling in multiple components by switching from value to pointer types for `core.Message`.
  - Transitioned from `big.Int` to `uint256` for handling large integers across various modules, improving performance and consistency.

- **Documentation**
  - Updated method signatures and parameters to reflect changes in data types and logging practices.


x/fungible/keeper/evm.go
```diff
@@ -735,19 +736,19 @@ func (k Keeper) CallEVMWithData(
...
-		value,         // amount
-		gasCap,        // gasLimit
-		big.NewInt(0), // gasFeeCap
-		big.NewInt(0), // gasTipCap
-		big.NewInt(0), // gasPrice
-		data,
-		ethtypes.AccessList{}, // AccessList
-		!commit,               // isFake
-	)
+	msg := &core.Message{
```
swift1337, line 739: consider aliasing `core` as `ethcore`

swift1337 (approved):

fbac (approved):

server/json_rpc.go
```diff
@@ -27,11 +29,51 @@
...
+		g.logger.Debug(record.Message, attrs)
+	case slog.LevelInfo:
+		g.logger.Info(record.Message, attrs)
+	case slog.LevelWarn:
+		g.logger.Info(record.Message, attrs)
+	case slog.LevelError:
+		g.logger.Error(record.Message, attrs)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
gartnera: The tendermint logger does not have a warning level. This is the logic that this is replacing:
```
		case ethlog.LvlInfo, ethlog.LvlWarn:
			logger.Info(r.Msg, r.Ctx...)
```

pkg/proofs/ethereum/proof.go
```diff
@@ -160,18 +159,18 @@ func NewTrie(list types.DerivableList) Trie {
 		// #nosec G115 iterator
 		indexBuf = rlp.AppendUint64(indexBuf[:0], uint64(i))
 		value := encodeForDerive(list, i, valueBuf)
-		hasher.Update(indexBuf, value)
+		_ = hasher.Update(indexBuf, value)
```
kingpinXD, line 162: Consider logging the error , even if we don't need to handle it

gartnera: There is no logger here. The error condition is not reachable in this case. I think the least bad thing to do if we want to handle this is to just panic (since it really should never happen)

ws4charlie (approved):

ws4charlie:
Nice. Once we have this update, the Ethereum blob-tx can be handled in a native way.

lumtis (approved):

PR #3038, opened 2024-10-24
fix: remove additional log line when exporting genesis

kingpinXD:
# Description

- This pr removes a a log line that prints the list of blocked addresses when initializing the app.
- It does minor refactoring for the code organization in `appExport` function

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced clarity and maintainability in the initialization of application components within the ZetaChain framework.
	- Improved state export functionality for the genesis file.
  
- **Bug Fixes**
	- Streamlined error handling for loading application state.

- **Refactor**
	- Reorganized keeper initialization and module management for better structure and clarity.
	- Updated variable names and comments for improved understanding.


app/export.go
```diff
@@ -17,7 +17,7 @@ import (
 func (app *App) ExportAppStateAndValidators(
 	forZeroHeight bool, jailAllowedAddrs []string, modulesToExport []string,
 ) (servertypes.ExportedApp, error) {
-	// as if they could withdraw from the start of the next block
```
kingpinXD, line 20: I removed the comment as I am not sure what it is referring to

fbac (approved):

cmd/zetacored/root.go
```diff
@@ -268,48 +268,44 @@ func (ac appCreator) newApp(
...
-			map[int64]bool{},
-			homePath,
-			uint(1),
-			ac.encCfg,
-			appOpts,
-		)
-
-		if err := anApp.LoadHeight(height); err != nil {
+		err := zetaApp.LoadHeight(height)
+		if err != nil {
```
lumtis, line 305:
```suggestion
		if err := zetaApp.LoadHeight(height); err != nil {
```

cmd/zetacored/root.go
```diff
@@ -268,48 +268,44 @@ func (ac appCreator) newApp(
...
 ) (servertypes.ExportedApp, error) {
-	var anApp *app.App
+	var zetaApp *app.App
 
 	homePath, ok := appOpts.Get(flags.FlagHome).(string)
 	if !ok || homePath == "" {
 		return servertypes.ExportedApp{}, errors.New("application home not set")
 	}
 
+	loadLatest := false
```
lumtis, line 284: Why this change in this file? looks unrelated to the purpose of the PR?

kingpinXD: It does not change the logic , just restructures the code to make it more readable .
I just came across this when investigating , the change seems too small to be part of a separate pr . lmk if you think otherwise

app/app.go
```diff
@@ -372,8 +372,6 @@ func New(
 		authAddr,
 	)
 
-	logger.Info("bank keeper blocklist addresses", "addresses", app.BlockedAddrs())
-
```
gartnera, line 376: Shouldn't this be written to stderr? If not we should ensure the logger always writes to stderr.

kingpinXD: Why should we write informational log messages to stderr ? 
Also , is this log line used for anything ? We can consider exposing the list of BlockedAddreses via some other means if needed

gartnera: Typically you always write log messages to stderr to avoid situations just like this (where the command output is confused with log output)

It matters less for server processes but it matters a lot for cli tools

gartnera:
```
func NewDefaultContext() *Context {
	return NewContext(
		viper.New(),
		tmcfg.DefaultConfig(),
		tmlog.NewTMLogger(tmlog.NewSyncWriter(os.Stderr)),
	)
}
```
is where the logger is configured to use stdout. This logic is called in cosmos-sdk `server.GetServerContextFromCmd` so it would be tricky to adjust.

kingpinXD: Okay, I get what you are saying. It might be useful to switch over to StdErr, but that would be out of scope for this pr. I created an issue; please add any comments that you can think of to the issue directly.
https://github.com/zeta-chain/node/issues/3046

For this PR, unless any one is consuming the log line, I would like to remove it for now.

gartnera (approved):

cmd/zetacored/root.go
```diff
@@ -268,48 +268,44 @@ func (ac appCreator) newApp(
...
+		map[int64]bool{},
+		homePath,
+		uint(1),
+		ac.encCfg,
+		appOpts,
+	)
+
+	// If height is -1, it means we are using the latest height.
+	// For all other cases, we load the specified height from the Store
 	if height != -1 {
```
lumtis, line 303:
```suggestion
	if !loadLatest {
```

lumtis (changes requested):

lumtis (approved):

PR #3049, opened 2024-10-28
fix: move btc std memo tests to advanced tests

skosito:
# Description

Move bitcoin std tests to advanced so they are not run before upgrade

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **Bug Fixes**
	- Restructured test cases for Bitcoin memo deposits, improving organization and clarity in testing processes.
  
- **Documentation**
	- Updated comments for enhanced clarity on the purpose of code sections.

### DIFF
--- a/cmd/zetae2e/local/local.go
+++ b/cmd/zetae2e/local/local.go
@@ -302,10 +302,6 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 			e2etests.TestBitcoinDepositName,
 			e2etests.TestBitcoinDepositAndCallName,
 			e2etests.TestBitcoinDepositAndCallRevertName,
-			e2etests.TestBitcoinStdMemoDepositName,
-			e2etests.TestBitcoinStdMemoDepositAndCallName,
-			e2etests.TestBitcoinStdMemoDepositAndCallRevertName,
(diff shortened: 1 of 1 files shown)


swift1337 (changes requested):
Why would we move standard tests to the advanced tests category?

cmd/zetae2e/local/local.go
```diff
@@ -318,6 +314,10 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 			e2etests.TestBitcoinWithdrawP2SHName,
 			e2etests.TestBitcoinWithdrawP2WSHName,
 			e2etests.TestBitcoinWithdrawRestrictedName,
+			e2etests.TestBitcoinStdMemoDepositName,
```
swift1337, line 317: consider creating an issue for moving them back after the upgrade

swift1337 (approved):

cmd/zetae2e/local/local.go
lumtis: As mentioned on Slack, no need if we should push this for v21.
The test should pass once we set v21 as the reference for the upgrade tests

lumtis (approved):

fbac (approved):

PR #3113, opened 2024-11-06
test(`e2e`): add amount args for ERC20 withdraw tests

lumtis:
# Description

Add args

- `TestV2ERC20WithdrawAndCallNoMessageName` actually expected an arg and it was missing (the test was also not run)
- Add for `TestV2ERC20WithdrawAndCall`, despite the previous note, better to add an arg for consistency with other test, it doesn't add overhead as you have a default value and in the future we could whitelist a ERC20 with a high base price (unlikely)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new end-to-end test case for the ERC20 token withdrawal workflow.
	- Enhanced argument definitions for existing ERC20 withdrawal tests to include an "amount" parameter with a default value of 1000.

- **Bug Fixes**
	- Improved validation for the withdrawal amount in the ERC20 withdrawal test, ensuring proper error handling for invalid inputs.


e2e/e2etests/e2etests.go
```diff
@@ -955,13 +955,17 @@ var AllE2ETests = []runner.E2ETest{
 	runner.NewE2ETest(
 		TestV2ERC20WithdrawAndCallName,
 		"withdraw ERC20 from ZEVM and authenticated call a contract using V2 contract",
-		[]runner.ArgDefinition{},
+		[]runner.ArgDefinition{
+			{Description: "amount", DefaultValue: "1000"},
```
swift1337, line 959: Nit: it was 10k (`big.NewInt(10000)`)

swift1337 (approved):

e2e/e2etests/e2etests.go
lumtis: Just to make the same as other test, the amount itself is not used in the function in fact

gartnera (approved):

Issue #3110, opened 2024-11-06
Drop support for block header proofs


swift1337:
Block header proofs are not used in mainnet. Let's remove them and simplify the code

swift1337:
Related: http://github.com/zeta-chain/node/pull/949

PR #3125, opened 2024-11-07
refactor(votes)!: drop support for header proofs

swift1337:
- [x] Remove block header proofs from VoteInboundTracker.
- [x] Remove block header proofs from VoteOutboundTracker.
- [x] Remove all mentions of block header proofs from `zetaclient`.
- [x] Reserve old grpc fields for backwards compatibility.

Note that we can't simply remove the lightclient module because it will break genesis export/import. To fully drop all of this code, we need to come up with a cleanup strategy.

Closes https://github.com/zeta-chain/node/issues/3110

## Release Notes

- **New Features**
  - Whitelist functionality for SPL tokens on Solana.
  - Improved build reproducibility.

- **Refactor**
  - Simplified authorization checks and removed proof verification in tracker methods.
  - Updated outbound tracker methods to streamline parameters.

- **Bug Fixes**
  - Replaced public DHT with a private gossip peer discovery mechanism.

- **Tests**
  - Expanded test coverage for concurrent withdrawals, deposits, and reverts.
  - Simplified test cases by removing unnecessary parameters.

- **Documentation**
  - Updated changelog to reflect recent changes and enhancements.


lumtis:
No need to run [V2_MIGRATION_TESTS](https://github.com/zeta-chain/node/labels/V2_MIGRATION_TESTS) this was to test gateway migration but will no longer work on current PR since gateway is already enabled in the localnet setup.
We keep the test in the codebase for now just in case but will completely remove it after gateway is enable on mainnet for good

lumtis:
> Note that we can't simply remove the lightclient module because it will break genesis export/import. To fully drop all of this code, we need to come up with a cleanup strategy.

If the module is well isolated, and doesn't touch any external part this is fine, more overhead removing it. We will still need to have a conversation if the entire idea should be dropped.

zetaclient/zetacore/tx_test.go
```diff
@@ -106,22 +101,6 @@ func Test_GasPriceMultiplier(t *testing.T) {
 	}
 }
 
-func getHeaderData(t *testing.T) proofs.HeaderData {
-	var header ethtypes.Header
-	file, err := os.Open("../../testutil/testdata/eth_header_18495266.json")
```
lumtis, line 111: These testdata might not be needed anymore, we should check

zetaclient/chains/interfaces/interfaces.go
```diff
@@ -141,16 +132,7 @@ type ZetacoreClient interface {
 	GetZetaHotKeyBalance(ctx context.Context) (sdkmath.Int, error)
 	GetInboundTrackersForChain(ctx context.Context, chainID int64) ([]crosschaintypes.InboundTracker, error)
```
lumtis, line 134: We can remove `GetBlockHeaderChainState` normally in this interface

lumtis (approved):

zetaclient/zetacore/tx_test.go
swift1337: it is used by `x/liteclient` and `pkg/proof`

swift1337:
cc @brewmaster012

Issue #3096, opened 2024-11-05
zetaclient: fix logging in orchestrator


swift1337:
We have an invalid check whether something was updated for chain params, thus we get redundant and misleading logs. I guess that we compare identical EVM addresses in different cases as strings (`0xABC...` == `0xabc...`), so the condition is always true. Ensure that:
- We set the address only if it's not empty.
- It doesn't equal the current address.

<img width="500" alt="Screenshot 2024-11-05 at 20 57 43" src="https://github.com/user-attachments/assets/b98693f5-5f20-408f-bbd3-7c140ca6ac00">
<img width="500" alt="Screenshot 2024-11-05 at 20 57 29" src="https://github.com/user-attachments/assets/fe849411-4ef2-4e67-8d68-512513694909">

Issue #2977, opened 2024-10-07
Create RFC for Zetaclient V2


swift1337:
**Describe the Issue**
We need a unified, simple, fault-tolerant, and efficient chain orchestration logic in zetaclientd for chain observation. Let’s design one and elaborate on this in the RFC docs.

**Expected Outcome**
An RFC with components, diagrams, interfaces, and an explanation of the new zetaclient orchestration.

lumtis:
I remove https://github.com/zeta-chain/node/issues/1611 from the milestone and add this one.

Completing this document will give a clearer and more descriptive path toward a well structured ZetaClient codebase

swift1337:
Related comment: https://github.com/zeta-chain/node/pull/3139#discussion_r1837982983

PR #3139, opened 2024-11-11
fix(zetaclient): orchestrator: signer resolution

swift1337:
Closes https://github.com/zeta-chain/node/issues/3096

## Summary by CodeRabbit

- **New Features**
	- Introduced Whitelist message capability for SPL tokens.
	- Integrated SPL deposits functionality.
  
- **Improvements**
	- Enhanced build reproducibility.
	- Improved error handling and logging for address updates.

- **Bug Fixes**
	- Resolved configuration issues in the orchestrator.
	- Prevented blocked CCTX on out-of-gas scenarios during omnichain calls.
  
- **Refactor**
	- Simplified address handling logic for Ethereum-related operations.
	- Removed unnecessary checks in the `SetGatewayAddress` methods across various chains.

### DIFF
--- a/zetaclient/chains/solana/signer/signer.go
+++ b/zetaclient/chains/solana/signer/signer.go
@@ -258,25 +258,36 @@ func (signer *Signer) prepareWhitelistTx(
 
 // SetGatewayAddress sets the gateway address
 func (signer *Signer) SetGatewayAddress(address string) {
-	// parse gateway ID and PDA
-	gatewayID, pda, err := contracts.ParseGatewayIDAndPda(address)
-	if err != nil {
-		signer.Logger().Std.Error().Err(err).Msgf("cannot parse gateway address: %s", address)
+	// noop
+	if address == "" {
 		return
 	}
 
-	// update gateway ID and PDA
-	signer.Lock()
-	defer signer.Unlock()
+	// parse gateway ID and PDA
+	gatewayID, pda, err := contracts.ParseGatewayIDAndPda(address)
+	if err != nil {
+		signer.Logger().Std.Error().Err(err).Str("address", address).Msgf("Unable to parse gateway address")
+		return
+	}
 
+	// noop
+	if signer.gatewayID.Equals(gatewayID) {
+		return
+	}
+
+	signer.Logger().Std.Info().
+		Str("signer.old_gateway_address", signer.gatewayID.String()).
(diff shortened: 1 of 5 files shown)


lumtis (approved):
LGTM

zetaclient/chains/solana/signer/signer.go
```diff
@@ -258,25 +258,36 @@ func (signer *Signer) prepareWhitelistTx(
 
 // SetGatewayAddress sets the gateway address
 func (signer *Signer) SetGatewayAddress(address string) {
+	// noop
+	if address == "" {
```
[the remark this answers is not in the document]
swift1337: > should we have check as other signers that same address is set?

Yes

> should we log error if empty address is set,

No e.g. if e2e is run w/o TON it will update with empty string

> address is not in solana address format?

Yes, it's already there

fbac (approved):

zetaclient/orchestrator/orchestrator.go
```diff
@@ -164,47 +164,13 @@ func (oc *Orchestrator) resolveSigner(app *zctx.AppContext, chainID int64) (inte
...
-		if params.GatewayAddress != signer.GetGatewayAddress() {
-			signer.SetGatewayAddress(params.GatewayAddress)
-			oc.logger.Info().
-				Str("signer.gateway_address", params.GatewayAddress).
-				Msgf("updated gateway address for chain %d", chainID)
-		}
-
+		signer.SetZetaConnectorAddress(eth.HexToAddress(params.ConnectorContractAddress))
+		signer.SetERC20CustodyAddress(eth.HexToAddress(params.Erc20CustodyContractAddress))
+		signer.SetGatewayAddress(params.GatewayAddress)
```
fbac, line 169: nit: it's out of this PR scope, but I'm curious regarding all this SetXXX functions that aims to set some value, but doesn't return an error in case of failure and just do `return`.

What would happen if for example `SetGatewayAddress` fails to set the Solana GW address? I suspect this could be failing into undefined behavior, and probably in the future we should consider returning an error whenever a SetXXX function fails.

swift1337: Linked https://github.com/zeta-chain/node/issues/2977#issuecomment-2470355952

PR #3141, opened 2024-11-11
fix: allow BTC revert with dust amount v20

lumtis:
# Description

Check if the amount for a BTC revert outbound is below dust amount. In this case, cancel the tx (same logic as restricted address where there will be a tx but with 0 btc), cctx will be reverted.

Add a E2E test to test revert with dust amount. To check the effect, the "dustAmount" can be set to false in ZetaClient, the test will be stall because the outbound can't be signed


brewmaster012:
LGTM

brewmaster012 (approved):

Issue #3146, opened 2024-11-12
`zetacore` : `emissions` : Add  e2e test for emissions withdraw


kingpinXD:
Add an end-to-end test, for emissions 
This should withdraw emissions and check `available emissions` amount before and after

PR #3151, opened 2024-11-13
test: `e2e` add withdraw emissions

kingpinXD:
# Description

This pr adds withdrawal of emissions to the e2e .

The withdrawal uses the following mechanism.  
- We create authz grants from each operator/observer to the `UserEmissionsWithdraw`, this is done to try to simulate the mainnet/testnet govops process as much as possible 
- At the end of e2e, the `UserEmissionsWithdraw` withdraws all available emissions by signing the withdraw tx.

Closes : https://github.com/zeta-chain/node/issues/3146

## Summary by CodeRabbit

- **New Features**
  - Introduced a new account for emissions withdrawal, enhancing account management.
  - Added functionality for funding and withdrawing emissions within the application.
  
- **Bug Fixes**
  - Improved error handling for emissions withdrawal processes.
  
- **Documentation**
  - Updated comments and documentation for new methods and account configurations.

- **Chores**
  - Minor formatting adjustments and code clean-up for consistency across the codebase.


e2e/runner/emissions.go
```diff
@@ -0,0 +1,90 @@
...
+		changeInBalance := balanceAfter.Balance.Sub(*balanceBefore.Balance).Amount
+		if !changeInBalance.Equal(availableCoin.Amount) {
+			return fmt.Errorf(
+				"invalid balance change for observer %s, expected %s, got %s",
+				observer,
+				availableCoin.Amount,
+				changeInBalance,
+			)
+		}
+	}
```
kingpinXD, line 87: It is difficult to check the emissions pool balance or available emissions here, as there can be some emissions in the blocks in between the tx's.

CharlieMc0: Why not withdraw all emissions by default? No need to specify a certain amount. 

That's how the commission withdrawal works for validators and I can't imagine a situation in which someone needs to withdraw some but not all emission rewards.

kingpinXD: We are withdrawing all the available emissions.

cmd/zetae2e/config/local.yml
```diff
@@ -65,6 +65,10 @@ additional_accounts:
     bech32_address: "zeta1nry9yeg6njhjrp2ctppa8558vqxal9fxk69zxg"
     evm_address: "0x98c852651A9CAF2185585843d3D287600Ddf9526"
     private_key: "bf9456c679bb5a952a9a137fcfc920e0413efdb97c36de1e57455763084230cb"
+  user_emissions_withdraw:
```
lumtis, line 68: Isn't one of the observer? I would rather call it `user_observer` (or `observer_hot_key` to be more precise), might be used for other purpose in the future

contrib/localnet/scripts/start-zetacored.sh
```diff
@@ -67,6 +67,55 @@ add_v17_message_authorizations() {
     ' $json_file > temp.json && mv temp.json $json_file
 }
 
+
+add_emissions_withdraw_authorizations() {
```
lumtis, line 71: Small note here: any alternative that could work in Go?

Ideally we want to reduce the shell codebase and have everything in Go

e2e/runner/emissions.go
```diff
@@ -0,0 +1,90 @@
...
+			return fmt.Errorf("failed to parse coin amount: %w", err)
+		}
+
+		if availableCoin.Amount.IsZero() {
+			r.Logger.Print("no emissions to withdraw for observer %s", observer)
+			continue
+		}
+
+		err = r.ZetaTxServer.WithdrawAllEmissions(availableCoin.Amount, e2eutils.UserEmissionsWithdrawName, observer)
+		if err != nil {
```
lumtis, line 69:
```suggestion
		if err := r.ZetaTxServer.WithdrawAllEmissions(availableCoin.Amount, e2eutils.UserEmissionsWithdrawName, observer); err != nil {
```

e2e/runner/emissions.go
lumtis: We can update the emissions params to set percentage to 0 before withdrawing.
So no emission happens after

kingpinXD: I considered that, but we would have to unset it after the check so that tss migration and upgrade tests are successful, 
It seemed to me a bit unnecessary to do that just for this check, considering we also have unit tests .
Happy to add it though , if needed . Its trivial effort

contrib/localnet/scripts/start-zetacored.sh
kingpinXD: Okay, we can just use the same logic in Go a function ,
or a command similar to `add-observer-list` 

But I guess it's okay for now since we are adding just one authorization for now. If we need to modify more items in the genesis file, we can consider baking that into Zetacored in the future

cmd/zetae2e/config/local.yml
swift1337: > ; it's a new account used just for withdrawal.

Maybe then call it user_emissions_receiver?

kingpinXD: it's, not the receiver; the operator is still the receiver . 
It just executed the withdraw tx

cmd/zetae2e/local/local.go
```diff
@@ -160,11 +160,17 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 
 	zetaTxServer, err := txserver.NewZetaTxServer(
 		conf.RPCs.ZetaCoreRPC,
-		[]string{utils.EmergencyPolicyName, utils.OperationalPolicyName, utils.AdminPolicyName},
+		[]string{
+			utils.EmergencyPolicyName,
+			utils.OperationalPolicyName,
+			utils.AdminPolicyName,
+			utils.UserEmissionsWithdrawName,
```
fbac, line 167: nit: what I did in the distribution precompile e2e tests was creating a different zetaTxServer (with just the precompile user) in the precompile runner config instead of adding users to this zetaTxServer, which is used by every e2e runner.

Not sure what is the desired/best approach here.

e2e/runner/emissions.go
```diff
@@ -0,0 +1,90 @@
...
+
+		availableCoin, err := sdk.ParseCoinNormalized(availableAmount.Amount)
+		if err != nil {
+			return fmt.Errorf("failed to parse coin amount: %w", err)
+		}
+
+		if availableCoin.Amount.IsZero() {
+			r.Logger.Print("no emissions to withdraw for observer %s", observer)
+			continue
+		}
```
fbac, line 66: Does `availableCoin` have an `IsValid` function if I recall correctly? That would be a more complete check instead of `IsZero`.

e2e/runner/emissions.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/zeta-chain/node/cmd/zetacored/config"
+	"github.com/zeta-chain/node/e2e/txserver"
+	e2eutils "github.com/zeta-chain/node/e2e/utils"
+	emissionstypes "github.com/zeta-chain/node/x/emissions/types"
+	observertypes "github.com/zeta-chain/node/x/observer/types"
+)
+
+// FundEmissionsPool funds the emissions pool on ZetaChain with the same value as used originally on mainnet (20M ZETA)
+func (r *E2ERunner) FundEmissionsPool() error {
+	r.Logger.Print("⚙️ funding the emissions pool on ZetaChain with 20M ZETA (%s)", txserver.EmissionsPoolAddress)
```
swift1337, line 19: Do we need log + 1 one function call as a separate function? Consider removing this

swift1337 (approved):

e2e/runner/emissions.go
kingpinXD: The `IsValid` check returns true for zero amount, as 0 is a valida balance, but we don't want to process that in this test.
It does add a check for validate denom , but  `ParseCoinNormalized` already does that

e2e/runner/emissions.go
kingpinXD: I think it's fine to keep it. The logs for the test aren't too much right now

cmd/zetae2e/local/local.go
kingpinXD: I think its fine here as the e2e is supposed to be running for every test

PR #3155, opened 2024-11-14
fix: avoid panic in inscription parsing

ws4charlie:
# Description

This PR is to avoid potential panic in method `checkInscriptionEnvelope`.

```
func checkInscriptionEnvelope(t *txscript.ScriptTokenizer) error {
    if !t.Next() || t.Opcode() != txscript.OP_DATA_32 {
        // calling 't.Opcode()' will panic on the line below if t.Next() returns false, because the opcode inside is a null pointer
	return fmt.Errorf("cannot obtain public key bytes op %d or err %s", t.Opcode(), t.Err())
    }

    if !t.Next() || t.Opcode() != txscript.OP_CHECKSIG {
       // same panic could happen here
	return fmt.Errorf("cannot parse OP_CHECKSIG, op %d or err %s", t.Opcode(), t.Err())
    }

    return nil
}
```

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/bitcoin/tx_script.go
+++ b/zetaclient/chains/bitcoin/tx_script.go
@@ -337,11 +337,11 @@ func decodeInscriptionPayload(t *txscript.ScriptTokenizer) ([]byte, error) {
 // OP_PUSHBYTES_32 <32 bytes> OP_CHECKSIG <Content>
 func checkInscriptionEnvelope(t *txscript.ScriptTokenizer) error {
 	if !t.Next() || t.Opcode() != txscript.OP_DATA_32 {
-		return fmt.Errorf("cannot obtain public key bytes op %d or err %s", t.Opcode(), t.Err())
+		return fmt.Errorf("public key not found")
 	}
 
 	if !t.Next() || t.Opcode() != txscript.OP_CHECKSIG {
-		return fmt.Errorf("cannot parse OP_CHECKSIG, op %d or err %s", t.Opcode(), t.Err())
+		return fmt.Errorf("OP_CHECKSIG not found")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -337,11 +337,11 @@ func decodeInscriptionPayload(t *txscript.ScriptTokenizer) ([]byte, error) {
 // OP_PUSHBYTES_32 <32 bytes> OP_CHECKSIG <Content>
 func checkInscriptionEnvelope(t *txscript.ScriptTokenizer) error {
 	if !t.Next() || t.Opcode() != txscript.OP_DATA_32 {
-		return fmt.Errorf("cannot obtain public key bytes op %d or err %s", t.Opcode(), t.Err())
```
swift1337, line 340: I assume the panic was due to `t.Err() == nil` ?

ws4charlie: It was a panic from call to `t.Opcode()`. When `t.Next()` evaluates to `false`, the opcode inside is a nil.

swift1337 (approved):

zetaclient/chains/bitcoin/tx_script.go
```diff
@@ -337,11 +337,11 @@ func decodeInscriptionPayload(t *txscript.ScriptTokenizer) ([]byte, error) {
...
+		return fmt.Errorf("public key not found")
```
lumtis, line 340: Isn't some context we can still keep in the error?

ws4charlie: Appended error (if any) to the log print: https://github.com/zeta-chain/node/pull/3155/commits/80d275a5b6cb9b9b8d9ae85a346b85f1bcc4208d

lumtis (approved):

PR #3152, opened 2024-11-13
chore: prepare zetaclient v22.1.0

gartnera:
Prepare v22.1.0 release which includes zetaclient TSS updates. This PR should be merged.

- https://github.com/zeta-chain/node/pull/3028
- https://github.com/zeta-chain/node/pull/3041
- https://github.com/zeta-chain/node/pull/3162

PR #3161, opened 2024-11-14
fix: skip deposit fee calculation if transaction doesn't involve TSS address

ws4charlie:
# Description

Zetaclient lagged behind by `80K~140K` blocks in the Bitcoin Testnet3 transaction scanning. I found that the `FilterAndParseIncomingTx` was a time consuming call in my local simulation against Testnet3.

The root cause of the low performance is the invocation to `CalcDepositorFee` (it calls Bitcoin RPCs) inside the `FilterAndParseIncomingTx`, and it is invoked on each of the tx in the block.

Solution: Moving the `CalcDepositorFee` call to downstream code such that the calculation is only performed on the transactions that involve TSS address.

The fix was tested in my live test that simulates the block scanning process against Bitcoin Testnet3.

No changelog needed as I (@gartnera) will update the changelog before v22.1.0 release.

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/bitcoin/observer/inbound.go
+++ b/zetaclient/chains/bitcoin/observer/inbound.go
@@ -254,12 +254,6 @@ func (ob *Observer) CheckReceiptForBtcTxHash(ctx context.Context, txHash string,
 		return "", fmt.Errorf("block %d is not confirmed yet", blockVb.Height)
 	}
 
-	// calculate depositor fee
-	depositorFee, err := bitcoin.CalcDepositorFee(ob.btcClient, tx, ob.netParams)
-	if err != nil {
-		return "", errors.Wrapf(err, "error calculating depositor fee for inbound %s", tx.Txid)
-	}
-
 	// #nosec G115 always positive
 	event, err := GetBtcEvent(
 		ob.btcClient,
@@ -268,7 +262,6 @@ func (ob *Observer) CheckReceiptForBtcTxHash(ctx context.Context, txHash string,
 		uint64(blockVb.Height),
 		ob.logger.Inbound,
 		ob.netParams,
-		depositorFee,
 	)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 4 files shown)


zetaclient/chains/bitcoin/observer/inbound.go
```diff
@@ -408,11 +394,14 @@ func GetBtcEventWithoutWitness(
 	blockNumber uint64,
 	logger zerolog.Logger,
 	netParams *chaincfg.Params,
-	depositorFee float64,
 ) (*BTCInboundEvent, error) {
-	found := false
-	var value float64
-	var memo []byte
+	var (
+		found        = false
```
CryptoFewka, line 399: Warning: I do not know go. 

I understand that := is for declare+assign, while = is for assign. It seems like this changed for defining `found`  here. Is that intended? Yes is 100% an acceptable answer as I truly do not know.

gartnera (changes requested):
Please target develop first and I'll pull it in to https://github.com/zeta-chain/node/pull/3152 ?

I don't think just commenting out all the tests is acceptable right?

ws4charlie:
> Please target develop first and I'll pull it in to #3152 ?
> 
> I don't think just commenting out all the tests is acceptable right?

Understood. The `CalcDepositorFee` calls RPCs lively to get transaction fee. The old version takes the `depositorFee` as an argument so the unit tests were easy. To bring the unit tests back, I'll create an official refactor PR to develop branch.

gartnera:
> > Please target develop first and I'll pull it in to #3152 ?
> > I don't think just commenting out all the tests is acceptable right?
> 
> Understood. The `CalcDepositorFee` calls RPCs lively to get transaction fee. The old version takes the `depositorFee` as an argument so the unit tests were easy. To bring the unit tests back, I'll create an official refactor PR to develop branch.
> 
> I can try creating a official fix with refactor pointing to `develop` branch. But it would take some time, hopefully get a PR by today.

@ws4charlie A potential easy way to make this work quickly is take a parameter `depositorFee func() (float64, err)` argument to `GetBtcEvent` which would allow you to set a static fee in unit tests but still call the rpc in the live environment.

ws4charlie:
replaced by https://github.com/zeta-chain/node/pull/3162

Issue #3174, opened 2024-11-15
e2e: upgrade height missed


gartnera:
The upgrade tests seems to always be missing the 224 upgrade height now. The height only used to be missed occasionally.

https://github.com/zeta-chain/node/actions/runs/11811585456/job/32905513449
https://github.com/zeta-chain/node/actions/runs/11831651220/job/32967192313
https://github.com/zeta-chain/node/actions/runs/11851172829/job/33027246570

This is manifested by the block height monitor timing out (you must be producing blocks while e2e is running) or random `connect: connection refused` errors while a test is running.

CharlieMc0:
Is this still an issue or can it be closed again?

gartnera:
> Is this still an issue or can it be closed again?

No just happened monday night run: https://github.com/zeta-chain/node/actions/runs/12762028788/job/35569952478

lumtis:
Do you still see the issue often @gartnera ? I rarely get this one

PR #3175, opened 2024-11-15
chore: test v22 upgrade with v20 base

gartnera:
Test v20 -> v22 upgrade since that will be what actually runs on mainnet.

https://github.com/zeta-chain/node/issues/3174 requires us to bump the upgrade height


CharlieMc0 (approved):

lumtis:
Closing as we upgraded to v22

PR #3199, opened 2024-11-22
test(e2e): fix add liquidity occasional error (tentative)

lumtis:
# Description

- Remove the old logic for adding liquidity for erc20 deposit revert test and reuse the existing AddLiquidityERC20 function
- Use a higher amount for liquidity, it might be the case that the price became too high previously because of the very low liquidity and the second add liquidity failed because of this
- Print error when add liquidity fail instead of failing as it might reflect that liquidity already exist

## Summary by CodeRabbit

- **New Features**
	- Increased the ERC20 token transfer amount during funding from 10 to 10,000 tokens.
	- Enhanced the test for ERC20 deposit and refund by integrating liquidity pool creation directly into the test logic.

- **Bug Fixes**
	- Improved error handling in liquidity addition methods, providing clearer logging for transaction failures.

- **Documentation**
	- Enhanced error message clarity in the cctx status checks, providing more context for debugging.

### DIFF
--- a/cmd/zetae2e/local/erc20.go
+++ b/cmd/zetae2e/local/erc20.go
@@ -37,7 +37,7 @@ func erc20TestRoutine(
 		startTime := time.Now()
 
 		// funding the account
-		txERC20Send := deployerRunner.SendERC20OnEvm(account.EVMAddress(), 10)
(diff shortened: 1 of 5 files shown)


swift1337 (approved):
Can you elaborate on the root cause?

PR #3203, opened 2024-11-22
Testnet4 localnet

gartnera:
How I did localnet testing in https://github.com/zeta-chain/node/pull/3195


contrib/localnet/scripts/start-zetaclientd.sh
```diff
@@ -115,4 +115,9 @@ if [[ -f /root/zetaclient-config-overlay.json ]]; then
   mv /tmp/merged_config.json /root/.zetacored/config/zetaclient_config.json
 fi
 
+# signet
+export CHAIN_18333_SCAN_FROM_BLOCK=latest
+# testnet4
+export CHAIN_18334_SCAN_FROM_BLOCK=latest
+
```
gartnera, line 122: Should no longer be required with https://github.com/zeta-chain/node/pull/3224

Issue #2973, opened 2024-10-07
Implement advanced crosschain interaction with simulation tests


lumtis:
Follow up of https://github.com/zeta-chain/node/issues/2972

Create advanced crosschain interactions with:

- Gas, ERC20 and ZETA cctxs
- cctx with custom messages
- no-asset call CCTX

kingpinXD:
Add more randomization for CCTX creation for inbound vote 
Right now we use a constant value for cctx creation , this is the same as the one used in unit tests for the successful condition.

[TestKeeper_VoteInbound/successfully vote on evm deposit ](https://github.com/zeta-chain/zeta-node/blob/5d6891773974fb91100812df629465abbdc0ac0e/x/crosschain/keeper/msg_server_vote_inbound_tx_test.go#L53-L64)

But this can improved by adding more randomization

Issue #3100, opened 2024-11-05
`zetacore` : Update documenation for operation weights


kingpinXD:
Simulation operation weights constants

Operation weights are used by the simulation program to simulate the weight of different operations.
This decides what percentage of a particular type of operation is part of a block.
Based on the weights assigned in the Cosmos SDK modules, 100 is the maximum weight used, which guarantees that at least one operation of that type is present in a block.

TODO : Add more details to comment based on what the number represents in terms of the percentage of operations in a block

Issue #2975, opened 2024-10-07
Add admin message simulation in simulation tests


lumtis:
**Describe the Issue**

Add the message simulation for various admin actions:

- Updating chain params
- Pausing coins
- etc..

These message should have a very low weight compared to crosschain messages

Issue #2974, opened 2024-10-07
Implement chain and ERC20 whitelisting in simulation tests


lumtis:
Add message simulation that allow to simulate chain connection and ZRC20 whitelisting.

Note:
- This could be a dependency for https://github.com/zeta-chain/node/issues/2972 to be able to vote on gas CCTX
- This one could depends on https://github.com/zeta-chain/node/issues/2972 to have the logic to simulate the outbound processing the ERC20 whitelist

Potentially the flow could be

- 1 implement this one to create chain connection and gas ZRC20
- 2 https://github.com/zeta-chain/node/issues/2972 to have logic outbound vote
- 3 work back on this one for ERC20 whitelisting
- 4 complete https://github.com/zeta-chain/node/issues/2973 for ERC20 cctx

PR #3207, opened 2024-11-22
test: add crosschain and observer operations

kingpinXD:
# Description
The pr 
- Adds all the operations for crosschain module except `MsgMigrateTssFunds` . MigrateTSSFunds expects inbound to be paused and no new pending nonces to be generated, which does not fit in well with how simulation tests run 

- Adds all the operations for the observer module except . `MsgRemoveChainParams` . Removing or disabling a chain causes a lot of other simulations to fail, I dont think its required to be added into simulation tests 

- Simplifies the init genesis for cross-chain module 

Closes 
https://github.com/zeta-chain/node/issues/2973
https://github.com/zeta-chain/node/issues/2974
https://github.com/zeta-chain/node/issues/2975
https://github.com/zeta-chain/node/issues/3100

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Release Notes

- **New Features**
	- Added simulation operations for adding, enabling, and disabling observers and cross-chain transactions.
	- Introduced new methods for retrieving policies, all validators, and handling gas price increase flags.
	- New telemetry endpoint `/systemtime` added to enhance monitoring capabilities.

- **Improvements**
	- Enhanced error handling and validation across various simulation operations and methods.
	- Improved clarity in error messages and logging for better debugging and tracking.

- **Bug Fixes**
	- Fixed issues related to the handling of outbound trackers and cross-chain transaction states.

- **Documentation**
	- Updated comments and documentation for clarity on new features and changes in logic.


x/observer/simulation/operation_reset_chain_nonces.go
```diff
@@ -0,0 +1,87 @@
...
+func SimulateResetChainNonces(k keeper.Keeper) simtypes.Operation {
+	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accounts []simtypes.Account, _ string,
+	) (OperationMsg simtypes.OperationMsg, futureOps []simtypes.FutureOperation, err error) {
+		policyAccount, err := GetPolicyAccount(ctx, k.GetAuthorityKeeper(), accounts)
+		if err != nil {
+			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgResetChainNonces, err.Error()), nil, nil
+		}
+
+		authAccount := k.GetAuthKeeper().GetAccount(ctx, policyAccount.Address)
+		spendable := k.GetBankKeeper().SpendableCoins(ctx, authAccount.GetAddress())
```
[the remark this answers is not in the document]
kingpinXD: Auth account cannot be nil as it is using policy address, which are set using simulation accounts.

x/crosschain/simulation/operation_add_outbound_tracker.go
```diff
@@ -0,0 +1,148 @@
...
+				types.TypeMsgAddOutboundTracker,
+				"no supported chains found",
+			), nil, nil
+		}
+
+		for _, chain := range supportedChains {
+			if chains.IsEthereumChain(chain.ChainId, []chains.Chain{}) {
+				chainID = chain.ChainId
+			}
+		}
```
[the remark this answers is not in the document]
kingpinXD: Continuing with 1337 is fine

x/crosschain/genesis.go
```diff
@@ -36,24 +34,22 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
...
-		for _, elem := range genState.CrossChainTxs {
-			if elem != nil {
-				k.SaveCCTXUpdate(ctx, *elem, tss.TssPubkey)
-			}
+	// Set the cross-chain transactions only,
+	// We don't need to call SaveCCTXUpdate as the other fields are being set already
+	for _, elem := range genState.CrossChainTxs {
+		if elem != nil {
+			cctx := *elem
+			k.SetCrossChainTx(ctx, cctx)
```
kingpinXD, line 49: While working on this PR, I realized that We were saving many different parts of the state and doing calculations during genesis. 
I have the opinion now that the init genesis should be as simple as possible, and we should only be importing the exported state without modifying it. 
I am open to hearing other thoughts and opinions.

Save update did the following 
SetNonceToCCTX - Part of the observer state and should be managed by observer module 
SetCrossChainTx - We are still doing this 
UpdateInboundHashToCCTX - This is already being set separately using `genState.InboundHashToCctxList`
UpdateZetaAccounting - This can be set using genstate.ZetaAccounting , (This was remove in this pr https://github.com/zeta-chain/node/pull/3230/files#diff-2dce2b3ecc78bbb08e9523c64c1ad6ec82694ea0d821d032570383d951c5989f , but I think it makes more sense to keep it )

simulation/simulation_test.go
```diff
@@ -544,12 +562,14 @@ func TestAppSimulationAfterImport(t *testing.T) {
...
 	require.NoError(t, err)
 
+	// Initialize the new app with the exported genesis state of the first run
 	t.Log("Importing genesis into the new app")
 	newSimApp.InitChain(abci.RequestInitChain{
 		ChainId:       SimAppChainID,
 		AppStateBytes: exported.AppState,
 	})
 
+	// Run simulation on the new app
```
swift1337, line 572: Not related to this line, but shouldn't it be `require.NoError(t, simErr)`
image

simulation/simulation_test.go
```diff
@@ -439,6 +453,12 @@ func TestAppImportExport(t *testing.T) {
 	}
 }
 
+// TestAppSimulationAfterImport tests the application simulation after importing the state exported from a previous simulation run.
+// It does the following steps
+// 1. It runs a full simulation and exports the state
+// 2. It creates a new app, and db
+// 3. It imports the exported state into the new app
+// 4. It runs a simulation on the new app and verifies that there is error in the second simulation
```
swift1337, line 461: "there is an error" or "there's NO error" ? 😅 
```suggestion
// 4. It runs a simulation on the new app and verifies that there is error in the second simulation
```

x/crosschain/simulation/operations_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/zeta-chain/node/x/crosschain/simulation"
+)
+
+func Test_Matrix(t *testing.T) {
+	r := rand.New(rand.NewSource(42))
+	numVotes := 10
+	for i := 0; i < 10; i++ {
+
+		ballotVotesTransitionMatrix, yesVotePercentageArray, ballotVotesState := simulation.BallotVoteSimulationMatrix()
```
swift1337, line 16: What does this test test?

x/observer/keeper/msg_server_update_observer.go
```diff
@@ -72,7 +73,12 @@ func (k msgServer) UpdateObserver(
 		return nil, errorsmod.Wrap(types.ErrLastObserverCountNotFound, "Observer count not found")
 	}
 	if lastBlockCount.Count != totalObserverCountCurrentBlock {
-		return nil, errorsmod.Wrap(types.ErrUpdateObserver, "Observer count mismatch")
+		return nil, errorsmod.Wrapf(
+			types.ErrUpdateObserver,
+			"Observer count mismatch current block : %d , last block : %d",
```
swift1337, line 78: nit
```suggestion
			"Observer count mismatch current block: %d , last block: %d",
```

x/observer/keeper/tss.go
```diff
@@ -66,6 +66,7 @@ func (k Keeper) RemoveTSS(ctx sdk.Context) {
 	store.Delete([]byte{0})
 }
 
+// TODO define a type for a list of TSS isntead of using []types.TSS
```
swift1337, line 69: Could you elaborate on this?

x/observer/simulation/operations.go
```diff
@@ -23,85 +27,389 @@ import (
...
+	OpWeightMsgTypeMsgRemoveChainParams           = "op_weight_msg_remove_chain_params"             // #nosec G101 not a hardcoded credential
+	OpWeightMsgTypeMsgResetChainNonces            = "op_weight_msg_reset_chain_nonces"              // #nosec G101 not a hardcoded credential
+	OpWeightMsgTypeMsgUpdateGasPriceIncreaseFlags = "op_weight_msg_update_gas_price_increase_flags" // #nosec G101 not a hardcoded credential
+	OpWeightMsgTypeMsgAddObserver                 = "op_weight_msg_add_observer"                    // #nosec G101 not a hardcoded credential
 
+	// DefaultWeightMsgTypeMsgEnableCCTX We use a high weight for this operation
+	// to ensure that it is present in the block more number of times than any operation that changes the validator set
 	// Arrived at this number based on the weights used in the cosmos sdk staking module and through some trial and error
-	DefaultWeightMsgTypeMsgEnableCCTX = 3650
+	DefaultWeightMsgTypeMsgEnableCCTX                  = 3650
```
swift1337, line 45: Previously you mention
> Based on the weights assigned in the cosmos sdk modules , 100 seems to the max weight used

swift1337 (commented):
Did a first round on the PR. Solid work!

x/observer/types/keys.go
```diff
@@ -71,6 +71,10 @@ const (
 	NodeAccountKey            = "NodeAccount-value-"
 	KeygenKey                 = "Keygen-value-"
 
+	// TODO remove unused keys
```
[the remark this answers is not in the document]
kingpinXD: Ill remove it as part of this issue , will address all of them together in a separate pr 
https://github.com/zeta-chain/node/issues/3098

x/crosschain/simulation/operations_test.go
kingpinXD: removing it,it was just some debugging code

x/observer/keeper/tss.go
kingpinXD: Removed that todo , we dont need it any more

x/observer/simulation/operations.go
kingpinXD: That is correct; we want to make sure that this message is included in blocks more times than any message that modifies the observer set. 
So, more than the combined number for any msg that changes the staking amount, removes the validator etc.


The reason we need it so , is becuase any of those changes automatically disable inbound transactions, and this is the only message to enable it back on.

x/crosschain/simulation/operation_vote_inbound.go
```diff
@@ -0,0 +1,190 @@
...
+		cf, found := k.GetObserverKeeper().GetCrosschainFlags(ctx)
+		if !found {
+			return simtypes.NoOpMsg(types.ModuleName, msg.Type(), "crosschain flags not found"), nil, nil
+		}
+
+		// Return early if inbound is not enabled.
+		if !cf.IsInboundEnabled {
+			return simtypes.NoOpMsg(types.ModuleName, msg.Type(), "inbound is not enabled"), nil, nil
+		}
+
```
kingpinXD, line 95: Added this check in this PR, which would fail this transaction if inbound is disabled , instead of erroring when delivering the message( which causes the the test to stop )

This would allow us to reduce the weight we used for MsgEnableCCTX , without causing the tests to break . Add the cost of increased failures for `VoteInbound` 

This comment thread from the last pr for reference 
https://github.com/zeta-chain/node/pull/3095#discussion_r1841447204

However, as mentioned in the comment thread, I think adding a flag would be the best way forward

pkg/chains/chain_filters_test.go
```diff
@@ -48,6 +48,21 @@ func TestFilterChains(t *testing.T) {
 				return chains.ChainListByConsensus(chains.Consensus_solana_consensus, []chains.Chain{})
 			},
 		},
+		{
+			name: "Filter vm evm chains",
```
lumtis, line 52: Not directly related to this PR but test name should be full lower case if we can make the change in this file here

pkg/coin/coint_test.go
```diff
@@ -0,0 +1,28 @@
...
+	tests := []struct {
+		name string
+		c    coin.CoinType
+		want bool
+	}{
+		{"ERC20", coin.CoinType_ERC20, true},
+		{"Gas", coin.CoinType_Gas, true},
+		{"Zeta", coin.CoinType_Zeta, true},
+		{"Cmd", coin.CoinType_Cmd, false},
+		{"Unknown", coin.CoinType(100), false},
```
lumtis, line 19: Test name full lowercase

simulation/state.go
```diff
@@ -190,6 +249,39 @@ func updateAuthorityState(
 	return authorityState
 }
 
+// updateCrossChainState updates the crosschain genesis state.
+// It adds the following
+// - A gas price list for each chain
+func updateCrossChainState(
```
lumtis, line 255:
```suggestion
func updateCrosschainState(
```
Name of the module is one word

simulation/state.go
```diff
@@ -36,6 +34,9 @@ const (
 	InitiallyBondedValidators = "initially_bonded_validators"
 )
 
+// updateBankState updates the bank genesis state.
```
lumtis, line 37: The update naming seems off in this file, we return a new genesis variable instead of update a variable, should rather be named
```
bankGenesisState
evmGenesisState
...
```
I would also rename the file into genesis.go

testutil/sample/crosschain.go
```diff
@@ -54,6 +54,36 @@ func RateLimiterFlags() types.RateLimiterFlags {
 	}
 }
 
+func RateLimiterFlagsFromRand(r *rand.Rand) types.RateLimiterFlags {
```
lumtis, line 57: We can update the function above
```
func RateLimiterFlags() types.RateLimiterFlags {
	RateLimiterFlagsFromRand(Rand())
}
```
Might be the same for other function in the file

x/crosschain/genesis.go
lumtis: Yes, it's correct, genesis logic should only use getter/setter function but it seems it would makemore sense to create a dedicated PR for these changes

x/crosschain/types/outbound_tracker.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+func (o *OutboundTracker) IsMaxed() bool {
```
lumtis, line 3:
```suggestion
func (o *OutboundTracker) MaxReached() bool {
```

x/crosschain/simulation/decoders_test.go
```diff
@@ -41,14 +41,37 @@ func TestDecodeStore(t *testing.T) {
...
 	}{
-		{"CrossChainTx", fmt.Sprintf("%v\n%v", *cctx, *cctx)},
-		{"LastBlockHeight", fmt.Sprintf("%v\n%v", *lastBlockHeight, *lastBlockHeight)},
-		{"GasPrice", fmt.Sprintf("%v\n%v", *gasPrice, *gasPrice)},
-		{"OutboundTracker", fmt.Sprintf("%v\n%v", outboundTracker, outboundTracker)},
-		{"InboundTracker", fmt.Sprintf("%v\n%v", inboundTracker, inboundTracker)},
-		{"ZetaAccounting", fmt.Sprintf("%v\n%v", zetaAccounting, zetaAccounting)},
-		{"RateLimiterFlags", fmt.Sprintf("%v\n%v", rateLimiterFlags, rateLimiterFlags)},
-		{"FinalizedInbounds", fmt.Sprintf("%v\n%v", []byte{1}, []byte{1})},
+		{"CrossChainTx", fmt.Sprintf("key %s value A %v value B %v", types.CCTXKey, *cctx, *cctx)},
```
lumtis, line 44: Test name lowecase

x/crosschain/simulation/operation_abort_stuck_cctx.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/zeta-chain/node/x/crosschain/keeper"
+	"github.com/zeta-chain/node/x/crosschain/types"
+)
+
+// SimulateMsgAbortStuckCCTX generates a MsgAbortStuckCCTX with random values
+func SimulateMsgAbortStuckCCTX(k keeper.Keeper) simtypes.Operation {
+	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accounts []simtypes.Account, _ string,
+	) (OperationMsg simtypes.OperationMsg, futureOps []simtypes.FutureOperation, err error) {
+		// Pick a ethereum chain to abort a stuck cctx
+		chainID := int64(1337)
```
lumtis, line 22: We should use constant from chains package

x/crosschain/simulation/operation_abort_stuck_cctx.go
```diff
@@ -0,0 +1,128 @@
...
+		if !found {
+			return simtypes.NoOpMsg(
+				types.ModuleName,
+				types.TypeMsgAbortStuckCCTX,
+				"no cctx found",
+			), nil, nil
+		}
+
+		cctx, found := k.GetCrossChainTx(ctx, nonceToCctx.CctxIndex)
+		if !found {
```
lumtis, line 93: We should check that the CCTX is pending, purpose of simulation tests is to send message that triggers state transitions

x/crosschain/simulation/operation_gas_price_voter.go
```diff
@@ -0,0 +1,85 @@
...
+			return simtypes.NoOpMsg(
+				types.ModuleName,
+				authz.GasPriceVoter.String(),
+				"no supported chains found",
+			), nil, nil
+		}
+		randomChainID := GetRandomChainID(r, supportedChains)
+
+		// Vote for random gas price. Gas prices do not use a ballot system, so we can vote directly without having to schedule future operations.
+
```
lumtis, line 41: We have a gasPrice method in sample package why not using it

x/crosschain/simulation/operation_migrate_erc20_custody_funds.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/cosmos/cosmos-sdk/x/simulation"
+
+	"github.com/zeta-chain/node/pkg/chains"
+	"github.com/zeta-chain/node/testutil/sample"
+	"github.com/zeta-chain/node/x/crosschain/keeper"
+	"github.com/zeta-chain/node/x/crosschain/types"
+)
+
+// SimulateMigrateERC20CustodyFunds generates a MsgMigrateERC20CustodyFunds with random values and delivers it
+func SimulateMigrateERC20CustodyFunds(k keeper.Keeper) simtypes.Operation {
```
lumtis, line 19: Message will be removed since we migrated to contracts v2, let's remove this one

x/observer/simulation/operation_disable_cctx.go
```diff
@@ -0,0 +1,55 @@
...
+	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accounts []simtypes.Account, _ string,
+	) (OperationMsg simtypes.OperationMsg, futureOps []simtypes.FutureOperation, err error) {
+		policyAccount, err := GetPolicyAccount(ctx, k.GetAuthorityKeeper(), accounts)
+		if err != nil {
+			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgDisableCCTX, err.Error()), nil, nil
+		}
+		authAccount := k.GetAuthKeeper().GetAccount(ctx, policyAccount.Address)
+		spendable := k.GetBankKeeper().SpendableCoins(ctx, authAccount.GetAddress())
+
+		msg := types.MsgDisableCCTX{
```
lumtis, line 27: Can't we had randomness to disable outbound as well

x/crosschain/simulation/operation_abort_stuck_cctx.go
kingpinXD: We are using pending nonces on the current block to fetch the CCTX . 
If the CCTX is not pending , then it would be a bug in the logic and should be reported

pkg/coin/coint_test.go
kingpinXD: I think it would be better to keep this as it is , since these are all proper nouns

simulation/state.go
kingpinXD: We extract the module specific portion of the genesis from the rawstate, update and return it. The default values are still sustained . I do think update matches the logic more in this case .

The main function for this file is the `AppStateFn` which , basically returns the application state and is used by `SimulateFromSeed` function by the sdk .

testutil/sample/crosschain.go
kingpinXD: Yeah we should infact do that for most functions in the sample directory , I would like to take that on as a spearate PR 
https://github.com/zeta-chain/node/issues/3266

x/crosschain/simulation/decoders_test.go
kingpinXD: Same as https://github.com/zeta-chain/node/pull/3207#discussion_r1889314691

x/observer/simulation/operation_disable_cctx.go
kingpinXD: We can , however I do not want to add more disable Inbound messages into the simulation as the Disable inbound is already a problem as mentioned above 
https://github.com/zeta-chain/node/pull/3207#discussion_r1884281128

If we keep DisableInbound:  false .  DisableOutbound has to be true .

x/crosschain/genesis.go
kingpinXD: Happy to create a separate pr for this , and merge that before trying to merge this if needed .
However since it does nto change any logic, I considered putting it in this pr directly , but if it helps with reviewing , I can create a pr

swift1337 (approved):

pkg/coin/coint_test.go
lumtis: "should support refund for ERC20"
"should support refund for Gas"
etc..

simulation/state.go
lumtis: Then extractBankGenesisState looks more adapted.
This is the primary action of the function, not updating.

kingpinXD: I dont really agree with that, 
It updates and sends back the update state

lumtis: Update imply updating a mutable object and returning this same object. We have a raw state as an input and return a parsed and processed value (bankState). Using update naming can lead to confusion.
We need to be rigorous in API clarity, we used to have a codebase very difficult to go through because of inconsistent and unclear API

Makefile
```diff
@@ -175,8 +175,7 @@ protoImageName=ghcr.io/cosmos/proto-builder:$(protoVer)
 protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace --user $(shell id -u):$(shell id -g) $(protoImageName)
 
 proto-format:
-	@echo "--> Formatting Protobuf files"
-	@$(protoImage) find ./ -name "*.proto" -exec clang-format -i {} \;
+	@echo "--> Formatting Protobuf files"	@$(protoImage) find ./ -name "*.proto" -exec clang-format -i {} \;
```
gartnera, line 178: Revert this?

lumtis (approved):

PR #3208, opened 2024-11-23
fix(zetaclient): distinguish between known and connected peers

gartnera:
"Known peers" should not be exposed as "connected peers". They are two completely separate things which should not be mixed to avoid confusion.

Add new `/knownpeers` endpoint which will expose this info. Expose connected peers in `/connectedpeers`.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new HTTP endpoint `/knownpeers` to retrieve and manage known peers.
	- Enhanced health check functionality to differentiate between known and connected peers.
	- Added methods to manage known peers within the telemetry service.

- **Bug Fixes**
	- Improved error handling for JSON marshaling in the known peers response.

- **Documentation**
	- Updated documentation to reflect new metrics and functionalities related to known peers.

### DIFF
--- a/zetaclient/tss/healthcheck.go
+++ b/zetaclient/tss/healthcheck.go
@@ -7,10 +7,13 @@ import (
 	"sync"
 	"time"
 
+	libp2p_network "github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+	maddr "github.com/multiformats/go-multiaddr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog"
+	"github.com/samber/lo"
 	"gitlab.com/thorchain/tss/go-tss/tss"
 
 	"github.com/zeta-chain/node/pkg/bg"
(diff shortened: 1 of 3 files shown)


zetaclient/tss/healthcheck.go
```diff
@@ -89,16 +96,30 @@ func HealthcheckWorker(ctx context.Context, server *tss.TssServer, p Healthcheck
 		return nil
 	}
 
-	peersCounter := func(_ context.Context, _ *ticker.Ticker) error {
+	knownPeersCounter := func(_ context.Context, _ *ticker.Ticker) error {
 		peers := server.GetKnownPeers()
 		p.NumConnectedPeersMetric.Set(float64(len(peers)))
-		p.Telemetry.SetConnectedPeers(peers)
+		p.Telemetry.SetKnownPeers(peers)
```
swift1337, line 102: The line above should be set _known_ peers metrics

zetaclient/tss/healthcheck.go
```diff
@@ -24,13 +27,17 @@ type HealthcheckProps struct {
...
 	NumConnectedPeersMetric prometheus.Gauge
+	NumKnownPeersMetric     prometheus.Gauge
 }
 
 // HealthcheckWorker checks the health of the TSS server and its peers.
 func HealthcheckWorker(ctx context.Context, server *tss.TssServer, p HealthcheckProps, logger zerolog.Logger) error {
 	if p.NumConnectedPeersMetric == nil {
 		return errors.New("missing NumConnectedPeersMetric")
 	}
+	if p.NumKnownPeersMetric == nil {
```
swift1337, line 38: You need to add this metric `healthCheckProps` in `setup.go`

zetaclient/tss/healthcheck.go
gartnera: Right I got confused during the rebase. I don't think we actually need knownpeers as a metric. I'll remove it.

zetaclient/tss/healthcheck.go
gartnera: I removed this as it's not particularly useful.

swift1337 (approved):

Issue #3221, opened 2024-11-26
Fix all priorityFee edge-cases


swift1337:
Both in `zetacore` and `zetaclient`

lumtis:
@swift1337  Can we please add some more information in the issue description with the summary from the incident channel?

lumtis:
Consideration regarding gas stability pool mechanism:
I'm wondering if the original design wasn't overengineered and could easily simplified. We currently apply an increase percentage, this percentage might be different for gasFee and priorityFee. But it seems that instead of applying an increase percentage, we could simply update the values with the last voted ones and pay up from gas stability pool for the increase

PR #3222, opened 2024-11-26
feat(zetaclient): disable EIP-1559

swift1337:
Related: 
- https://github.com/zeta-chain/node/issues/3221

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging for the signing process, now including the public key used for signing.
- **Bug Fixes**
	- Skipped a test case for signing dynamic fee transactions due to an existing issue.
- **Refactor**
	- Simplified transaction creation logic to focus solely on legacy transactions, removing dynamic fee handling.


gartnera:
Shouldn't the base branch be v23?

lumtis (commented):
We also need the ZetaClient to not vote new priorityFee, this is actually what would mitigate the bug in the gas stability pool logic

swift1337:
> We also need the ZetaClient to not vote new priorityFee, this is actually what would mitigate the bug in the gas stability pool logic

With recent changes, all EVMs should have `observer.priorityFeeConfig.supported = false` which effectively forces `priorityFee` to zero for all votes

zetaclient/chains/evm/observer/observer.go
```diff
@@ -95,7 +95,12 @@ func NewObserver(
 		evmJSONRPC:                    evmJSONRPC,
 		outboundConfirmedReceipts:     make(map[string]*ethtypes.Receipt),
 		outboundConfirmedTransactions: make(map[string]*ethtypes.Transaction),
-		priorityFeeConfig:             priorityFeeConfig{},
+		priorityFeeConfig: priorityFeeConfig{
```
lumtis, line 98: Just to confirm, to prevent: 
- 1 voting new priorityFee?
- 2 Using EIP1559 for the transactions?

If we have 1 but not 2, isn't some possible issue of using eip1559 while priorityFee is not updated?

swift1337: checked bool means we've already determined whether this chain supports priority fee. from the second time and on  i will directly use supported bool

lumtis (approved):
LGTM, running some more tests

Issue #3227, opened 2024-11-27
zetaclient scheduler becomes desynced


gartnera:
It seems that `TryOutboundProcessor` will be called at different times on different machines (~5 seconds difference). This will degrade keysign performance.

<img width="891" alt="image" src="https://github.com/user-attachments/assets/37019d27-bc40-46ac-81a6-ec48f35ebfbf">

This can partially be attributed desync based on the start time of zetaclient:

https://github.com/zeta-chain/node/blob/08ff8818e4cd4d46e31b681e679ee04cac6a83ac/zetaclient/orchestrator/orchestrator.go#L292

We should subscribe to new block events via websockets [like this](https://github.com/zeta-chain/node/blob/771317f784db8f79861a8d1e1fa1479e690ef60d/cmd/zetae2e/local/monitor_block_production.go/#L24-L55) so that we can react to new block immediately without polling.

PR #3228, opened 2024-11-27
refactor(zetaclient): subscribe to new blocks in scheduler

gartnera:
Use cometbft block subscription to improve signer synchronization.

Closes https://github.com/zeta-chain/node/issues/3227

Performance tests show ~25% improvement in overall runtime and massive reductions p50+ stats:

[Before](https://github.com/zeta-chain/node/actions/runs/12044690764/job/33582117707):
```
perf_eth_wit | Latency report:
perf_eth_wit | min:  11.52
perf_eth_wit | max:  183.42
perf_eth_wit | mean: 84.64
perf_eth_wit | std:  48.25
perf_eth_wit | p50:  76.64
perf_eth_wit | p75:  120.73
perf_eth_wit | p90:  157.83
perf_eth_wit | p95:  171.35
perf_eth_wit | all withdraws completed
perf_eth_wit | ✅ completed in 6m2.423686072s - stress test Ether withdrawal
```

[After](https://github.com/zeta-chain/node/actions/runs/12057755759/job/33624981175?pr=3228):
```
perf_eth_wit | Latency report:
perf_eth_wit | min:  9.03
perf_eth_wit | max:  13.04
perf_eth_wit | mean: 11.60
perf_eth_wit | std:  1.20
perf_eth_wit | p50:  12.03
perf_eth_wit | p75:  12.53
perf_eth_wit | p90:  12.54
perf_eth_wit | p95:  13.03
perf_eth_wit | all withdraws completed
perf_eth_wit | ✅ completed in 4m26.799384212s - stress test Ether withdrawal
```

I expect improvements on mainnet and testnet but not as extreme without [some additional latency correction](https://github.com/zeta-chain/node/issues/3260).

## Release Notes

- **New Features**
  - Integrated CometBFT client for enhanced transaction scheduling and event handling.
  - Updated orchestration process to react to new block events from CometBFT.

- **Bug Fixes**
  - Improved error handling for WebSocket events and block event subscriptions.

- **Tests**
  - Enhanced testing framework with mock CometBFT client for better simulation of interactions and error scenarios.


zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@
 		return err
 	}
 
-	observeTicker := time.NewTicker(3 * time.Second)
-	var lastBlockNum int64
+	blockEventChan, err := oc.cometbftClient.Subscribe(ctx, "", "tm.event='NewBlock'")
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
gartnera: This is handled internally in cometbft

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@
...
+		case <-time.After(time.Second * 10):
+			// the subscription should automatically reconnect after zetacore
+			// restart, but we should log this just in case that logic is not
+			// working
+			oc.logger.Warn().Msg("runScheduler: no blocks after 10 seconds")
+		case event := <-blockEventChan:
+			newBlockEvent, ok := event.Data.(cometbft_types.EventDataNewBlock)
+			if !ok {
+				return fmt.Errorf("expecting new block event, got %T", event.Data)
+			}
```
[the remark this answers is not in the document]
gartnera: This is not a fatal error. zetacore may take some time to restart.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@
...
+					oc.ScheduleCctxEVM(ctx, zetaHeight, chainID, cctxList, ob, signer)
+				case chain.IsBitcoin():
+					oc.ScheduleCctxBTC(ctx, zetaHeight, chainID, cctxList, ob, signer)
+				case chain.IsSolana():
+					oc.ScheduleCctxSolana(ctx, zetaHeight, chainID, cctxList, ob, signer)
+				case chain.IsTON():
+					oc.ScheduleCCTXTON(ctx, zetaHeight, chainID, cctxList, ob, signer)
+				default:
+					oc.logger.Error().Msgf("runScheduler: no scheduler found chain %d", chainID)
+					continue
```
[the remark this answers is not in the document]
gartnera: That is unrelated refactoring.

cmd/zetaclientd/start.go
```diff
@@ -217,12 +219,24 @@ func Start(_ *cobra.Command, _ []string) error {
...
+	cometbftURL := fmt.Sprintf("http://%s:%d", cfg.ZetaCoreURL, 26657)
+	cometbftClient, err := cometbft_client.New(cometbftURL, "/websocket")
+	if err != nil {
+		return errors.Wrapf(err, "new cometbft client (%s)", cometbftURL)
+	}
+	// start websockets
+	err = cometbftClient.WSEvents.Start()
+	if err != nil {
+		return errors.Wrap(err, "cometbft start")
+	}
```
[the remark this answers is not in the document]
gartnera: 1. we hardcode other port numbers in zetaclient. unrelated refactoring.
2. sure
3. error messages are fine imo

cmd/zetaclientd/start.go
```diff
@@ -217,12 +219,24 @@ func Start(_ *cobra.Command, _ []string) error {
 		return errors.Wrap(err, "unable to create chain observer map")
 	}
 
+	cometbftURL := fmt.Sprintf("http://%s:%d", cfg.ZetaCoreURL, 26657)
```
kingpinXD, line 222: Consider adding this as a function for Config struct.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
 		return err
 	}
 
-	observeTicker := time.NewTicker(3 * time.Second)
-	var lastBlockNum int64
+	blockEventChan, err := oc.cometbftClient.Subscribe(ctx, "", "tm.event='NewBlock'")
```
kingpinXD, line 299: Can we define the query filter as a constant?

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
...
+					oc.ts.AddFeeEntry(bn, diff.Int64())
+					oc.lastOperatorBalance = balance
+				}
+			}
+
+			// set current hot key burn rate
+			metrics.HotKeyBurnRate.Set(float64(oc.ts.HotKeyBurnRate.GetBurnRate().Int64()))
+
+			// get chain ids without zeta chain
+			chainIDs := lo.FilterMap(app.ListChains(), func(c zctx.Chain, _ int) (int64, bool) {
```
kingpinXD, line 336: Could you use the defined FilterChains  instead of importing an external library? 
I see this library being used in other places in the codebase. Minimizing usage would be a good idea.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
...
+					WithLabelValues(chain.Name()).
+					Set(float64(len(cctxList)))
+
+				if len(cctxList) == 0 {
 					continue
 				}
-				if lastBlockNum == 0 {
-					lastBlockNum = bn - 1
+
+				if !app.IsOutboundObservationEnabled() {
```
kingpinXD, line 383: Can we add a log here and maybe move this checkup? 
Do we even need to try to start processing an outbound if it's disabled?

cmd/zetaclientd/start.go
gartnera: I'd prefer to keep this very simple for now and refactor it later. We also instantiate this client internally in zetaclient/zetacore/client_cosmos.go but it really should probably be injected from the top imo.

zetaclient/orchestrator/orchestrator.go
gartnera: `app.FilterChains(...)` returns `[]zctx.Chain` not `[]int64` as required by `oc.GetPendingCctxsWithinRateLimit`

This is not new logic (it's just unindented).

zetaclient/orchestrator/orchestrator.go
gartnera: This logic does seem a bit strange to me too. But it's not new, this is unindenting the existing logic.

cmd/zetaclientd/start.go
lumtis, line 222:
```suggestion
	cometBFTURL := fmt.Sprintf("http://%s:%d", cfg.ZetaCoreURL, 26657)
```

cmd/zetaclientd/start.go
```diff
@@ -217,12 +219,24 @@ func Start(_ *cobra.Command, _ []string) error {
...
+	cometbftClient, err := cometbft_client.New(cometbftURL, "/websocket")
```
lumtis, line 223:
```suggestion
	cometBFTClient, err := cometbft_client.New(cometbftURL, "/websocket")
```

cmd/zetaclientd/start.go
lumtis: Can we open an issue for the future consideration?

lumtis (commented):
@ws4charlie @swift1337 please have a look

zetaclient/orchestrator/orchestrator.go
```diff
@@ -84,6 +89,7 @@ type multiLogger struct {
 func New(
 	ctx context.Context,
 	client interfaces.ZetacoreClient,
+	cometbftClient cometbft_rpc_client.Client,
```
swift1337, line 92: Let's integrate this into zetacore client wrapper in `zetaclient/zetacore/*`

zetaclient/orchestrator/orchestrator.go
swift1337, line 299: `oc.zetacore.NewBlockSubscriber(ctx) (chan ..., error)`

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +296,114 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
 		return err
 	}
 
-	observeTicker := time.NewTicker(3 * time.Second)
```
swift1337, line 292: This ticker was running every 3s. Now, it runs every block or slower. What would happen if the next tick occurred at N+2 instead of N+1?

zetaclient/orchestrator/orchestrator.go
gartnera: Isn't it better to ultimately have smaller focused interfaces rather than one giant interface.

zetaclient/orchestrator/orchestrator.go
gartnera: I'm not exactly sure what you're saying. We weren't doing anything until we got a new block.

zetaclient/orchestrator/orchestrator.go
swift1337: Generally yes, but here it introduces some unnecessary details to the orchestrator. Also mocking an interface for `oc.zetacore.NewBlockSubscriber(ctx) (chan ..., error)` is easier than for websocket client

zetaclient/orchestrator/orchestrator.go
swift1337: Is it possible to "lose" a "tick" and jump from the 100th to the 102nd block?

gartnera: Yes that is possible if zetacored restarts. But that would have resulted in a missed tick in the polling style too.

The actual work to do comes from the `GetPendingCctxsWithinRateLimit` call.

zetaclient/orchestrator/orchestrator.go
kingpinXD: yeah , we could iterate over the chains to fetch the chain IDs
The main concern is to avoid external libs as much as possible and stick to the std lib.
 
I think keeping it as it is in the current PR is fine, but we can open an issue to track this.

zetaclient/orchestrator/orchestrator.go
kingpinXD: would be great if a issue can be added , maybe under the zetaclient refactor epic .
Resolving for now

zetaclient/orchestrator/orchestrator.go
kingpinXD: Is missing a block a problem? 
Since we have the block number and the last block number, couldn't we use that to verify that no blocks are missed?

zetaclient/orchestrator/orchestrator.go
gartnera: Ok I've moved it into the zetacore client.

zetaclient/orchestrator/orchestrator.go
gartnera: Done

zetaclient/orchestrator/orchestrator.go
gartnera: No missing a block should be fine. The one scenario we should consider guarding against is if the zetacored node is resyncing. But that would break a lot of other stuff in zetaclient anyway.

gartnera:
Ok this is ready for another round of reviews. There is still some more cleanup that could be done in zetaclient regarding the tendermint/cometbft naming and multiple instantiation of cometbft rpc clients but I would like to defer that.

zetaclient/chains/interfaces/interfaces.go
```diff
@@ -145,6 +146,7 @@ type ZetacoreClient interface {
 	GetUpgradePlan(ctx context.Context) (*upgradetypes.Plan, error)
 
 	PostOutboundTracker(ctx context.Context, chainID int64, nonce uint64, txHash string) (string, error)
+	NewBlockSubscriber(ctx context.Context) (chan cometbft_types.EventDataNewBlock, error)
```
swift1337, line 149:
```suggestion
	NewBlockSubscriber(ctx context.Context) (chan cometbfttypes.EventDataNewBlock, error)
```

zetaclient/zetacore/client.go
```diff
@@ -6,7 +6,8 @@ import (
 	"strings"
 	"sync"
 
-	rpchttp "github.com/cometbft/cometbft/rpc/client/http"
+	cometbft_rpc_client "github.com/cometbft/cometbft/rpc/client"
```
swift1337, line 9:
```suggestion
	cometbftrpc "github.com/cometbft/cometbft/rpc/client"
	cometbfthttp "github.com/cometbft/cometbft/rpc/client/http"
```

https://google.github.io/styleguide/go/decisions.html#package-names

zetaclient/zetacore/client_subscriptions.go
```diff
@@ -0,0 +1,35 @@
+package zetacore
+
+import (
+	"context"
+
+	cometbft_types "github.com/cometbft/cometbft/types"
```
swift1337, line 6: same here

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +289,113 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
...
 			return nil
-		case <-observeTicker.C:
-			{
-				bn, err := oc.zetacoreClient.GetBlockHeight(ctx)
+		case <-time.After(time.Second * 10):
+			// the subscription should automatically reconnect after zetacore
+			// restart, but we should log this just in case that logic is not
+			// working
+			oc.logger.Warn().Msg("runScheduler: no blocks after 10 seconds")
+		case newBlock := <-newBlockChan:
```
ws4charlie, line 307: There is a slight difference I noticed when compared to non-subscription model. So can `newBlockChan` hold multiple items (blocks) inside? For example, if it contains `[block100, block101, block102]` somehow (due to delay), do we still action on old blocks `block100, block101`? Th non-subscription model only deal with whatever block latest.

zetaclient/orchestrator/orchestrator.go
```diff
@@ -289,122 +289,113 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
...
 	for {
 		select {
 		case <-oc.stop:
 			oc.logger.Warn().Msg("runScheduler: stopped")
 			return nil
-		case <-observeTicker.C:
-			{
-				bn, err := oc.zetacoreClient.GetBlockHeight(ctx)
+		case <-time.After(time.Second * 10):
+			// the subscription should automatically reconnect after zetacore
```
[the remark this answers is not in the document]
gartnera: I did manually test it (docker restart zetacore1) during an e2e test run. The reconnect is all cometbft logic so it is a bit challenging to test more on our side.

zetaclient/orchestrator/orchestrator.go
gartnera: > do we still action on old blocks

We are not actually doing work in response to block, that comes from the `GetPendingCctxsWithinRateLimit`. There is already logic which will prevent outbound from being processed again during the next tick.

I don't think this is a huge issue but I could add another check to just ignore block notifications if the block time is older than 10 seconds or something like that.

zetaclient/orchestrator/orchestrator.go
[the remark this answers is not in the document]
ws4charlie: okay, as long as we already tested it manually.

zetaclient/orchestrator/orchestrator.go
ws4charlie: yeah, ignoring block time older than 10 seconds or maybe just set the buffered channel size as 1 or 2, like
`blockEventChan := make(chan cometbfttypes.EventDataNewBlock, 2)`. You could do whatever convenient to you, just in case the channel gets blocks piled up.

ws4charlie (approved):
looks good overall. minor comment

lumtis (approved):

Issue #3157, opened 2024-11-14
Aborted CCTX `MsgAbortStuckCCTX` are still listed in list-pending-cctx query


lumtis:
**Describe the Issue**
CCTX aborted through the `MsgAbortStuckCCTX` can still be listed in the pending cctx query.

These CCTXs should be filtered https://github.com/zeta-chain/node/blob/19c929c50ef409ebd858b66767b60413b07de399/x/crosschain/keeper/grpc_query_cctx.go#L158

We might need to add a call to
- Check the cctx is of the lowest nonce
- Use `RemoveFromPendingNonces` to remove the nonce

but we need to investigate if an underlying condition causing the cctx to be listed as it supposed to be filtered per the line above

PR #3230, opened 2024-11-28
fix: update pending nonces when aborting a cctx through MsgAbortStuckCCTX

kingpinXD:
# Description

Closes https://github.com/zeta-chain/node/issues/3157
https://github.com/zeta-chain/node/issues/3243

The pr does the following
- Calls `SaveOutbound` at the end of MsgAbortCCTX , which processes the outbound in the same way as a regular  cctx 

- Add simple functions for the status check , in the types package 
```
func (c CctxStatus) IsTerminalStatus() bool {
	return c == CctxStatus_Aborted || c == CctxStatus_Reverted || c == CctxStatus_OutboundMined
}

func (c CctxStatus) IsPendingStatus() bool {
	return c == CctxStatus_PendingInbound || c == CctxStatus_PendingOutbound || c == CctxStatus_PendingRevert
}
```

- Refactors the genesis import for the cross-chain module to remove the double counting of aborted amounts when setting ZetaAccounting.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Release Notes

- **New Features**
	- Introduced methods to easily determine if a cross-chain transaction status is terminal or pending.
  
- **Bug Fixes**
	- Improved handling of unsupported transaction versions for Solana.
	- Enhanced transaction management by updating pending nonces when aborting a transaction.
	- Fixed retry behavior for inbound vote message validation failures.

- **Tests**
	- Added new tests for evaluating terminal and pending statuses of cross-chain transactions.
	- Enhanced existing tests for aborting transactions to ensure accurate nonce handling and status message validation.


e2e/utils/zetacore.go
```diff
@@ -113,7 +113,7 @@ func WaitCctxsMinedByInboundHash(
 		allFound := true
 		for j, cctx := range res.CrossChainTxs {
 			cctx := cctx
-			if !IsTerminalStatus(cctx.CctxStatus.Status) {
+			if !cctx.CctxStatus.Status.IsTerminalStatus() {
```
swift1337, line 116: Let's  remove redundant suffixes
```suggestion
			if !cctx.CctxStatus.Status.IsTerminal() {
```

x/crosschain/keeper/cctx.go
```diff
@@ -48,15 +59,39 @@ func (k Keeper) SetCctxAndNonceToCctxAndInboundHashToCctx(
 		in.CctxIndex = append(in.CctxIndex, cctx.Index)
 	}
 	k.SetInboundHashToCctx(ctx, in)
+}
 
+func (k Keeper) UpdateZetaAccounting(
```
swift1337, line 64: Should we export these functions to other packages? If not, let's un-export most of them to retain only the essential API methods for external use.

kingpinXD: 40fade6e75ce72da71aa17ff87625d4dd40d6890

x/crosschain/genesis.go
```diff
@@ -10,7 +11,9 @@ import (
 // InitGenesis initializes the crosschain module's state from a provided genesis
 // state.
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
-	k.SetZetaAccounting(ctx, genState.ZetaAccounting)
+	// Always set the zeta accounting to zero at genesis.
+	// ZetaAccounting value is build by iterating through all the cctxs and adding the amount to the zeta accounting.
+	k.SetZetaAccounting(ctx, types.ZetaAccounting{AbortedZetaAmount: sdkmath.ZeroUint()})
```
kingpinXD, line 16: https://github.com/zeta-chain/zeta-node/blob/b2c1b3220a0890aa644f3a85d6629c8ed93e186e/x/crosschain/genesis.go#L93-L96

The export is not needed anymore, but it is still fine to keep it as it is .

ws4charlie: Q: so the given `genState` is empty in its cctx list and that is why we set ZetaAccounting to zero?

x/crosschain/keeper/msg_server_vote_outbound_tx.go
```diff
@@ -100,6 +100,11 @@ func (k msgServer) VoteOutbound(
 		return &types.MsgVoteOutboundResponse{}, nil
 	}
 
+	// If the CCTX is in a terminal state, we do not need to process it.
+	if cctx.CctxStatus.Status.IsTerminal() {
```
ws4charlie, line 104: Is there any downstream rewards distribution? If yes, this could block a late voter from getting rewards; if no, it is fine to return early.

x/crosschain/types/status.go
```diff
@@ -86,3 +86,11 @@ func stateTransitionMap() map[CctxStatus][]CctxStatus {
 	}
 	return stateTransitionMap
 }
+
+func (c CctxStatus) IsTerminal() bool {
+	return c == CctxStatus_Aborted || c == CctxStatus_Reverted || c == CctxStatus_OutboundMined
+}
+
+func (c CctxStatus) IsPending() bool {
```
ws4charlie, line 94: Is it a good idea to just return `!c.IsTerminal` to indicates the two mutually exclusive states?

x/crosschain/keeper/cctx.go
```diff
@@ -48,15 +59,42 @@ func (k Keeper) SetCctxAndNonceToCctxAndInboundHashToCctx(
...
+) {
+	if cctx.CctxStatus.Status == types.CctxStatus_Aborted &&
+		cctx.InboundParams.CoinType == coin.CoinType_Zeta &&
+		cctx.CctxStatus.IsAbortRefunded == false {
 		k.AddZetaAbortedAmount(ctx, GetAbortedAmount(cctx))
 	}
 }
 
+// setNonceToCCTXMapping updates the mapping between a nonce and a cctx index if the cctx is in a PendingOutbound or PendingRevert state
+func (k Keeper) setNonceToCCTXMapping(
```
lumtis, line 77:
```suggestion
func (k Keeper) updateNonceToCCTX(
```
We update an existing value
`NonceToCCTX` is the name of the variable, no need for mapping suffix

x/crosschain/keeper/cctx.go
```diff
@@ -12,29 +12,40 @@ import (
...
+
+// 3. Update the mapping inboundHash -> cctxIndex
+// A new value is added to the mapping when a single inbound hash is connected to multiple cctx indexes
+// If the inbound hash to cctx mapping does not exist, a new mapping is created and the cctx index is added to the list of cctx indexes
+
 // 4. update the zeta accounting
+// Zeta-accounting is updated aborted cctxs of cointtype zeta.When a cctx is aborted it means that `GetAbortedAmount`
+//of zeta is locked and cannot be used.
+
 func (k Keeper) SetCctxAndNonceToCctxAndInboundHashToCctx(
```
lumtis, line 32: Can we update this name on the way? It is unnecessarily verbose
This can simply be
```
SaveCCTXUpdate
```

x/crosschain/genesis.go
kingpinXD: Not sure I understand the question , 
ZetaAccounting only accounts for the aborted zeta amount right now , which is calculated by iterating over cctxs. 
We do not need to set it manually.

x/crosschain/keeper/msg_server_vote_outbound_tx.go
kingpinXD: This section of logic should only be executed once for every CCTX , and this part is reponsible to either 
- Finalize cctx 
- Create a revert 


The only reason when the status can `Aborted` at line 104 is if the CCTX has been aborted using `AbortStuckCCTX` without waiting for all the votes to go through . 
In this condition, I don't think we need to payout rewards for the ballot , wdyt?  
Can you think of any other condition for this clause ?

x/crosschain/keeper/cctx.go
```diff
@@ -48,15 +59,42 @@ func (k Keeper) SetCctxAndNonceToCctxAndInboundHashToCctx(
 		in.CctxIndex = append(in.CctxIndex, cctx.Index)
 	}
 	k.SetInboundHashToCctx(ctx, in)
+}
 
-	if cctx.CctxStatus.Status == types.CctxStatus_Aborted && cctx.InboundParams.CoinType == coin.CoinType_Zeta {
+// updateZetaAccounting updates the zeta accounting with the amount of zeta that was locked in an aborted cctx
+func (k Keeper) updateZetaAccounting(
```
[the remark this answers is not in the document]
kingpinXD: You mean `ZetaAccounting` , 
It was initially added as part of a recommendation of an audit to track the zeta supply on the chain. 
It was ment to be used with the `ZetaSupplyChecker` to add checks for the supply. In case we are not planning to use this design for the supply checks , we can remove this ( In a separate PR )

x/crosschain/keeper/msg_server_vote_outbound_tx.go
ws4charlie: it is fine if there is not rewards distribution.

ws4charlie (approved):

PR #3245, opened 2024-12-05
chore: update codeowner

lumtis:
## Summary by CodeRabbit

- **Chores**
	- Updated the list of code owners by removing `@fbac` from the root directory ownership.


CryptoFewka (approved):
Is there any reason not to use a group, rather than the explicit list of code owners used right now?

Issue #3248, opened 2024-12-05
zetaclient: add minimum version check


gartnera:
zetaclient should automatically shut down if does not match the zetacored major version or if it is not greater than or equal to a configured minimum version.

kingpinXD:
We should only consider a protocol / policy-based solution like this if we have exhausted all other options. 
If we move towards a more decentralized network in the future, this will add one more thing to worry about.

With that said, I do not see any problems with this approach right now

PR #3252, opened 2024-12-06
feat: add coordinated zetaclient restart

gartnera:
Add an `OperationalFlags` resource to zetacore which will be used for various operational functionality. The initial field will be `restart_height` which will be used for scheduling a coordinated zetaclient restart (https://github.com/zeta-chain/node/issues/3248).

The next thing I will add to this resource is a `minimum_version` field for for the [minimum version check](https://github.com/zeta-chain/node/issues/3248).

This PR is just zetacore functionality at the moment. Once I get initial reviews, I will continue to add the zetaclient side, add additional unit test coverage, and generate files.

TODO:

- [x] zetacore test coverage
- [x] state import/export
- [x] basic zetaclient implementation
- [x] zetaclient unit tests
- [x] full e2e tests

## Release Notes

- **New Features**
	- Introduced functionality for managing and querying operational flags within the ZetaChain framework.
	- Added command-line interface command for updating operational flags.
	- Expanded end-to-end testing for operational flags functionality.

- **Bug Fixes**
	- Enhanced test coverage for authorization management related to operational flags.

- **Documentation**
	- Updated message types and RPC methods for operational flags in the API documentation.

- **Chores**
	- Added new constants and message types to improve operational flags handling.


x/observer/keeper/operational_flags.go
```diff
@@ -0,0 +1,26 @@
...
+func (k Keeper) GetOperationalFlags(ctx sdk.Context) (val types.OperationalFlags, found bool) {
+	found = false
+	store := ctx.KVStore(k.storeKey)
+	b := store.Get(types.KeyPrefix(types.OperationalFlagsKey))
+	if b == nil {
+		return
+	}
+	found = true
+	k.cdc.MustUnmarshal(b, &val)
+	return
```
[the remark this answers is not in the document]
gartnera: Both panic?

x/observer/keeper/operational_flags.go
```diff
@@ -0,0 +1,26 @@
...
+}
```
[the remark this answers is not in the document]
gartnera: If it's failing to unmarshal when there is data found there are much larger problems. This scenario should never be hit.

proto/zetachain/zetacore/observer/operational.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+package zetachain.zetacore.observer;
+
+option go_package = "github.com/zeta-chain/node/x/observer/types";
+
+message OperationalFlags {
+  // Height for a coordinated zetaclient restart.
+  // Will be ignored if missed.
+  int64 restart_height = 1;
```
ws4charlie, line 9: Not a strong opinion. Are `zetaclient_restart_height` and `OperationParams` also good names? It's up to you.

ws4charlie (dismissed):
the idea looks clear

proto/zetachain/zetacore/observer/operational.proto
gartnera: It's already under the `observer` section so doesn't that imply that it's zetaclient functionality.

x/observer/client/cli/tx_update_operational_flags.go
```diff
@@ -0,0 +1,59 @@
...
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			var operationalFlags types.OperationalFlags
+
+			flagSet := cmd.Flags()
+			file, _ := flagSet.GetString("file")
+			restartHeight, _ := flagSet.GetInt64("restart-height")
```
kingpinXD, line 30: Why not just use a command line argument and accept a filepath . 
That would be similar to the logic we have for all other queries.

The values should be declared as constants if you need to use a flag. Although I would still be in favour of using the args as it keep the codebase uniform

x/observer/types/keys.go
```diff
@@ -80,6 +80,8 @@ const (
 	NonceToCctxKeyPrefix   = "NonceToCctx-value-"
 
 	ParamsKey = "Params-value-"
+
+	OperationalFlagsKey = "OperationalFlags"
```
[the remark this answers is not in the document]
kingpinXD: I agree; we should maintain consistency

proto/zetachain/zetacore/observer/operational.proto
kingpinXD, line 9: Consider using unit64

x/observer/types/message_update_operational_flags.go
```diff
@@ -0,0 +1,53 @@
+package types
```
kingpinXD, line 1: Please add a unit test for this file

x/observer/keeper/grpc_query_operational_flags.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/zeta-chain/node/x/observer/types"
+)
+
+func (k Keeper) OperationalFlags(
```
kingpinXD, line 13: A cli query would be good to have for these flags.

x/authority/types/authorization_list_test.go
```diff
@@ -399,6 +399,7 @@ func TestDefaultAuthorizationsList(t *testing.T) {
 			sdk.MsgTypeURL(&observertypes.MsgUpdateChainParams{}),
 			sdk.MsgTypeURL(&observertypes.MsgEnableCCTX{}),
 			sdk.MsgTypeURL(&observertypes.MsgUpdateGasPriceIncreaseFlags{}),
+			sdk.MsgTypeURL(&observertypes.MsgUpdateOperationalFlags{}),
```
kingpinXD, line 402: We would also need to update the authorization list saved to the state via `AddAuthorisation`.
Can you add an issue to track that

kingpinXD (changes requested):
- Unit tests need to be added , ideally one for every new file .

- Cosider adding operational Flags to the genesis state ,and implementing import export .

x/observer/types/message_update_operational_flags.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (msg *MsgUpdateOperationalFlags) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Creator); err != nil {
+		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	if msg.OperationalFlags.RestartHeight < 0 {
+		return cosmoserrors.Wrap(sdkerrors.ErrInvalidRequest, "restart height cannot be negative")
+	}
```
kingpinXD, line 50: Please create and `OperationalFlags.Validate()` function in the types package

x/observer/keeper/msg_server_update_operational_flags.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	authoritytypes "github.com/zeta-chain/node/x/authority/types"
+	"github.com/zeta-chain/node/x/observer/types"
+)
+
+func (k msgServer) UpdateOperationalFlags(
```
kingpinXD, line 13: Please add a comment to explain how these values would be used

x/observer/client/cli/tx_update_operational_flags.go
gartnera: > Why not just use a command line argument and accept a filepath .

I wanted to make interactive testing easier.

> The values should be declared as constants if you need to use a flag.

Yes

proto/zetachain/zetacore/observer/operational.proto
gartnera: It seems we use int64 for most other heights in the observer package so I would prefer to stay consistent.

Examples:
- `keygen_zeta_height`
- `finalizedZetaHeight`
- `keyGenZetaHeight`

x/authority/types/authorization_list_test.go
gartnera: I think we could do that via a migration right?

kingpinXD: I don't think we would need a migration script, it might just be easier to add it via message, wdyt ?

proto/zetachain/zetacore/observer/operational.proto
kingpinXD: We use uint64 as well, for example 
`GasPrice` [For external heights]
`LastBlockHeight`

The cosmos block height is int64; I guess it might make sense to use int64, where we are planning to compare against the cosmos height, Which I don't think is the case here; I don't believe this height can ever be negative.

lumtis: We use int64 for block height in general, even in geth, it's an antipattern but better to maintain consistency and avoiding conversion with overflow risk, this value is set through admin prop so low risk

x/authority/types/authorization_list_test.go
lumtis: Yes, better to add it manually and avoid migration script, we can open a operation issue in protocol-private

proto/zetachain/zetacore/observer/operational.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+package zetachain.zetacore.observer;
+
+option go_package = "github.com/zeta-chain/node/x/observer/types";
+
+message OperationalFlags {
```
lumtis, line 6: Would be good to add a comment to explain idea of this structure as well

lumtis (commented):
The structure and the idea look good to me.
For new state variable, we also need to add a field in the genesis and update export/import

proto/zetachain/zetacore/observer/operational.proto
kingpinXD: I agree with the fact that we should use one type if choosing int64 it would be good to have an issue to refactor the ones I mentioned to int64 at some point

gartnera: >  issue to refactor the ones I mentioned to int64 at some point

breaking changes are 10x worse than inconsistent api design imo.

x/observer/keeper/msg_server_update_operational_flags.go
```diff
@@ -0,0 +1,28 @@
...
+	// check permission
+	err := k.GetAuthorityKeeper().CheckAuthorization(ctx, msg)
+	if err != nil {
+		return nil, errors.Wrap(authoritytypes.ErrUnauthorized, err.Error())
+	}
+
+	k.Keeper.SetOperationalFlags(ctx, msg.OperationalFlags)
+
+	return &types.MsgUpdateOperationalFlagsResponse{}, nil
+}
```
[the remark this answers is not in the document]
gartnera: Events are not needed imo

x/observer/keeper/operational_flags.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/zeta-chain/node/x/observer/types"
+)
+
+func (k Keeper) SetOperationalFlags(ctx sdk.Context, operationalFlags types.OperationalFlags) {
+	store := ctx.KVStore(k.storeKey)
+	b := k.cdc.MustMarshal(&operationalFlags)
+	key := types.KeyPrefix(types.OperationalFlagsKey)
+	store.Set(key, b)
+}
```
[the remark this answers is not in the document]
gartnera: Events are not needed imo

x/authority/types/authorization_list_test.go
kingpinXD: Migration scripts are meant for state migrations of incompatible data between two releases and not for tasks like this. 
If you could create an issue for any problem that occurred when adding authorizations, I would be happy to look into it

CharlieMc0: If an upgrade will break an authorization that already exists, then we should use migration scripts to ensure continuity of service and that everything continues to work as it did prior to the upgrade. 

If the authorization is new then it should be added on-chain manually instead of changing authorizations unnecessarily via migration scripts. 

I am not 100% sure `AddAuthorisation` is going to be doing. But it looks like a brand new message type right? I think we should set up the permissions manually using messages after the upgrade goes live.

x/observer/keeper/msg_server_update_operational_flags.go
gartnera: The values are already commented on the proto side?

x/authority/types/authorization_list_test.go
lumtis: >I am not 100% sure AddAuthorisation is going to be doing. But it looks like a brand new message type right? I think we should set up the permissions manually using messages after the upgrade goes live.

No we've been using it multiple time.

The upgrade doesn't break anything, the authorization can be added anytime before the first use of the message.

>I definitely disagree since we seem to always mess up these manual authorization migrations. We really should aim to do as few manual steps as possible IMO.

We did once for a message that was needed right at the upgrade (VoteGasPrice) this would have required a migration script to be functional as soon as the upgrade happen. We don't have this requirement here. We just need to track an issue to add it.

lumtis: https://github.com/zeta-chain/protocol-private/issues/204

e2e/e2etests/e2etests.go
```diff
@@ -132,6 +132,7 @@ const (
 	TestMigrateERC20CustodyFundsName  = "migrate_erc20_custody_funds"
 	TestMigrateTSSName                = "migrate_TSS"
```
lumtis, line 133:
```suggestion
	TestMigrateTSSName                = "migrate_tss"
```
Not related to this PR but just easier to update here instead of opening a new PR

lumtis (approved):
LGTM

zetaclient/maintenance/shutdown_listener.go
```diff
@@ -0,0 +1,125 @@
...
+func (o *ShutdownListener) Listen(ctx context.Context, action func()) {
+	var (
+		withLogger = bg.WithLogger(o.logger)
+		onComplete = bg.OnComplete(action)
+	)
+
+	bg.Work(ctx, o.waitForUpdate, bg.WithName("shutdown_listener.wait_for_update"), withLogger, onComplete)
+}
+
+func (o *ShutdownListener) waitForUpdate(ctx context.Context) error {
```
swift1337, line 45: consider using `ticker` package as it contains exactly the same logic but makes code shorter

swift1337 (approved):

x/observer/types/operational.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	"errors"
+)
+
+var (
```
kingpinXD, line 7: Please move this to errors.go

x/observer/keeper/grpc_query_operational_flags_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+		// set the value and ensure it's returned by the query
+		restartHeight := int64(100)
+		k.SetOperationalFlags(ctx, types.OperationalFlags{
+			RestartHeight: restartHeight,
+		})
+		res, err = k.OperationalFlags(wctx, &types.QueryOperationalFlagsRequest{})
+		require.NoError(t, err)
+		require.Equal(t, restartHeight, res.OperationalFlags.RestartHeight)
+
```
kingpinXD, line 31: Can you add a test when operation flags are not set

x/observer/keeper/msg_server_update_operational_flags.go
kingpinXD: Yeah , but it might be usefull to have a comment here as well .

x/observer/keeper/grpc_query_operational_flags_test.go
gartnera: @kingpinXD that's the default case on line 18-21?

zetaclient/maintenance/shutdown_listener.go
gartnera: With `ticker` I'd return a sentinel error when we want to signal for shutdown right? I'm not sure I like that.

swift1337: it will return `nil`. Example: https://github.com/zeta-chain/node/blob/cfcf706956f9e01473a2e781c1bacdb44f70a938/zetaclient/orchestrator/contextupdater.go#L39

x/observer/keeper/msg_server_update_operational_flags_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/require"
+
+	keepertest "github.com/zeta-chain/node/testutil/keeper"
+	"github.com/zeta-chain/node/testutil/sample"
+	authoritytypes "github.com/zeta-chain/node/x/authority/types"
+	"github.com/zeta-chain/node/x/observer/keeper"
+	"github.com/zeta-chain/node/x/observer/types"
+)
+
+func TestMsgServer_UpdateOperationalFlags(t *testing.T) {
```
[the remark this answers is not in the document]
gartnera: I guess I could separate the "value was previously set" case from the "initial set" case. But it seems like they are closely related enough to share the same case.

kingpinXD (approved):

zetaclient/maintenance/shutdown_listener.go
gartnera: I will consider refactoring this in one of the next PRs

PR #3254, opened 2024-12-06
test(e2e): rename v2 as evm tests and rename old evm tests as legacy

lumtis:
# Description

- Rename v2 as normal EVM tests
- Move old EVM tests in legacy directory, still defined here since it is still supported but will make it easier in the future to remove
- Remove some unused and useless tests (zrc20 swap is already covered in crosschain swap for example)
- Remove V2 migration tests


ws4charlie (approved):
looks a nice move to fully switching to v2

cmd/zetae2e/local/legacy.go
swift1337, line 1: Maybe `cmd/zetae2e/local/evm_legacy.go`?

swift1337 (approved):

gartnera (approved):
CI stamp

cmd/zetae2e/local/legacy.go
lumtis: I think it's fine, just important to mention legacy to know we need to remove these in the future

PR #3262, opened 2024-12-09
ci: add option to run e2e twice

gartnera:
Add option to run default e2e tests twice in CI. Always run twice in merge queue.

We want to ensure that the default e2e tests are never dependent on the system state.

Adds ~8 minutes to mergequeue times.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new output variable `TWICE_TESTS` to conditionally run tests multiple times based on pull request labels.
	- Added a new input parameter `run-twice` to the reusable E2E testing workflow, allowing for a second execution of tests.
	- Implemented additional steps for a second test run and log monitoring based on the `run-twice` input.

- **Bug Fixes**
	- Enhanced the conditional execution logic for tests in the E2E workflow.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -123,6 +123,7 @@ jobs:
       V2_TESTS: ${{ steps.matrix-conditionals.outputs.V2_TESTS }}
       V2_MIGRATION_TESTS: ${{ steps.matrix-conditionals.outputs.V2_MIGRATION_TESTS }}
       ENABLE_MONITORING: ${{ steps.matrix-conditionals.outputs.ENABLE_MONITORING }}
+      TWICE_TESTS: ${{ steps.matrix-conditionals.outputs.TWICE_TESTS }}
     steps:
       # use api rather than event context to avoid race conditions (label added after push)
       - id: matrix-conditionals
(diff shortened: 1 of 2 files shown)


lumtis (commented):
>Adds ~8 minutes to mergequeue times.

Can't we run these concurrently instead of sequentially?

kingpinXD (commented):
Can we just the TSS MIGRATION test when merging instead ? 
It might take a bit more that just running e2e twice , but it would also cover more use cases .

gartnera:
> Can't we run these concurrently instead of sequentially?

No since the point of the second run is to ensure that the tests work given the state leftover from the second run.

> Can we just add the TSS MIGRATION test when merging instead ?

Yeah we could do that but it would take a bit longer.

lumtis (approved):
>No since the point of the second run is to ensure that the tests work given the state leftover from the second run.

I see. I'm in if it can catch edge cases

kingpinXD:
> Yeah we could do that but it would take a bit longer (looks like +5 minutes)

Just curious, is 5 minutes a significant delay here? It does provide the value of testing the TSS migration as well.

gartnera:
> Just curious, is 5 minutes a significant delay here? It does provide the value of testing the TSS migration as well. And reuse exiting code

Yeah maybe an extra 5 minutes isn't very significant. And I think the tss migrations tests may be stable enough to run in the mergequeue now too.

lumtis:
@gartnera why was this one put in draft?

gartnera:
> @gartnera why was this one put in draft?

I was going to try to get the tss migration tests running instead like Tanmay suggested. But I haven't gotten around to seeing if it will actually work yet.

PR #3277, opened 2024-12-11
refactor(e2e): use address_book.seed rather than bootstrap peer

gartnera:
Generate the address_book.seed file for zetaclient when running in localnet. This is to ensure that the localnet/e2e setup matches the testnet/mainnet zetaclient configuration where we configure an address book.

This command lives under `zetae2e` rather than `zetaclient` as this is localnet/e2e specific functionality.

Related to https://github.com/zeta-chain/go-tss/pull/40 (Remove all dynamic peer discovery)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new CLI command for retrieving bootstrap address book entries for the Zeta client.
	- Enhanced the `zetae2e` command-line interface with the new command registration.
	- Modified the `start-zetaclientd.sh` script to streamline the initialization process and retrieve the bootstrap address book.

- **Bug Fixes**
	- Improved error handling for gRPC client creation and hostname resolution in the new command.

### DIFF
--- a/cmd/zetae2e/get_zetaclient_bootstrap.go
+++ b/cmd/zetae2e/get_zetaclient_bootstrap.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/spf13/cobra"
+	"github.com/zeta-chain/node/pkg/rpc"
+	"github.com/zeta-chain/node/pkg/sdkconfig"
+	observertypes "github.com/zeta-chain/node/x/observer/types"
+	"gitlab.com/thorchain/tss/go-tss/conversion"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+func NewGetZetaclientBootstrap() *cobra.Command {
+	var ConfigureZetaclientBootstrapCmd = &cobra.Command{
+		Use:   "get-zetaclient-bootstrap",
+		Short: "get bootstrap address book entries for zetaclient",
+		RunE:  getZetaclientBootstrap,
+	}
+
+	return ConfigureZetaclientBootstrapCmd
+}
+
+func getZetaclientBootstrap(cmd *cobra.Command, _ []string) error {
+	sdkconfig.SetDefault(true)
+	rpcClient, err := rpc.NewGRPCClients(
+		"zetacore0:9090",
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+	)
+	if err != nil {
+		return fmt.Errorf("get zetacore rpc client: %w", err)
+	}
+	var res *observertypes.QueryAllNodeAccountResponse
+	for {
+		res, err = rpcClient.Observer.NodeAccountAll(cmd.Context(), &observertypes.QueryAllNodeAccountRequest{})
+		if err != nil {
+			return fmt.Errorf("get all node accounts: %w", err)
+		}
+		if len(res.NodeAccount) > 1 {
+			break
+		}
+		fmt.Fprintln(cmd.OutOrStderr(), "waiting for node accounts")
+	}
+
+	// note that we deliberately do not filter ourselfs/localhost
+	// to mirror the production configuration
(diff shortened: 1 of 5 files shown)


cmd/zetae2e/get_zetaclient_bootstrap.go
```diff
@@ -0,0 +1,97 @@
...
+		RunE:  getZetaclientBootstrap,
+	}
+
+	return ConfigureZetaclientBootstrapCmd
+}
+
+func getZetaclientBootstrap(cmd *cobra.Command, _ []string) error {
+	sdkconfig.SetDefault(true)
+	rpcClient, err := rpc.NewGRPCClients(
+		"zetacore0:9090",
```
lumtis, line 32: This is a value in the config, why not reading it from here. We can also put this command under `local` to make explicit it's only used for local testing

cmd/zetae2e/get_zetaclient_bootstrap.go
```diff
@@ -0,0 +1,97 @@
...
+		validatorRes, err := rpcClient.Staking.Validator(cmd.Context(), &stakingtypes.QueryValidatorRequest{
+			ValidatorAddr: valAddr,
+		})
+		if err != nil {
+			return fmt.Errorf("getting validator info for %s: %w", account.Operator, err)
+		}
+		// in localnet, moniker is also the hostname
+		moniker := validatorRes.Validator.Description.Moniker
+
+		peerId, err := conversion.Bech32PubkeyToPeerID(account.GranteePubkey.Secp256k1.String())
```
lumtis, line 68:
```suggestion
		peerID, err := conversion.Bech32PubkeyToPeerID(account.GranteePubkey.Secp256k1.String())
```

cmd/zetae2e/get_zetaclient_bootstrap.go
gartnera: > This is a value in the config, why not reading it from here.

The e2e config isn't in currently available in the zetanode/zetaclient container.

> We can also put this command under local to make explicit it's only used for local testing

Ok

cmd/zetae2e/get_zetaclient_bootstrap.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		valAddr := sdk.ValAddress(accAddr).String()
+		validatorRes, err := rpcClient.Staking.Validator(cmd.Context(), &stakingtypes.QueryValidatorRequest{
+			ValidatorAddr: valAddr,
+		})
+		if err != nil {
+			return fmt.Errorf("getting validator info for %s: %w", account.Operator, err)
+		}
+		// in localnet, moniker is also the hostname
+		moniker := validatorRes.Validator.Description.Moniker
```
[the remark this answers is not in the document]
gartnera: Validator and Description are not pointers. An error will be returned if validatorRes is nil (not found).

cmd/zetae2e/get_zetaclient_bootstrap.go
lumtis: If it's in local then it's fine

lumtis (approved):

cmd/zetae2e/local/get_zetaclient_bootstrap.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/zeta-chain/node/pkg/rpc"
+	"github.com/zeta-chain/node/pkg/sdkconfig"
+	observertypes "github.com/zeta-chain/node/x/observer/types"
+)
+
+const grpcURLFlag = "grpc-url"
+
+func NewGetZetaclientBootstrap() *cobra.Command {
+	var ConfigureZetaclientBootstrapCmd = &cobra.Command{
```
swift1337, line 23:
```suggestion
	var cmd = &cobra.Command{
```

swift1337 (approved):

kingpinXD (commented):
NodeAccounts list can contain accounts which are not part of the TSS yet , is it okay to use the list for the seed ? 

Okay dont hink its a problem since we are using for starting the network only

kingpinXD (approved):

gartnera:
> NodeAccounts list can contain accounts which are not part of the TSS yet , is it okay to use the list for the seed ?
> 
> Okay dont hink its a problem since we are using for starting the network only

Yes I think the address book should contain all possible peers. Only those that are needed will be actually used.

Issue #3274, opened 2024-12-10
Add back checksum format conversion when retrieving ZRC20 from asset address


lumtis:
**Describe the Issue**

We need to add back https://github.com/zeta-chain/node/pull/3124/files#diff-f6010145b17ec94c7cd761c7735a87e171346d7dc77cf49cabc8daf261ce6a83L120 while keep support for Solana SPL

lumtis:
Better fix: instead of adding back the checksum conversion in the code, we enforce asset using checksum format and use a migration script to modify all existing one to checksum format.

PR #3278, opened 2024-12-11
fix: enforce checksum format for asset address in ZRC20

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/3274

Add a check to enforce the checksum format for EVM address while whitelisting a new asset

Add a migration script for the current asset address

## Summary by CodeRabbit

- **New Features**
	- Introduced a `/systemtime` telemetry endpoint to enhance telemetry capabilities.
	- Added a `LiquidityCap` field to the `ForeignCoins` struct.
	- Implemented a migration mechanism for the fungible module to handle state updates.

- **Bug Fixes**
	- Improved error handling for unsupported transaction versions in Solana.
	- Enhanced validation for asset addresses in the `MsgWhitelistERC20` structure.

- **Refactor**
	- Revamped the TSS package for better structure and maintainability.
	- Updated the testing framework and renamed E2E tests for consistency.

- **Tests**
	- Added comprehensive unit tests for Ethereum address validation functions.
	- Introduced tests for the migration process in the fungible module.


pkg/crypto/evm_address.go
```diff
@@ -0,0 +1,30 @@
...
+	return address == (common.Address{}) || address.Hex() == constant.EVMZeroAddress
+}
+
+// IsEVMAddress returns true if the string is an EVM address
+// independently of the checksum format
+func IsEVMAddress(address string) bool {
+	return len(address) == 42 && strings.HasPrefix(address, "0x") && common.IsHexAddress(address)
+}
+
+// IsChecksumAddress returns true if the EVM address string is a valid checksum address
```
swift1337, line 22:
```suggestion
// IsChecksumAddress returns true if the EVM address string is a valid checksum address
// See https://eips.ethereum.org/EIPS/eip-55
```

swift1337 (approved):

pkg/crypto/evm_address_test.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+		{
+			name:    "all uppercase",
+			address: "0x5A4F260A7D716C859A2736151CB38B9C58C32C64",
+			want:    "0x5a4f260A7D716c859A2736151cB38b9c58C32c64",
+		},
+		{
+			name:    "empty address returns null address",
+			address: "",
+			want:    "0x0000000000000000000000000000000000000000",
```
ws4charlie, line 142: Nit: converting empty string to a coinbase `0x0000000000000000000000000000000000000000` looks a bit not that intuitive to me. If we could return simply return "", the condition `if fc.Asset != ""` can then be removed in function `MigrateStore`. wdyt

lumtis: This is the behavior of `HexToAddress` , the reasoning is I want to keep the function minimal like small brick, the check if it's non EVM address is done separatly

ws4charlie (approved):
lgtm

Issue #3260, opened 2024-12-09
zetaclient: add configurable signer latency correction


gartnera:
After https://github.com/zeta-chain/node/pull/3228, we can further improve signer synchronization by adding a configurable wait period between the block time and scheduling.

The new `OperationalFlags` section being added in https://github.com/zeta-chain/node/pull/3252 would be a good place for this parameter.

We would determine this parameter by looking at `max(zetaclient_core_block_latency)` on each network.

Related to https://github.com/zeta-chain/node/issues/3247

PR #3235, opened 2024-12-02
feat(zetaclient): add /systemtime telemetry endpoint

gartnera:
This endpoint will be used to debug issued related to bad system time values.

Related to #3227

PR #3317, opened 2024-12-18
feat: add configurable signer latency correction

gartnera:
Add a `signer_block_time_offset` field in the `OperationalFlags`. This will allow us to adjust our signer synchronization on a per network basis (we see much different latencies across mainnet and testnet).

Closes https://github.com/zeta-chain/node/issues/3260

Related to:
- https://github.com/zeta-chain/node/pull/3228
- https://github.com/zeta-chain/node/pull/3235

## Summary by CodeRabbit

- **New Features**
	- Introduced new test cases for operational flags in end-to-end tests.
	- Added new subcommands and updated existing ones in the `zetacored` CLI documentation.
	- Introduced a new property `signer_block_time_offset` in the OpenAPI specification.
	- Added a new command-line flag `signerBlockTimeOffsetFlag` for operational flags.
	- Enhanced the `Orchestrator` with a new field for dynamic signer block time offset.

- **Bug Fixes**
	- Updated validation for `SignerBlockTimeOffset` to prevent negative values and exceed limits.

- **Documentation**
	- Updated `zetacored` CLI documentation with new and modified subcommands.
	- Revised message definitions for consistency in naming conventions.

- **Tests**
	- Expanded test coverage for operational flags and their validation scenarios.


lumtis (approved):

x/observer/client/cli/tx_update_operational_flags.go
```diff
@@ -33,6 +34,7 @@ func CmdUpdateOperationalFlags() *cobra.Command {
 			flagSet := cmd.Flags()
 			file, _ := flagSet.GetString(fileFlag)
 			restartHeight, _ := flagSet.GetInt64(restartHeightFlag)
+			signerBlockTimeOffset, _ := flagSet.GetDuration(signerBlockTimeOffsetFlag)
```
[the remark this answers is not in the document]
gartnera: As per usual I think this check is pointless so long as the default on the flag is sane.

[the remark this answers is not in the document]
gartnera: No in that case this code will never be touched. Cobra will ensure that the flags are of the correct type. Here is an example:

```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

func main() {
	rootCmd := &cobra.Command{
		Use:   "app",
		Short: "An example CLI with a duration flag",
		RunE: func(cmd *cobra.Command, args []string) error {
			duration, _ := cmd.Flags().GetDuration("duration")
			fmt.Printf("Duration provided: %v\n", duration)
			return nil
		},
	}

	rootCmd.Flags().Duration("duration", 0, "Specify a duration (e.g., 1h30m)")

	if err := rootCmd.Execute(); err != nil {
		os.Exit(1)
	}
}
```

```
go run . --duration asdf
Error: invalid argument "asdf" for "--duration" flag: time: invalid duration "asdf"
Usage:
  app [flags]

Flags:
      --duration duration   Specify a duration (e.g., 1h30m)
  -h, --help                help for app

exit status 1
```

zetaclient/orchestrator/orchestrator.go
```diff
@@ -318,6 +327,13 @@ func (oc *Orchestrator) runScheduler(ctx context.Context) error {
 				continue
 			}
 
+			sleepDuration := time.Until(newBlock.Block.Time.Add(oc.signerBlockTimeOffset))
```
swift1337, line 330: Can we abstract this from orchestrator and move it to zetacore client?  E.g.

```
oc.zetacoreClient.NewBlockSubscriber(ctx, true) // syncBlock=true
```

gartnera: Maybe but I'm not sure if we'll want the same latency offset for different processes.

swift1337: > if we'll want the same latency offset for different processes.

In this case, you can add the third param `jitter time.Duration` that would simulate some randomness. 

But the point is that the orchestrator should care about observers/signers/ and this logic is a bit out of its responsibilities IMO

kingpinXD (approved):

Issue #3299, opened 2024-12-16
Implement Scheduler component


swift1337:
> Don't confuse it with cctx scheduling
> 

We have a lot of repeated logic where we want to run certain tasks in the background on a predefined interval. We might also want to skip certain runs (e.g., `app.IsOutboundObservationEnabled()`). Additionally, the interval can be changed dynamically.

For these use cases, we have `bg` and `ticker` packages, but we lack a combined component that also supports metric reporting per run (e.g., last run, latency, outcome, ...). Thus, it's very handy to have something like this:

```go
type Task interface {
	// ...
}

type Scheduler interface {
	Register(Task, Opts...)
}
```

Each ObserverSigner would receive Scheduler in `Start(ctx context.context, s Scheduler) error` and register its jobs there

```go
func (os *Solana) Start(ctx context.Context, s *scheduler.Scheduler) error {
   s.Register(os.observer.ObserverInbound, /* options... */)
   s.Register(os.observer.ObserveGasPrice, /* options... */)
   s.Register(os.ScheduleOutbounds, /* options... */)
   
   return nil
}
```

![image](https://github.com/user-attachments/assets/31e85c59-dd66-47ab-afce-30b4862ee00a)

```[tasklist]
### Tasks
- [ ] Scheduler
- [ ] Scheduler groups
- [ ] Test Coverage
- [ ] Automatic metrics provisioning
```

PR #3319, opened 2024-12-19
feat(pkg): scheduler package

swift1337:
`pkg/scheduler` provides a background task scheduler that allows for the registration, execution, and management of periodic tasks. 

Tasks can be grouped, named, and configured with various options such as custom intervals, log fields, and skip conditions.

The scheduler supports dynamic interval updates and can gracefully stop tasks either individually or by group.

## Features
- Task registration
- Task configuration (interval updates, skip invocation, etc..)
- Task grouping
- Block-based tickers 🔥  
- Graceful shutdown

## Changes
- [x] Scheduler package
- [x] Minor improvements in `ticker` package
- [x] Add `ticker.StopBlocking()`
- [x] Metrics

Closes https://github.com/zeta-chain/node/issues/3299


pkg/scheduler/chan.go
```diff
@@ -0,0 +1,117 @@
...
+	blockChan <-chan cometbft.EventDataNewBlock
+
+	// stopChan is used to stop the ticker
+	stopChan chan struct{}
+
+	// doneChan is used to signal that the ticker has stopped (i.e. "blocking stop")
+	doneChan chan struct{}
+
+	// atomic flag. `1` for RUNNING, `0` for STOPPED
+	status int32
```
gartnera, line 27: Any reason not to use `atomic.Int32{}` directly?

pkg/scheduler/scheduler.go
gartnera, line 1: Not sure I like mixing the timer based tickers and block ticker logic. Maybe a common interface would be a better idea (implementing Stop() for example) would work better?

pkg/scheduler/opts.go
```diff
@@ -0,0 +1,46 @@
...
+// IntervalUpdater sets interval updater function.
+func IntervalUpdater(intervalUpdater func() time.Duration) Opt {
+	return func(d *Definition) { d.intervalUpdater = intervalUpdater }
+}
+
+// BlockTicker makes Definition to listen for new zeta blocks instead of using interval ticker.
+// IntervalUpdater is ignored.
+func BlockTicker(blocks <-chan cometbft.EventDataNewBlock) Opt {
+	return func(d *Definition) { d.blockChan = blocks }
+}
```
gartnera, line 46: I think a run every N blocks ticker would be useful too.

Just mod block number == 0 right?

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,234 @@
...
+	// skip tick
+	if d.skipper != nil && d.skipper() {
+		return nil
+	}
+
+	d.logger.Debug().Msg("Invoking task")
+
+	err := d.task(ctx)
+
+	// todo metrics (TBD)
```
swift1337, line 208: @gartnera would be great to hear your thoughts on this

gartnera (commented):
Note https://github.com/zeta-chain/node/pull/3317

pkg/scheduler/scheduler.go
gartnera: Yeah I think we add one metric then use the task name as a label.

Metrics for:
- count of times run
- error rate
- duration

pkg/scheduler/opts.go
swift1337: I think this can be implemented as a separate "middleware" or Opt that wraps chan with another chan, etc. Block ticker should not be aware of this.

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
+// Scheduler represents background task scheduler.
+type Scheduler struct {
+	definitions map[uuid.UUID]*Definition
+	mu          sync.RWMutex
+	logger      zerolog.Logger
+}
+
+// Task represents scheduler's task
```
lumtis, line 29: Can we maybe be more precise on this one? What differentiates a task from a normal function?

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,234 @@
...
+
+// New Scheduler instance.
+func New(logger zerolog.Logger) *Scheduler {
+	return &Scheduler{
+		definitions: make(map[uuid.UUID]*Definition),
+		logger:      logger.With().Str("module", "scheduler").Logger(),
+	}
+}
+
+// Register registers and starts new task in the background
```
lumtis, line 75: Can we explain the returned value? Definition is configuration of a Task, to me a configuration is something that would be given as an input

lumtis (approved):
LGTM, the API looks clear to me, maybe some comments can be more explicative

NIT PR is named `feat(zetaclient)` but we don't do any ZetaClient change here

pkg/scheduler/scheduler.go
swift1337: UPD: simplified the package, added `tickable` interface

pkg/scheduler/scheduler.go
swift1337: https://github.com/zeta-chain/node/pull/3319/commits/6cc8e435b0b677fe004c50d97f6480d9aaf85326

ws4charlie (approved):
looks cool.

zetaclient/metrics/metrics.go
```diff
@@ -170,6 +170,27 @@ var (
...
+		[]string{"status", "task_group", "task_name"},
+	)
+
+	// SchedulerTaskExecutionDuration measures the execution duration of tasks
+	SchedulerTaskExecutionDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: ZetaClientNamespace,
+			Name:      "scheduler_task_duration_seconds",
+			Help:      "Histogram of task execution duration in seconds",
+			Buckets:   []float64{0.05, 0.1, 0.2, 0.3, 0.5, 1, 1.5, 2, 3, 5, 7.5, 10, 15}, // 50ms to 15s
```
gartnera, line 190: `prometheus.DefBuckets` is probably fine

swift1337: I think it's not suitable in this case because most likely the latency will come from RPC networking which is definitely more than`5ms`

```go
// 5ms to 10s
var DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

Issue #3308, opened 2024-12-17
Potentially missed observation from Polygon


fadeev:
https://polygonscan.com/tx/0x06a0ec87f5c6dab11eed5da1357035590d7d54559679c4b1dc3d07264cb51b30

lumtis:
Found the tx, the validation can't be processed https://zetachain.explorers.guru/transaction/764A5A9207CC844EC3A26300C3FB8C7B35E58E194EF25E5CD8C03FB68E6E9614

On the condition that produces the error is that it seems we're doing a BTC withdraw in the crosschain call that doesn't work because it is less than theBTC dust amount.

On the error itself, it seems that under this workflow, the observation tx fails at the Cosmos level, failing the whole txs making the processing of the inbound impossible

nzmpi:
@lumtis That's my tx. Thank you for you answer!
Does this mean that if I am trying to send less BTC than dust amount (1000?) the transaction will revert, and without calling `OnRevert` on the original EVM chain? So those funds are just lost?

lumtis:
No, it is not lost, it is blocked until there is a fix applied.
Did you make other deposits that didn't get observed?

nzmpi:
> Did you make other deposits that didn't get observed?

Not yet, do I need to update my contract on Zetachain to prevent this in future? Like reverting if a user tries to send too little BTC?

lumtis:
We need to upgrade the protocol to fix the workflow and make the CCTX revert, once implemented this currently blocked CCTX will get observed and reverted.
But before upgrade, yes, you can add a check in the smart contract. The Bitcoin dust amount is 1000sat, if you add a `require(withdrawAmount > 1000)` then the CCTX should be reverted. This check will be no longer necessary after Bitcoin upgrade.

PR #3321, opened 2024-12-20
fix: make crosschain-call with invalid withdraw revert

lumtis:
# Description

Solves https://github.com/zeta-chain/node/issues/3308

If a crosschain call initiate an invalid withdraw (example dust amount for Bitcoin withdraw) makes the CCXT reverts

**Solution**

In `HandleEVMCall` where crosschain calls are handled, when `ProcessLogs` logs fails, we consider that the crosschain call reverts because the withdraw initiated in this call can't be processed, we therefore returns `contractReverted == true` so it enters the revert workflow.

Since in this case the deposit contract call was successful and there is an effective state transition in the EVM, we now use `tmpCtx`, we only commit it if the contract has a regular revert or the call is sucessful and `ProcessLogs` doesn't fail.

The E2E test checks that the BTC balance remains 0 after the calls, this means the EVM state transition has not been processed.

## Release Notes

- **New Features**
	- Introduced a new telemetry endpoint `/systemtime` in the application.
	- Added a crosschain-call feature to handle invalid withdraw reverts.
	- New simulation tests for custom zetachain modules and performance tests for Solana E2E.

- **Bug Fixes**
	- Fixed issues with unsupported Solana transaction versions and inbound vote message validations.
	- Updated handling of pending nonces during transaction abortion.

- **Documentation**
	- Updated changelog to reflect new features, tests, and fixes.

- **Tests**
	- Added a new test for Bitcoin deposit and withdrawal operations involving dust amounts.


e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
```diff
@@ -0,0 +1,58 @@
...
+// It tests the edge case where during a cross-chain call, a invaild withdraw is initiated (processLogs fails)
+func TestBitcoinDepositAndWithdrawWithDust(r *runner.E2ERunner, args []string) {
+	// Given "Live" BTC network
+	stop := r.MineBlocksIfLocalBitcoin()
+	defer stop()
+
+	require.Len(r, args, 0)
+
+	// ARRANGE
+
```
ws4charlie, line 25:
```suggestion
	// ARRANGE
```

lumtis: What's that?

ws4charlie: extra new line

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
```diff
@@ -0,0 +1,58 @@
...
+	// ASSERT
+	// Now we want to make sure the cctx is reverted with expected error message
+
+	// cctx status would be pending revert if using v21 or before
+	cctx := utils.WaitCctxRevertedByInboundHash(r.Ctx, r, txHash.String(), r.CctxClient)
+	require.Equal(r, crosschaintypes.CctxStatus_Reverted, cctx.CctxStatus.Status)
+	require.Contains(r, cctx.CctxStatus.ErrorMessage, crosschaintypes.ErrCannotProcessWithdrawal.Error())
+
+	// check the contract has no BTC balance, this would mean the contract call state transition is not reverted
```
ws4charlie, line 53: the expected behavior is `not reverted` or `reverted`?

x/crosschain/keeper/evm_deposit.go
```diff
@@ -145,7 +154,11 @@ func (k Keeper) HandleEVMDeposit(ctx sdk.Context, cctx *types.CrossChainTx) (boo
 				)
 			}
 		}
+
```
kingpinXD, line 157: Consider putting lines 105 - 157 in a block , `{ }` 
For readability to show, where the context is branched and committed back

kingpinXD (approved):
Great catch , PR looks good to me

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
```diff
@@ -0,0 +1,58 @@
...
+	require.Equal(r, receipt.Status, uint64(1))
+
+	// Given a list of UTXOs
+	utxos, err := r.ListDeployerUTXOs()
+	require.NoError(r, err)
+	require.NotEmpty(r, utxos)
+
+	// ACT
+	// Deposit 0.01 BTC to withdrawer, this is an arbitrary amount, must be greater than dust amount
+	txHash, err := r.SendToTSSFromDeployerWithMemo(0.01, utxos, withdrawerAddr.Bytes())
```
ws4charlie, line 41: This looks to me a pure `deposit` of 0.01 BTC. How would the `onCrossChainCall` on withdrawer contract be called in this scenario? a bit confused

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
lumtis: There is no change in your suggestion

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
lumtis: Need to make the comment more clear

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
lumtis: Mmh weird, the logs shows smart contract call happens

lumtis: Ok, just checked, independently of the message, we do a contract call if dest is a contract in the V1 workflow

x/crosschain/keeper/evm_deposit.go
lumtis: I don't think you should use block if it doesn't have reason in the workflow

e2e/e2etests/test_bitcoin_deposit_and_withdraw_with_dust.go
ws4charlie: okay. That makes sense then.

ws4charlie (approved):
lgtm

lumtis:
Received a security review

PR #3342, opened 2025-01-09
fix!: set per transaction gas limit to 10M

gartnera:
Set the per-transaction gas limit so that the 99.9th-percentile transaction is allowed, but the maximum gas-value transaction is not. A limit of 10M is significantly higher than the 99.9th-percentile value to avoid disrupting legitimate transactions.

This will ensure that ridiculously large transactions do not degrade the network performance.

Will backport to develop.

image

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -143,8 +143,9 @@ func init() {
 }
 
 var (
-	NodeDir         = ".zetacored"
-	DefaultNodeHome = os.ExpandEnv("$HOME/") + NodeDir
+	NodeDir                    = ".zetacored"
+	DefaultNodeHome            = os.ExpandEnv("$HOME/") + NodeDir
+	TransactionGasLimit uint64 = 10_000_000
 )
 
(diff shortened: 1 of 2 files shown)


gartnera:
> Wouldn't this break the consensus?

Yes that's why we're doing it before releasing v25.0.0

It wouldn't hit the CI check though since we don't have anything that actually used that much gas in CI

brewmaster012 (approved):

brewmaster012:
LGTM just a note: 
what happens when a tx requests bigger than 10M gas limit? 
If it's an EVM tx?

gartnera:
> LGTM just a note: what happens when a tx requests bigger than 10M gas limit? If it's an EVM tx?

It should be reverted but I have not specifically tested that. I could test it pretty easily on the EVM side but I'm not sure about the cosmos side.

ws4charlie (approved):

kingpinXD (approved):

PR #3345, opened 2025-01-10
chore: upgrade cosmovisor and go

gartnera:
https://github.com/cosmos/cosmos-sdk/releases/tag/cosmovisor%2Fv1.7.0

Requires go 1.23 for some reason.

goreleaser-cross hasn't published a 1.23.4 so sticking with 1.23.3

## Summary by CodeRabbit

- **Chores**
	- Updated `cosmovisor` tool from version 1.6.0 to version 1.7.0 in the build process

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -33,7 +33,7 @@ runs:
     - uses: actions/setup-go@v5
       if: ${{ inputs.skip_go == 'false' }}
       with:
-        go-version: '1.22'
+        go-version: '1.23'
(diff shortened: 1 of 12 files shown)


lumtis (approved):
Wondering, any benefits we would get from upgrading the repo to 1.23 as well at some point?

gartnera:
> Wondering, any benefits we would get from upgrading the repo to 1.23 as well at some point?

Until we import a package that specifically requires and uses a 1.23 feature no.

kingpinXD (approved):

PR #3346, opened 2025-01-10
test: disable preflight check so we can see failure message in explorer

ws4charlie:
# Description

Skip pre-flight check to force send the deposit so we can see details of the failure for debugging purpose.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


ws4charlie:
close this PR, @skosito identified that the issue was caused by compute unit and price

Issue #3328, opened 2024-12-30
zetaclient stops connecting to external chain RPCs after receiving HTTP error codes. 


CharlieMc0:
**Describe the Bug**
When zetaclientd gets an HTTP error back from the BTC RPC it stops working and never retries until zetaclientd is restarted. 

**To Reproduce**
Would need to replicate the the 409 or other HTTP errors codes from the BTC RPC. It isn't clear if this is limited to BTC or an issue for the RPC connectivity for all networks. 

**Expected Behavior**
Ideally all attempts to connect to RPCs for external networks should have a backoff timer and continue to retry even if they receive errors from the external RPC server.  

**Screenshots**

<img width="1058" alt="Screenshot 2024-12-29 at 6 07 06 PM" src="https://github.com/user-attachments/assets/c9816062-0fb2-4854-b00e-bce701063683" />

You can see from the logs one one validator was falling behind on BTC blocks
<img width="610" alt="Screenshot 2024-12-29 at 6 07 32 PM" src="https://github.com/user-attachments/assets/49f20736-2bc1-403d-8121-3a2dc415d5a8" />
<img width="582" alt="Screenshot 2024-12-29 at 6 11 04 PM" src="https://github.com/user-attachments/assets/fc23af34-36e7-4ece-af32-51e9cc6f8a9f" />

gartnera:
Both of the places were `FetchUTXOs` is called do not exit if error. So if this is really happening then the log messages are probably unrelated.

What we really need in cases like this is a goroutine dump via `curl -o goroutine.dump http://localhost:6061/debug/pprof/goroutine?debug=2`

https://github.com/zeta-chain/node/blob/80ca9214bf531668b0b18199f0743fa81957f3ca/zetaclient/chains/bitcoin/observer/observer.go#L308-L319

https://github.com/zeta-chain/node/blob/80ca9214bf531668b0b18199f0743fa81957f3ca/zetaclient/chains/bitcoin/signer/signer.go#L206-L213

ws4charlie:
Seems we only kept 7 days of logs in Datadog and above log prints are not tracked any more. After searching above log prints in code, the initial error was on Bitcoin RPC `GetBlockCount` (error getting block height), and then it failed `FetchUTXOs`. When `FetchUTXOs` fails, zetaclient won't be able to get balance and outbound will not work properly.

[...]

PR #3349, opened 2025-01-13
feat(btc): implement new btc rpc package

swift1337:
This PR implements Bitcoin client in Go-"native" way
-  It leverages existing `btcsuite` types, but removes all redundant features such as websockets or backward compatibility with old BTC versions that are not used anymore
-  It ensures `ctx` for each call 
- The over-complicated `chan`-based design is replaced with a simple "stateless" request-response approach
- It also fixed a significant drawback of the original RPC: lack of `ctx` might **block** execution indefinitely & block other routines (see [this](https://github.com/btcsuite/btcd/blob/master/rpcclient/infrastructure.go#L940) and [this](https://github.com/btcsuite/btcd/blob/master/rpcclient/infrastructure.go#L89))

---

- [x] Implement BTC Client
- [x] Port methods from `zetaclient/chains/bitcoin/rpc`
- [x] Implement prometheus metrics
- [x] Port live tests 
- [x] Replace usage `zetaclient/chains/bitcoin/rpc` & `interfaces.BTCRPCClient` with the new client
- [x] Fix unit tests
- [x] Fix e2e tests

### One of examples

**Before**
```go
// zetaclient/chains/bitcoin/rpc.go
func GetTxResultByHash(
	rpcClient interfaces.BTCRPCClient,
	txID string,
) (*chainhash.Hash, *btcjson.GetTransactionResult, error) {
	hash, err := chainhash.NewHashFromStr(txID)
	if err != nil {
		return nil, nil, errors.Wrapf(err, "GetTxResultByHash: error NewHashFromStr: %s", txID)
	}

	// The Bitcoin node has to be configured to watch TSS address
	txResult, err := rpcClient.GetTransaction(hash)
	if err != nil {
		return nil, nil, errors.Wrapf(err, "GetTxResultByHash: error GetTransaction %s", hash.String())
	}
	return hash, txResult, nil
}
```

**After**
```go
tx, err := btcClient.GetTransaction(ctx, txHash)

// or

tx, err := btcClient.GetTransactionByStr(ctx, "my-string-hash")
```

Related: 
- https://github.com/zeta-chain/infrastructure/issues/1936
- https://github.com/zeta-chain/node/issues/3328

## Summary by CodeRabbit

Based on the comprehensive summary, here are the release notes:

## Release Notes

- **New Features**
	- Implemented a new Bitcoin RPC client with improved performance and observability
	- Enhanced context handling for Bitcoin-related operations
	- Added new metrics for tracking RPC client interactions

- **Improvements**
	- Streamlined Bitcoin client interfaces
	- Updated Bitcoin RPC method signatures to support context management
	- Improved error handling and logging for Bitcoin-related functionalities

- **Changes**
	- Replaced existing Bitcoin RPC client implementation
	- Updated mock clients for Bitcoin RPC testing
	- Removed legacy Bitcoin RPC interfaces

- **Performance**
	- Added metrics tracking for RPC client calls
	- Optimized Bitcoin client interactions

- **Testing**
	- Enhanced test coverage for Bitcoin client operations
	- Updated test utilities to support new client implementation


zetaclient/chains/bitcoin/client/client.go
```diff
@@ -0,0 +1,231 @@
...
+		c.recordMetrics(method, start, out, err)
+		c.logger.Debug().Err(err).
+			Str("rpc.method", method).Dur("rpc.duration", time.Since(start)).
+			Msg("Sent request")
+	}()
+
+	res, err := c.client.Do(req)
+	if err != nil {
+		return rawResponse{}, errors.Wrap(err, "unable to send the request")
+	}
```
gartnera, line 137: I think there should be a response code is 200 check here too?

gartnera (commented):
Generally LGTM.

If you directly copy pasted something from btcd please make sure to notate the original license on the file.

zetaclient/chains/bitcoin/client/client_test.go
gartnera, line 1: I wonder if there is a way to actually run this in CI? Maybe just throw it on the e2e side? Not sure if it's actually terribly important since the methods will have coverage on e2e anyway.

zetaclient/metrics/metrics.go
gartnera, line 1: We already have the `rpc_count` and `rpc_duration_seconds` why can't we use those metrics?

zetaclient/chains/bitcoin/client/client.go
```diff
@@ -0,0 +1,270 @@
...
+	params, err := resolveParams(cfg.RPCParams)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to resolve chain params")
+	}
+
+	clientName := fmt.Sprintf("btc:%d", chainID)
+
+	c := &Client{
+		hostURL:    normalizeHostURL(cfg.RPCHost, true),
+		client:     defaultHTTPClient(),
```
gartnera, line 88: use `metrics.GetInstrumentedHTTPClient` to get a http client with automatic metrics? related to my comment on metrics.go

zetaclient/chains/bitcoin/client/client.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// send sends RPC command to the server via http post request
+func (c *Client) sendCommand(ctx context.Context, cmd any) (json.RawMessage, error) {
+	method, reqBody, err := c.marshalCmd(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to marshal cmd")
+	}
+
+	// ps: we can add retry logic if needed
```
gartnera, line 112: imo the best place to do retries in our case is from the top down (orchestrator/scheduler level)

zetaclient/chains/bitcoin/client/client.go
```diff
@@ -0,0 +1,270 @@
...
+
+	if c.config.RPCPassword != "" || c.config.RPCUsername != "" {
+		req.SetBasicAuth(c.config.RPCUsername, c.config.RPCPassword)
+	}
+
+	return req, nil
+}
+
+func (c *Client) sendRequest(req *http.Request, method string) (out rawResponse, err error) {
+	c.logger.Debug().Str("rpc.method", method).Msg("Sending request")
```
gartnera, line 148: somewhat nitpicky but it would be nice if we injected the logger into the context and extracted it when logging so that the log is attributed to the correct subsystem.

zetaclient/chains/bitcoin/observer/observer.go
```diff
@@ -26,6 +28,43 @@ import (
 	clienttypes "github.com/zeta-chain/node/zetaclient/types"
 )
 
+type RPC interface {
```
gartnera, line 31: I think I preferred this in the separate interfaces package. Or at least next to the client.

It's useful to have an assertion that a type implements the actual interface (example: `var _ GasTx = (*legacytx.StdTx)(nil)`). With this structure I don't think it's possible because it's a circular import.

gartnera (changes requested):
First pass, I will try to look more later. Generally LGTM but I think the metrics should be consistent with our existing approach.

zetaclient/chains/bitcoin/client/commands.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil, errors.Wrap(err, "unable to marshal params")
+	}
+
+	body := fmt.Sprintf(
+		`{"jsonrpc":"%s","id":%d,"method":"%s","params":%s}`,
+		rpcVersion,
+		commandID,
+		method,
+		paramsBytes,
+	)
```
gartnera, line 412: coderabbit already noted this is an issue but I'm opening another thread because this is a must fix issue.

zetaclient/metrics/metrics.go
swift1337:
```go
RPCCount = promauto.NewCounterVec(
		prometheus.CounterOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_count",
			Help:      "A counter for number of total RPC requests",
		},
		[]string{"host", "code"},
	)

	// RPCLatency is a histogram of the RPC latency
	RPCLatency = promauto.NewHistogramVec(
		prometheus.HistogramOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_duration_seconds",
			Help:      "A histogram of the RPC duration in seconds",
			Buckets:   prometheus.DefBuckets,
		},
		[]string{"host"},
	)
```
vs
```go
	RPCClientCounter = promauto.NewCounterVec(
		prometheus.CounterOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_client_calls_total",
			Help:      "Total number of rpc calls",
		},
		[]string{"status", "client", "method"},
	)

	RPCClientDuration = promauto.NewHistogramVec(
		prometheus.HistogramOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_client_duration_seconds",
			Help:      "Histogram of rpc client calls durations in seconds",
			Buckets:   []float64{0.05, 0.1, 0.2, 0.3, 0.5, 1, 1.5, 2, 3, 5, 7.5, 10, 15}, // 50ms to 15s
		},
		[]string{"status", "client", "method"},
	)
```

the latter support status (`ok`), client (`btc:18443`), and method (`getbestblockhash`). I plan to use them for all RPC clients to have 2 metrics that cover all basic needs.

zetaclient/chains/bitcoin/client/client.go
swift1337: It's not usefull for RPC as it has only 1 endpoint instead of /eth_this, /eth_that, ...

zetaclient/chains/bitcoin/client/client.go
swift1337: Let's revisit in future if needed

zetaclient/chains/bitcoin/observer/observer.go
swift1337: I don't see this as an issue at all. IMO this is the best idiomatic way of doing this. Observer doesn't need to care who is a client, only what it can do. The rest is a concern of someone who calls `NewObserver`.

Also, we have linters, test cases, and basically the code won't compile if something is not implemented.

zetaclient/metrics/metrics.go
gartnera: Especially on the histogram metric that's really huge cardinality if you track every method. That would probably increase our metrics storage size like 20-50%.

I understand this is more valuable than the high level metrics. Let's talk about it tomorrow morning.

swift1337: **UPD**: Yeah, I get your args from the SRE perspective. 

Let's keep it this way (below). IMO this is a good balance between metrics being useful, and insightful, but not too resource-consuming:

```go
        // left as is
	RPCClientCounter = promauto.NewCounterVec(
		prometheus.CounterOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_client_calls_total",
			Help:      "Total number of rpc calls",
		},
		[]string{"status", "client", "method"},
	)

        // cardinality decreased significantly 
        // (x2 for `status`, and maybe x10 for `method`)
	RPCClientDuration = promauto.NewHistogramVec(
		prometheus.HistogramOpts{
			Namespace: ZetaClientNamespace,
			Name:      "rpc_client_duration_seconds",
			Help:      "Histogram of rpc client calls durations in seconds",
			Buckets:   []float64{0.05, 0.1, 0.2, 0.3, 0.5, 1, 1.5, 2, 3, 5, 7.5, 10, 15}, // 50ms to 15s
		},
		[]string{"client"},
	)
```

So each RPC client will have `client` label that's actually `$chain_type:$chain_id` (`evm:1`, `sol:900`, ...)

zetaclient/chains/bitcoin/client/client_test.go
swift1337: Yeah, I think generally we can have a `LIVE_TESTS` label that sets necessary env vars and runs live tests. It would be good to run specific packages instead of `go test ./...` to avoid testing regular unit tests.

swift1337: https://github.com/zeta-chain/node/issues/3359

zetaclient/metrics/metrics.go
swift1337: https://github.com/zeta-chain/node/pull/3349/commits/8705c9af446b48f1049b44f832eabf13b1b3ea68

lumtis (approved):
Looks good to me
Would be great @ws4charlie you have a look

zetaclient/metrics/metrics.go
gartnera: This is better. But I still think we should still try to be consistent on how we do this across all our rpc clients. So either remove the existing metrics and adapt the clients that use them or use the existing metrics and remove the new ones.

ws4charlie (approved):
looks good to me. Can we merge it asap so that I can resolve conflicts and split my PR into small ones?

zetaclient/metrics/metrics.go
gartnera: https://github.com/zeta-chain/node/issues/3367

gartnera (approved):

Issue #3296, opened 2024-12-13
/cosmos/consensus/v1/params is missing from the swagger api docs


gartnera:
https://zetachain-testnet-api.itrocket.net/cosmos/consensus/v1/params

gartnera:
Yep it seems to be there on the v50 branch:

<img width="2032" alt="Image" src="https://github.com/user-attachments/assets/5146f92d-371a-4331-9735-c62cb6131cf6" />

Issue #3219, opened 2024-11-26
Upgrade Cosmos SDK to v0.50


lumtis:
**Describe the Issue**

Guide: https://docs.cosmos.network/main/build/migrations/upgrading#v050x

Will enable fast governance proposals

PR #3357, opened 2025-01-16
feat: cosmos v0.50 upgrade

skosito:
# Description

Upgrades cosmos to v50.
Ethermint side: https://github.com/zeta-chain/ethermint/pull/179

Closes:
- https://github.com/zeta-chain/node/issues/3219
- https://github.com/zeta-chain/node/issues/3296

Note: mockery is broken for authority module as it can't mock inline interface for sdk.Msg

Also need to check if changes in simulation tests are working.

Some parts like server/start.go should be reviewed in more details to ensure something is not missing.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


rpc/namespaces/ethereum/eth/filters/api.go
```diff
@@ -353,6 +355,8 @@ func (api *PublicFilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, er
 		return &rpc.Subscription{}, err
 	}
 
+	// TODO: use events
```
lumtis, line 358: Can we track this TODO somewhere?
What does it implies not using event currently, what benefices we'll get from events in the future?

rpc/websockets.go
```diff
@@ -427,6 +428,8 @@ func (api *pubSubAPI) subscribeNewHeads(wsConn *wsConn, subID rpc.ID) (pubsub.Un
 		return nil, errors.Wrap(err, "error creating block filter")
 	}
 
+	// TODO: use events
```
lumtis, line 431: Same

lumtis (commented):
I was initially expected `sdk.Context` to be replaced by `context.Context` everywhere but this is finally not needed?

app/export.go
```diff
@@ -75,21 +79,35 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []str
 	/* Handle fee distribution state. */
 
 	// withdraw all validator commission
-	app.StakingKeeper.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) (stop bool) {
-		_, err := app.DistrKeeper.WithdrawValidatorCommission(ctx, val.GetOperator())
+	err := app.StakingKeeper.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) (stop bool) {
+		valAddr, err := sdk.ValAddressFromBech32(val.GetOperator())
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
swift1337: Let's switch to error.Wrap() because this is called during the runtime, not during app init

cmd/zetae2e/local/monitor_block_production.go
```diff
@@ -46,7 +46,7 @@ func monitorBlockProduction(ctx context.Context, conf config.Config) error {
 				return fmt.Errorf("expecting new block event, got %T", event.Data)
 			}
 			latestNewBlockEvent = newBlockEvent
-		case <-time.After(5 * time.Second):
+		case <-time.After(10 * time.Second):
```
swift1337, line 49: Do we have a const of zeta block time? It would be good to use smth like `zetaBlockTime + 3 * time.Second)

pkg/mempool/custom_proposal_handler.go
```diff
@@ -145,7 +145,7 @@ func (h *CustomProposalHandler) PrepareProposalHandler() sdk.PrepareProposalHand
 			if err != nil {
 				err := h.mempool.Remove(memTx)
 				if err != nil && !errors.Is(err, mempool.ErrTxNotFound) {
-					panic(err)
+					return nil, err
```
swift1337, line 148: error wrap?

scripts/mocks-generate.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Install mockery
-go install github.com/vektra/mockery/v2@v2.38.0
+go install github.com/vektra/mockery/v2@v2.45.0
```
swift1337, line 4: Maybe we should you the latest version? 2.50+

server/json_rpc.go
```diff
@@ -35,6 +34,10 @@ import (
 	"github.com/zeta-chain/node/server/config"
 )
 
+const (
+	ServerStartTime = 5 * time.Second
```
swift1337, line 38: Can you elaborate on this one?

lumtis:
> it seems to be replaced in cosmos-sdk modules, but when code interacts with it (eg. in hooks) it can unwrap sdk.Ctx from context.Context
> 
> we can also refactor our modules i guess, but seems out of scope

Ok, let's keep it less complex

cmd/zetae2e/local/monitor_block_production.go
gartnera: No the block time can be quite variable. I think 10 is a bit more stable number.

contrib/localnet/scripts/start-zetaclientd.sh
```diff
@@ -44,7 +44,7 @@ done
 # need to wait for zetacore0 to be up
 while ! curl -s -o /dev/null zetacore0:26657/status ; do
     echo "Waiting for zetacore0 rpc"
-    sleep 10
+    sleep 20
```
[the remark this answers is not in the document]
gartnera: Yeah this doesn't seem to be needed

gartnera (approved):
Nothing looks too crazy in this PR. I think we should move to get it merged as fast as possible to avoid conflicts.

We do need to get the ethermint integration tests running before the next release though.

lumtis:
> Nothing looks too crazy in this PR. I think we should move to get it merged as fast as possible to avoid conflicts.
> 
> We do need to get the ethermint integration tests running before the next release.

Agree on this

contrib/localnet/scripts/start-zetaclientd.sh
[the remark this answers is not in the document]
gartnera: Not blocking merge but I will revert it eventually

gartnera (approved):

swift1337 (approved):

PR #3360, opened 2025-01-16
refactor: update import with new protocol contract path

lumtis:
# Description

- Update to the latest commit of protocol contract
- Move Uniswap binding to this repo directly
- Update all imports


pkg/contracts/uniswap/v2-core/contracts/uniswapv2pair.sol/uniswapv2pair.go
```diff
@@ -0,0 +1,1864 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package uniswapv2pair
```
swift1337, line 4: are these codegens ported on purpose?

swift1337 (approved):

gartnera (approved):

pkg/contracts/uniswap/v2-core/contracts/uniswapv2pair.sol/uniswapv2pair.go
lumtis: We don't make any modification to the Uniswap contract, just imported the code and generated binding.
It is easier moving the binding directly in pkg and removing it from `protocol-contracts` since it requires an older version of Solidity to be built

Issue #3339, opened 2025-01-08
`zetaclient` `zetatool` : command to fetch ballot index from inbound hash and chain id 


kingpinXD:
We have an initial implementation of the command under zetaclient 
https://github.com/zeta-chain/zeta-node/blob/47c9444bbd115176a2736f31be8cb7d23b0059e7/cmd/zetaclientd/inbound.go#L28-L190

This should be improved to 
- Work with solana inbound 
- Moved to the zetatool directory 
- The command should standalone, and contain all the necessary configuration. It should not require a zetaclient config to function.

Expected Outcome 
- A user should be able to use a cli command such as  
get-ballot [inbound hash] [chain-id] , which returns a ballot identifier .

PR #3358, opened 2025-01-16
feat: register aborted CCTX for Bitcoin inbound that carries insufficient depositor fee

ws4charlie:
# Description

Solution:
To allow users to be able to track failed inbounds caused by low depositor fee, we register an aborted CCTX with `zero` amount and and proper error message.

Some considerations:

- Creating extra CCTX will cause extra inbound votes from observers, therefore incurs extra gas and traffic.
- Creating extra CCTX will require extra chain state and storage.
- Considering that end users (depositors) have already paid gas fees in Bitcoin, flooding Zetachain with low-value deposits will not be a cheap spamming.

Changes:
- `zetaclient` sets inbound status `InboundStatus_INSUFFICIENT_DEPOSITOR_FEE` in the `MsgVoteInbound` struct if deposit amount is lower then depositor fee.
- In the`zetacore`, it creates a new CCTX struct with field `InboundParams.Status` initialized as `InboundStatus_INSUFFICIENT_DEPOSITOR_FEE`.
- In the `zetacore`, the `CCTXGatewayZEVM` registers an aborted CCTX when it sees above status.
- E2E & unit tests.

Closes: https://github.com/zeta-chain/node/issues/3193

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Tested in development environment
- [x] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions

## Summary by CodeRabbit

Based on the comprehensive summary, here are the release notes:

## Release Notes

- **New Features**
	- Enhanced error handling for Bitcoin inbound transactions with insufficient depositor fees
	- Added ability to register and track aborted cross-chain transactions (CCTX)

- **Improvements**
	- Updated cross-chain transaction vote message to include error reporting
	- Improved Bitcoin transaction processing with more detailed error messages
	- Enhanced logging and traceability for cross-chain transaction failures

- **Testing**
	- Added new end-to-end test cases for Bitcoin deposit scenarios
	- Expanded test coverage for cross-chain transaction handling

- **Bug Fixes**
	- Implemented robust handling of transactions with insufficient fees
	- Improved error tracking for cross-chain transaction observations

PR #3368, opened 2025-01-16
feat(`zetatools`): cli command to fetch inbound ballot from inbound hash and move to zetatools

kingpinXD:
# Description

- Adds a cli command under zetatool which can print the ballot-identifier / cctx -identifier , from the inbound hash. 
- This tool can used before the cctx is actually created on Zetacore to check the voting progress
Closes : https://github.com/zeta-chain/node/issues/3339

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

Based on the provided summary, here are the release notes:

- **Refactor**
  - Simplified CLI command for fetching inbound ballot using inbound hash
  - Restructured configuration management for zetatool
  - Removed deprecated filter deposit functionality

- **New Features**
  - Added support for fetching ballot identifiers across multiple blockchain networks (Ethereum, Bitcoin, Solana)
  - Enhanced inbound transaction handling in zetatool

- **Chores**
  - Updated error handling in command-line utilities
  - Improved modularity of inbound event processing


cmd/zetatool/inbound_ballot/bitcoin.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if event == nil {
+		return "", fmt.Errorf("no event built for btc sent to TSS")
+	}
+
+	return identifierFromBtcEvent(event, senderChainID, zetacoreChainID)
+}
+
+func identifierFromBtcEvent(event *zetaclientObserver.BTCInboundEvent,
```
kingpinXD, line 122: This follows similar logic to as processing Bitcoin Inbound Trackers , we can consider refactoring the tracker logic in the future so that it can be reused .

Similar comments for EVM

Solana logic could be used as it needed only minor changes on the zetaclient side

changelog.md
```diff
@@ -5,6 +5,7 @@
 ### Refactor
 
 * [3332](https://github.com/zeta-chain/node/pull/3332) - implement orchestrator V2. Move BTC observer-signer to V2
+* [3368](https://github.com/zeta-chain/node/pull/3368) - refactor cli command to fetch inbound ballot from inbound hash and move to zetatools
```
lumtis, line 8: I would consider it as a feature instead, we really make this command usable for the user

cmd/zetatool/inbound_ballot/evm.go
```diff
@@ -0,0 +1,363 @@
...
+func resolveRPC(chain chains.Chain, cfg config.Config) string {
+	return map[chains.Network]string{
+		chains.Network_eth:     cfg.EthereumRPC,
+		chains.Network_base:    cfg.BaseRPC,
+		chains.Network_polygon: cfg.PolygonRPC,
+		chains.Network_bsc:     cfg.BscRPC,
+	}[chain.Network]
+}
+
+func EvmInboundBallotIdentified(ctx context.Context,
```
lumtis, line 40:
```suggestion
func EVMInboundBallotIdentified(ctx context.Context,
```

lumtis, line 40: Name unclear, a function reflects an action, past participle is not adapted

lumtis, line 40: We can use private lowercase for all function names

cmd/zetatool/inbound_ballot/inbound.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/zeta-chain/node/cmd/zetatool/config"
+	"github.com/zeta-chain/node/pkg/chains"
+	zetacorerpc "github.com/zeta-chain/node/pkg/rpc"
+)
+
+func NewFetchInboundBallotCMD() *cobra.Command {
+	return &cobra.Command{
+		Use:   "get-ballot [inboundHash] [chainID]",
```
lumtis, line 17: Instead of using config, would be simpler to just directly pass the argument here. For EVM and most chain, you just need a RPC? Command would be simpler to use as
```
get-ballot hash chain_id rpc
```
rather than setting up a config we can also set a default public rpc used automatically for each chain

lumtis (changes requested):
Need to include a readme explaining how the command work

cmd/zetatool/inbound/bitcoin.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/zeta-chain/node/zetaclient/chains/bitcoin/common"
+	zetaclientObserver "github.com/zeta-chain/node/zetaclient/chains/bitcoin/observer"
+)
+
+func btcInboundBallotIdentifier(
+	cfg config.Config,
+	zetacoreClient rpc.Clients,
+	inboundHash string,
+	inboundChain chains.Chain,
+	zetaChainID int64) (string, error) {
```
kingpinXD, line 30: I think this comes up a lot for naming 
ChainID vs ChainId 
We have both in the codebase; I lean more towards ChainID when defining variable names, please comment on this thread if you disagree

cmd/zetatool/inbound_ballot/inbound.go
kingpinXD: We have a default value for now for each chain; the config is not mandatory.
I think we should just make sure the default values are working, having just two parameters is optimal . Specially if we want to use a HTTP server in the future instead of the CLI

cmd/zetatool/inbound_ballot/bitcoin.go
kingpinXD: The solution also bypasses som checks such as 
- Checking for restricted address 
- Checking for Donation message (IsEventProcessable check) 

I don't think those are required for a debugging tool, but they can be added if needed

We should consider though to refactor the client logic so derive out the common logic from this `GetBallotTool` , `ProcessInboundTracker` and in some cases `ProcessInbound`

cmd/zetatool/inbound/bitcoin.go
gartnera: `foreignChainID` would be my preference when referring to foreign chains.

But it's hard to find the right thing because we have our own internal IDs too. `zetaForeignChainID` in that case?

cmd/zetatool/inbound/bitcoin.go
```diff
@@ -0,0 +1,233 @@
...
+		User:         cfg.BtcUser,
+		Pass:         cfg.BtcPassword,
+		HTTPPostMode: true,
+		DisableTLS:   true,
+		Params:       params.Name,
+	}
+	rpcClient, err := rpcclient.New(connCfg, nil)
+	if err != nil {
+		return "", fmt.Errorf("error creating rpc client: %s", err)
+	}
```
gartnera, line 47: Please use the new internal bitcoin rpc client introduced in https://github.com/zeta-chain/node/pull/3349 if possible

cmd/zetatool/inbound/bitcoin.go
```diff
@@ -0,0 +1,233 @@
...
+
+func btcInboundBallotIdentifier(
+	cfg config.Config,
+	zetacoreClient rpc.Clients,
+	inboundHash string,
+	inboundChain chains.Chain,
+	zetaChainID int64) (string, error) {
+	params, err := chains.BitcoinNetParamsFromChainID(inboundChain.ChainId)
+	if err != nil {
+		return "", fmt.Errorf("unable to get bitcoin net params from chain id: %s", err)
```
gartnera, line 33:
```suggestion
		return "", fmt.Errorf("get bitcoin net params from chain id: %w", err)
```

I would prefer if we do not include "unable" or "error" in error messages. Also errors should be wrapped with `%w`.

Only commenting this seems to be a pattern throughout this file.

cmd/zetatool/inbound/bitcoin.go
```diff
@@ -0,0 +1,233 @@
...
+		return "", fmt.Errorf("error creating rpc client: %s", err)
+	}
+
+	err = rpcClient.Ping()
+	if err != nil {
+		return "", fmt.Errorf("error ping the bitcoin server: %s", err)
+	}
+	res, err := zetacoreClient.Observer.GetTssAddress(context.Background(), &types.QueryGetTssAddressRequest{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get tss address %s", err.Error())
```
gartnera, line 55: do not squash the error please

cmd/zetatool/inbound/evm.go
```diff
@@ -0,0 +1,380 @@
...
+	tssEthAddress := res.GetEth()
+
+	if tx.To() == nil {
+		return "", fmt.Errorf("invalid trasnaction,to field is empty %s", inboundHash)
+	}
+
+	confirmationMessage := ""
+	confirmed, err := zetaclientrpc.IsTxConfirmed(ctx, evmClient, inboundHash, chainParams.ConfirmationCount)
+	if err != nil {
+		return "", fmt.Errorf("unbale to confirm tx %s", err.Error())
```
gartnera, line 81:
```suggestion
		return "", fmt.Errorf("confirm tx: %w", err)
```

gartnera (dismissed):
Need to wrap errors properly. Also please check spelling throughout.

kingpinXD:
> Need to wrap errors properly. Also please check spelling throughout.

Fixed the formating , will open for review again after updating the btc client

cmd/zetatool/inbound/bitcoin.go
kingpinXD: In this case, this is the zetachainID, so it's fine.
In general, though, we are starting to use connected chains as the terminology

cmd/zetatool/inbound_ballot/inbound.go
lumtis: Ok, make sense, I think it would still better having flags instead of a config.
```
get-ballot hash chain_id --rpc "..."

get-ballot hash chain_id --rpc "..." --bitcoin-username "...."
```

cmd/zetatool/inbound/bitcoin.go
lumtis: Yes, Connected Chain for chain we connect to. Foreign Chain terminology needs to be removed from the codebase

docs/cli/zetatool/overview.md
```diff
@@ -0,0 +1,30 @@
...
+
+To install ZetaTool, clone the repository and build the project:
+
+```sh
+git clone https://github.com/zeta-chain/node.git
+cd node/cmd/zetatool
+go build -o zetatool
+```
+
+Alternatively you can also use the target `make install-zetatool`
```
lumtis, line 15: Let's just put this instruction, people just use the make to build. No need to specify about cloning the repo.
Let's just put
```
make install-zetatool
```

docs/cli/zetatool/overview.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Usage 
+
+### Fetching the Ballot Identifier
```
lumtis, line 19: Can we add an example of the command

docs/cli/zetatool/overview.md
```diff
@@ -0,0 +1,30 @@
...
+
+```shell
+zetatool get-ballot [inboundHash] [chainID] --config <filename.json>
+```
+
+- `inboundHash`: The inbound hash of the transaction for which the ballot identifier is to be fetched
+- `chainID`: The chain ID of the chain to which the transaction belongs
+- `config`: [Optional] The path to the configuration file. When not provided, the configuration in the file is user. A sample config is provided at `cmd/zetatool/config/sample_config.json`
```
lumtis, line 27: Let's specify that the config contains RPCs and default RPCs are used

docs/cli/zetatool/overview.md
```diff
@@ -0,0 +1,30 @@
...
+
+The command returns a ballot identifier for the given inbound hash.
```
lumtis, line 29: Let's put an example of the output

docs/cli/zetatool/overview.md
```diff
@@ -0,0 +1,30 @@
+# ZetaTool
+
+ZetaTool is a utility tool for the Zeta-Chain project. It provides a command to fetch the ballot/cctx identifier from the inbound hash
```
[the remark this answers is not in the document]
kingpinXD: I dit not want to add CLI , as its quite likely we will add a http server .

docs/cli/zetatool/overview.md
kingpinXD: This tool would be used by people who might not be running nodes or have the repo clones already; I think having the clone and build option would be good.

cmd/zetatool/inbound_ballot/inbound.go
kingpinXD: I am not sure if that is a good idea, 
For example for a getting a btc ballot we would need flags for 
- User 
- Password 
- Params 
- Host 
- ZetachainRPC 
- ZetachainID ( Good to have ) 

Which is a minimum of 5 flags. Imo, it's just better to use a rather use a config file , which can be created easily using the sample or a shell scripts.

The config also does not need all values to be present, just the rpc that you want to use

lumtis: Bitcoin is very specific because requires your own node with auth, most if not all other chains will just require a rpc so I wouldn't make the system more complex for just one edge case. We can also have config specifically for Bitcoin. But having the ability to provide the rpc directly in the command line for most chians would make the solution simpler

docs/cli/zetatool/overview.md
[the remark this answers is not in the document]
lumtis: What do you mean a http server? It's a CLI utility tool right now

docs/cli/zetatool/overview.md
lumtis: The tool will almost exclusively be used internally

lumtis:
@kingpinXD please make sure associated issue are linked to the PR

docs/cli/zetatool/overview.md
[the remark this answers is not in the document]
kingpinXD: I think the original request was to be able to get these values over http. But I can make this change int he docs for now

cmd/zetatool/inbound_ballot/inbound.go
kingpinXD: It would still need three inputs, `rpc` , `zetachainRpc` and `zetachainId` 

I think it's just cleaner to have the option of a config rather that having 3 flags , and a config (for bitcoin only ) 

Also, we should ideally just focus on having the correct rpcs listed in the default file directly, and not really provide any configuration at all , the config is primarily for unexpected situations. 

With that said , I can make the change to using the flags if its still needed

lumtis: Ok, let's try this way for now

lumtis (commented):
There will be a new consideration from https://github.com/zeta-chain/node/pull/3358

We add a inboundStatus that will be part of the digest for index. Most of the time will use default success value but for some cases we would want to check the index for a specific status.

I think the simplest solution will be to use success status by default and have a flag `--all-status` that shows the indexes for each possible status

kingpinXD:
> There will be a new consideration from #3358
> 
> We add a inboundStatus that will be part of the digest for index. Most of the time will use default success value but for some cases we would want to check the index for a specific status.
> 
> I think the simplest solution will be to use success status by default and have a flag `--all-status` that shows the indexes for each possible status

Sure that makes sense , We can however merge this pr In and do a refactor after https://github.com/zeta-chain/node/pull/3358 is merged

gartnera (approved):

lumtis (approved):
LGTM

PR #3374, opened 2025-01-17
fix: remove minimum rent exempt check for SPL token withdrawals

ws4charlie:
# Description

The minimum rent exempt `1_000_000 lamports` was intended to avoid SOL token withdraw failure on gateway program when a Solana receiver account doesn't have enough lamports.

This rent exempt restriction should not be applied on SPL token because withdrawing SPL token already needs a non trivial amount of SOL for potential ATA creation so we can skip the minimal withdraw amount checking. Its already anti spam.

Changes:

1. Skip rent exempt check in method `validateZRC20Withdrawal` for non-Gas token.
2. Reduce default SPL withdraw amount `1_000_000 -> 100_000` in SPL withdraw E2E test.

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Release Notes

- **Fixes**
	- Removed minimum rent exempt check for SPL token withdrawals
	- Updated validation logic for ZRC20 withdrawal events to support different coin types

- **Testing**
	- Enhanced test coverage for withdrawal events across different coin types
	- Updated test cases for BTC and SOL withdrawals

- **Improvements**
	- Added coin type parameter to withdrawal validation methods
	- Refined validation process for cross-chain transactions

### DIFF
--- a/x/crosschain/keeper/evm_hooks.go
+++ b/x/crosschain/keeper/evm_hooks.go
@@ -342,6 +342,8 @@ func (k Keeper) validateZRC20Withdrawal(
 			return errorsmod.Wrapf(types.ErrInvalidAddress, "unsupported address %s", string(to))
 		}
 	} else if chains.IsSolanaChain(chainID, additionalChains) {
+		// The rent exempt check is not needed for ZRC20 (SPL) tokens because withdrawing SPL token
+		// already needs a non-trivial amount of SOL for potential ATA creation so we can skip the check.
 		if coinType == coin.CoinType_Gas && value.Cmp(big.NewInt(constant.SolanaWalletRentExempt)) < 0 {
 			return errorsmod.Wrapf(
 				types.ErrInvalidWithdrawalAmount,
--- a/x/crosschain/keeper/evm_hooks.go
+++ b/x/crosschain/keeper/evm_hooks.go
@@ -133,7 +133,7 @@ func (k Keeper) ProcessZEVMInboundV1(
 
 		// If Validation fails, we will not process the event and return and error. This condition means that the event was correct, and emitted from a registered ZRC20 contract
 		// But the information entered by the user is incorrect. In this case we can return an error and roll back the transaction
-		if err := k.ValidateZRC20WithdrawEvent(ctx, eventZRC20Withdrawal, coin.ForeignChainId); err != nil {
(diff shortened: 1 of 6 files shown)


lumtis (commented):
Can we change the withdraw SLP test to use a lower value by default (so the if would be triggered for the check)

ws4charlie:
> Can we change the withdraw SLP test to use a lower value by default (so the if would be triggered for the check)

Yeah. I'm modifying that and running the E2E test in local.

x/crosschain/keeper/evm_hooks.go
```diff
@@ -331,7 +342,7 @@ func (k Keeper) validateZRC20Withdrawal(ctx sdk.Context, chainID int64, value *b
 			return errorsmod.Wrapf(types.ErrInvalidAddress, "unsupported address %s", string(to))
 		}
 	} else if chains.IsSolanaChain(chainID, additionalChains) {
-		if value.Cmp(big.NewInt(constant.SolanaWalletRentExempt)) < 0 {
+		if coinType == coin.CoinType_Gas && value.Cmp(big.NewInt(constant.SolanaWalletRentExempt)) < 0 {
```
lumtis, line 345: Can we add a comment explaining why we filter this

x/crosschain/keeper/evm_hooks.go
```diff
@@ -305,14 +305,25 @@ func (k Keeper) ProcessZetaSentEvent(
 
 // ValidateZRC20WithdrawEvent checks if the ZRC20Withdrawal event is valid
 // It verifies event information for BTC chains and returns an error if the event is invalid
-func (k Keeper) ValidateZRC20WithdrawEvent(ctx sdk.Context, event *zrc20.ZRC20Withdrawal, chainID int64) error {
+func (k Keeper) ValidateZRC20WithdrawEvent(
+	ctx sdk.Context,
+	event *zrc20.ZRC20Withdrawal,
+	chainID int64,
+	coinType coin.CoinType,
```
lumtis, line 312: Maybe we can pss the foreign coin object instead of these two?

ws4charlie: https://github.com/zeta-chain/node/pull/3374/commits/a409a1a011a7e13685bada4e062f69d63467f7d7

lumtis (approved):

PR #3377, opened 2025-01-17
feat: (BitcoinRBF-Step1) `zetacore` increast gas price in pending Bitcoin CCTXs and burns additional fees

ws4charlie:
# Description

**The problem:**

In some scenarios, a broadcasted Bitcoin outbound tx can get stuck in the mempool when its CCTX carries an `outdated low fee rate` in the struct. Bitcoin block time is `10 mins`, which is much longer than other chains. When a Bitcoin tx gets stuck, it can sit in the mempool for hours or days depending on traffic.

**The change: RBF (Replace-by-Fee)**

- Have `zetacore` update gas prices for all long-pending Bitcoin CCTXs (older than `GasPriceUpdateRetryIntervalBTC = 40 mins`).
- The `zetacore` increases all pending CCTXs' gas price to latest `medianGasPrice`. The latest `medianGasPrice` is the recent market price that will be used by `zetaclient` to bump the average fee rate of all stuck txs to the market level. Note: we also have a gas price multiplier of `2.0` to tolerate fee market fluctuation.
- When `zetacore` increasing CCTXs' gas price, it also set the same `medianGasPrice` value to field `GasPriorityFee` as a signal to `zetaclient` that the CCTX's gas price is increased and RBF is allowed.
- Gas price in CCTXs are increase only when there is enough gas fees in the gas stability pool. Otherwise, do nothing.
- When `zetaclient` detects stuck outbound in mempool, it will use the increased gas price to perform an `RBF`.

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Release Notes

- **New Features**
  - Enhanced gas price management for cross-chain transactions, allowing Zetacore to feed the latest gas prices to pending Bitcoin cross-chain transactions.
  - Improved support for gas price updates across different blockchain networks.

- **Improvements**
  - Refined gas price update mechanisms for EVM and Bitcoin chains.
  - Updated naming conventions for cross-chain transaction gas price functions.
  - Added more robust error handling for gas price updates.
  - Introduced a retry interval for gas price adjustments to improve efficiency.

- **Technical Updates**
  - Modified Bitcoin address handling in test utilities to ensure consistent string representation.
  - Introduced new methods for chain-specific gas price calculations.


x/crosschain/keeper/abci.go
```diff
@@ -52,46 +52,47 @@ func (k Keeper) IterateAndUpdateCctxGasPrice(
...
-			// iterate through all pending cctx
-			for _, pendingCctx := range res.CrossChainTx {
-				if pendingCctx != nil {
-					gasPriceIncrease, additionalFees, err := updateFunc(ctx, k, *pendingCctx, gasPriceIncreaseFlags)
-					if err != nil {
-						ctx.Logger().Info("GasStabilityPool: updating gas price for pending cctx failed",
-							"cctxIndex", pendingCctx.Index,
+		// iterate through all pending cctx
+		for _, pendingCctx := range res.CrossChainTx {
+			if pendingCctx != nil {
```
swift1337, line 73: How can `pendingCctx` == `nil`? If so, let's flip the logic and reduce nesting.

```go
if pendingCctx == nil {
 // ... e.g. continue
}

// the rest of the flow in the loop

Comment (lumtis) [reply x/crosschain/keeper/abci.go:73]: There have been extensive usage of pointers in type definition even when they were not necessary.\

In many case, there is no workflow path where the value can be nil, but a panic in the begin blockers would halt the chain so better add those verification in any case

File: x/crosschain/keeper/abci.go [hunk 2]
```diff
@@ -175,3 +200,32 @@ func CheckAndUpdateCctxGasPrice(
...
+	// zetacore simply update 'GasPriorityFee', and zetaclient will use it to schedule RBF tx
+	// there is no priority fee in Bitcoin, the 'GasPriorityFee' is repurposed to store latest fee rate in sat/vB
+	cctx.GetCurrentOutboundParam().GasPriorityFee = medianGasPrice.String()
+	k.SetCrossChainTx(ctx, cctx)
+
+	return math.ZeroUint(), math.ZeroUint(), nil
+}
+
+// IsCCTXGasPriceUpdateSupported checks if the given chain supports gas price update
+func IsCCTXGasPriceUpdateSupported(chainID int64, additionalChains []zetachains.Chain) bool {
```
lumtis, line 220: Can be private method

x/crosschain/keeper/abci.go
```diff
@@ -175,3 +200,32 @@ func CheckAndUpdateCctxGasPrice(
...
+	switch {
+	case zetachains.IsZetaChain(chainID, additionalChains):
+		return false
+	case zetachains.IsEVMChain(chainID, additionalChains):
+		return true
+	case zetachains.IsBitcoinChain(chainID, additionalChains):
+		return true
```
lumtis, line 227:
```suggestion
	case zetachains.IsEVMChain(chainID, additionalChains):
	case zetachains.IsBitcoinChain(chainID, additionalChains):
		return true
```

lumtis (commented):
Same problematic as in the old PR.
The increase in the fee should be paid somewhere. We should pay from the gas stability pool.

x/crosschain/keeper/abci_test.go
```diff
@@ -449,3 +605,125 @@ func TestCheckAndUpdateCctxGasPrice(t *testing.T) {
...
+			require.True(t, found)
+			newGasPrice, err := cctx.GetCurrentOutboundParam().GetGasPriorityFeeUInt64()
+			require.NoError(t, err)
+			require.Equal(t, tc.medianGasPrice, newGasPrice)
+			require.EqualValues(t, tc.blockTimestamp.Unix(), cctx.CctxStatus.LastUpdateTimestamp)
+		})
+	}
+}
+
+func Test_IsCCTXGasPriceUpdateSupported(t *testing.T) {
```
kingpinXD, line 674: Can we add testNet chains to this unit test as well.

x/crosschain/keeper/abci.go
kingpinXD, line 73: NIT : think it might cleaner to 
```suggestion
			if pendingCctx == nil {
			 continue
			}
```

testutil/sample/crypto.go
```diff
@@ -91,15 +92,23 @@ func EthAddressFromRand(r *rand.Rand) ethcommon.Address {
 }
 
 // BtcAddressP2WPKH returns a sample btc P2WPKH address
-func BtcAddressP2WPKH(t *testing.T, net *chaincfg.Params) string {
+func BtcAddressP2WPKH(t *testing.T, net *chaincfg.Params) *btcutil.AddressWitnessPubKeyHash {
```
kingpinXD, line 95: Is it possible to refactor this function to accept randomness as an input? 
Can be used to generate the private key  on line 96

x/crosschain/keeper/abci_test.go
```diff
@@ -254,104 +243,279 @@ func TestCheckAndUpdateCctxGasPrice(t *testing.T) {
...
+		withdrawFromGasStabilityPoolReturn     error
+		expectWithdrawFromGasStabilityPoolCall bool
+		expectedGasPriceIncrease               math.Uint
+		expectedAdditionalFees                 math.Uint
+		isError                                bool
+	}{
+		{
+			name: "can update gas price",
+			cctx: types.CrossChainTx{
+				Index: "a1",
```
kingpinXD, line 389: consider using sample.ZetaIndex(t), or a hash value from ` ethcrypto.Keccak256Hash`

testutil/sample/crypto.go
```diff
@@ -91,15 +92,23 @@ func EthAddressFromRand(r *rand.Rand) ethcommon.Address {
 }
 
 // BtcAddressP2WPKH returns a sample btc P2WPKH address
```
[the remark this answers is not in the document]
ws4charlie: updated description https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
ws4charlie: reduced nesting: https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
```diff
@@ -128,6 +129,30 @@ func CheckAndUpdateCctxGasPrice(
...
+
+// CheckAndUpdateCCTXGasPriceEVM updates the gas price for the given EVM chain CCTX
+func CheckAndUpdateCCTXGasPriceEVM(
+	ctx sdk.Context,
+	k Keeper,
+	medianGasPrice math.Uint,
+	medianPriorityFee math.Uint,
+	cctx types.CrossChainTx,
+	flags observertypes.GasPriceIncreaseFlags,
+) (math.Uint, math.Uint, error) {
```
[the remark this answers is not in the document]
ws4charlie: https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
```diff
@@ -175,3 +200,32 @@ func CheckAndUpdateCctxGasPrice(
...
 	return gasPriceIncrease, additionalFees, nil
 }
+
+// CheckAndUpdateCCTXGasPriceBTC updates the fee rate for the given Bitcoin chain CCTX
+func CheckAndUpdateCCTXGasPriceBTC(
+	ctx sdk.Context,
+	k Keeper,
+	medianGasPrice math.Uint,
+	cctx types.CrossChainTx,
+) (math.Uint, math.Uint, error) {
```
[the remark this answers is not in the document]
ws4charlie: https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci_test.go
```diff
@@ -254,104 +243,279 @@ func TestCheckAndUpdateCctxGasPrice(t *testing.T) {
...
+			medianPriorityFee:                  20,
+			withdrawFromGasStabilityPoolReturn: nil,
+			expectedGasPriceIncrease:           math.NewUint(0),
+			expectedAdditionalFees:             math.NewUint(0),
+		},
+	}
+	for _, tc := range tt {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			k, ctx := testkeeper.CrosschainKeeperAllMocks(t)
```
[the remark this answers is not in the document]
ws4charlie: updated https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
ws4charlie: Yeah. The drawback is that unit tests won't be able to access it and verify 16 different chain IDs in a straightforward way.

x/crosschain/keeper/abci.go
ws4charlie: https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci_test.go
ws4charlie: added testnets https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
ws4charlie: yeah https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
```diff
@@ -175,3 +200,32 @@ func CheckAndUpdateCctxGasPrice(
...
+	cctx.GetCurrentOutboundParam().GasPriorityFee = medianGasPrice.String()
+	k.SetCrossChainTx(ctx, cctx)
+
+	return math.ZeroUint(), math.ZeroUint(), nil
+}
+
+// IsCCTXGasPriceUpdateSupported checks if the given chain supports gas price update
+func IsCCTXGasPriceUpdateSupported(chainID int64, additionalChains []zetachains.Chain) bool {
+	switch {
+	case zetachains.IsZetaChain(chainID, additionalChains):
```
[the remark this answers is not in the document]
ws4charlie: simplified https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci_test.go
ws4charlie: updated https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

x/crosschain/keeper/abci.go
```diff
@@ -52,46 +52,47 @@ func (k Keeper) IterateAndUpdateCctxGasPrice(
...
+			if pendingCctx != nil {
+				gasPriceIncrease, additionalFees, err := updateFunc(ctx, k, *pendingCctx, gasPriceIncreaseFlags)
+				if err != nil {
+					ctx.Logger().Info("GasStabilityPool: updating gas price for pending cctx failed",
+						"cctxIndex", pendingCctx.Index,
+						"err", err.Error(),
+					)
+					continue IterateChains
+				}
+				if !gasPriceIncrease.IsNil() && !gasPriceIncrease.IsZero() {
```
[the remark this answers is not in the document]
ws4charlie: same commit https://github.com/zeta-chain/node/pull/3377/commits/279335bf423ac338cf96c984e6bb27df7e68eb61

ws4charlie:
> Same problematic as in the old PR. The increase in the fee should be paid somewhere. We should pay from the gas stability pool.

**For the draft PR:**

In the draft PR, it was the TSS self pays the additional bump fee. The difficulties I have seen for the `GasStabilityPool` to pay the bump fee in block begin are:

1. Say we have pending CCTXS `cctx1, cctx2, cctx3, cctx4, cctx5, ..., cctx10` and have only `Tx1, Tx2, Tx3` broadcasted stuck in mempool. Only `Tx3` is the `LastStuckOutbound` needing a fee bumping. The `zetacore` does not know which one is the last stuck tx. What it can do is to bump fees for all the pending CCTXs, and it is far from correct (even less correct than EVM chains)

2. Determining `which tx becomes the LastStuckOutbound` is done by actively checking `mempool` and see how long the tx has been pending in Bitcoin `mempool`, it involves RPC calls. Determing `whether or not a stuck tx is no longer stuck` is a similar process. The `zetacore` is not aware of outbound status.


**For future gas fee saving:**

Now, ignore the difference in tx replacement mechanism. I believe a long-term goal for Zetachain will be lowering gas fee for cross-chain service. Here is some understanding around the current `GasStabilityPool` model and some thoughts around gas fee saving.

The `GasStabilityPool` works in a `always-increase-gas-price` model in `zetacore`.

Taking `Ethereum` as example and assuming a scenario when we've got `Tx1, Tx2, Tx3, Tx4,...Tx10` stuck due to outdated gasPrice (low price), later on we get 10 more CCTXs and broadcasted another `Tx10, Tx11, Tx12, Tx13,...Tx20` with normal gasPrice. The latter 10 also get stuck because the initial txs can't go through. Now, the `GasStabilityPool` bumps all 20 pending txs, hoping `zetaclient` to refresh gas prices; after `zetaclient` completing re-broadcastings for `Tx1, Tx2, Tx3, Tx4,...Tx10` with higher gas price (haven't do it for 10~20 yet), the Ethereum network just take all 20 pending txs because the latter 10 already carry market gas prices. So, the prior burned fee (in `zetacore`) for `Tx10, Tx11, Tx12, Tx13,...Tx20` needs to be compensated through `FundGasStabilityPool`, which is perfect.

It is reasonable to try `increasing` gas price in the pending EVM-chain CCTXs if they don't finalize in a period of time (`10 mins` for now). In most cases, `the gas price is probably too low` is a good guess. For scenarios like TSS signers are not working well or RPC not working well. From the fee saving perspective, the `GasStabilityPool` dare not try `lowering` the gas price in pending CCTXs because the external chain tx status is invisible to it.

If lowering gas fee and multiplier are the goals on our roadmap, a `always-use-market-gas-price` model worths consideration. Like `Metamask` wallet is doing, it always broadcasts tx with market rate by default and rarely gets stuck. To make it happen, handing over the fee bumping logic from `zetacore` to an actor (most likely `zetaclient`) who knows the fee market and effectively sends out the txs is a good direction to move, because `zetacore` lacks confidence in coordinating the fee bumping in a timely manner.


**For future gas fee accounting:**

The moment `GasStabilityPool` updates the `GasPrice` in original CCTX struct, the CCTX losts track of the original fees paid on withdraw because the parameter changed in the formula: `fees = gasPrice * gasLimit + protocolFlatFee`. Also, the new `gasPrice` doesn't really reflect the actual price (it will be higher after bumping) used for external transaction.

To accurately track outbound txs fees, we could:

- Add one additional field `Fees` in CCTX outbound, and have `zetaclient` to provide the fee amount in `MsgVoteOutbound`.
- `GasStabilityPool` stops bumping fees and updating `GasPrice` in the `zetacore`, leaving the params intact in the formula: `fees = gasPrice * gasLimit + protocolFlatFee`

lumtis:
>For the draft PR:
In the draft PR, it was the TSS self pays the additional bump fee. The difficulties I have seen for the GasStabilityPool to pay the bump fee in block begin are:
Say we have pending CCTXS cctx1, cctx2, cctx3, cctx4, cctx5, ..., cctx10 and have only Tx1, Tx2, Tx3 broadcasted stuck in mempool. Only Tx3 is the LastStuckOutbound needing a fee bumping. The zetacore does not know which one is the last stuck tx. What it can do is to bump fees for all the pending CCTXs, and it is far from correct (even less correct than EVM chains)
Determining which tx becomes the LastStuckOutbound is done by actively checking mempool and see how long the tx has been pending in Bitcoin mempool, it involves RPC calls. Determing whether or not a stuck tx is no longer stuck is a similar process. The zetacore is not aware of outbound status.

I don't understand, in the PR we already have the logic iterating the pending CCTX and increasing the fee in the CCTX, what I'm asking is to pay this increase from the gas stability pool, I don't think this adds up to the problematic you mention based on the current implementation. TSS should any infer the cost for a withdraw.

On a second hand I don't think it's problematic increasing all cctx with outdated fee is problematic.

ws4charlie:
> > For the draft PR:
> > In the draft PR, it was the TSS self pays the additional bump fee. The difficulties I have seen for the GasStabilityPool to pay the bump fee in block begin are:
> > Say we have pending CCTXS cctx1, cctx2, cctx3, cctx4, cctx5, ..., cctx10 and have only Tx1, Tx2, Tx3 broadcasted stuck in mempool. Only Tx3 is the LastStuckOutbound needing a fee bumping. The zetacore does not know which one is the last stuck tx. What it can do is to bump fees for all the pending CCTXs, and it is far from correct (even less correct than EVM chains)
> > Determining which tx becomes the LastStuckOutbound is done by actively checking mempool and see how long the tx has been pending in Bitcoin mempool, it involves RPC calls. Determing whether or not a stuck tx is no longer stuck is a similar process. The zetacore is not aware of outbound status.
> 
> I don't understand, in the PR we already have the logic iterating the pending CCTX and increasing the fee in the CCTX, what I'm asking is to pay this increase from the gas stability pool, I don't think this adds up to the problematic you mention based on the current implementation. TSS should never infer the cost for a withdraw.
> 
> On a second hand I don't think it's problematic increasing all cctx with outdated fee, not paying the fees is problematic.

- The TSS actually doesn't infer any cost for the withdraw in RBF tx, it uses the updated fee rate (the `GasPriorityFee` field) to bump the RBF tx.
- Say, we have pending `CCTX1, CCTX2, CCTX3,...CCTX10`, and TSS has broadcasted only `Tx1, Tx2, Tx3` and stops signing (the nonce UTXO is not present due to stuck, it can't not sign more tx). In this case, only `TX3` needs fee bump, but `zetacore` doesn't know only `TX3` needs the increase. What `zetacore` can do is to force bumping all CCTXs.
- There can be late-registered CCTXs (already carry market fee and not having outbounds signed yet), they don't need extra higher fees. If `zetacore` forces burning extra fees for all pending CCTXs, the TSS signers should still not be guided by an unreasonable higher rate, for long-term fee saving goal.
<br />

A question:

What's the purpose of having `GasStabilityPool` burning the fee? Is it intended to track something like how much extra fees burned?

Again, the `GasStabilityPool` works in a `always-increase-gas-price` model even if we already have `EVMOutboundGasPriceMultiplier = 1.2`. The system should be able to reduce fee rate (if CCTX gasPrice is too high and outdated) as well.

ws4charlie:
Here are some diagrams to visualize the current fee bumping vs. RBF fee bumping.

Here is a typical case when outbounds get stuck:

- We have `CCTX1, CCTX2, CCTX3` pending in `zetacore` for a while, carrying low `rate = 5`
- The corresponding outbounds `tx1, tx2, tx3` were broadcasted and stuck, `latest rate = 10`
- Later we have newly created `CCTX4, CCTX5, CCTX6`, carrying `rate=8`, `rate=12` and `rate=10` respectively.

Note: In Bitcoin, TSS signers will stop signing future txs `tx3, tx4, tx5` due to a stuck tx. The reason is that the RPC `GetTransaction` returns error for stuck txs, so `zetaclient` can't verify the stuck txs  and stops processing more.

image


### 1. When current fee bumping mechanism bumps fee, it bumps all CCTXs.

In below diagram, `zetacore` increase gas rate in all CCTXs and burns fee for each of them, even if `tx3, tx4, tx5` haven't ever been broadcasted before. In this case, TSS overpays `(13/17/15 - originalGasPrice) * gasUsed` for the corresponding CCTX.

image


### 2. When RBF fee bumping mechanism bumps fee, it bumps `tx3` only.

In RBF fee bumping, `zetaclient` only increase the rate in `tx3`, making an average of `tx1, tx2, tx3` as `10`. The latest `rate = 10` is fed by `zetacore` in `GasPriorityFee` field as the hint. For remaining `tx4, tx5, tx6`, the same latest `rate = 10` will be used, ignoring of the `gasPrice` field.

Why bump `tx3` only?
Bumping `tx1` incurs more complicity and more fee cost.
1. Replacing `tx1` immediately voids all its child transactions `tx2, tx3`, which will decrease the pending nonce and brings extra logical complexity in `zetaclient`.
2. It requires additional fee `> feeTx2 + feeTx3 + minRelayFee`, because we're deleting `tx2, tx3`, meaning immediate lost of the fees already paid in `tx2, tx3`.

For `tx4`, it will be broadcasted at `rate = 10`. The `rate=8` should not be used as it may cause another stuck tx.
For `tx5`, it will be broadcasted at `rate=10`. Using an outdated `rate=12` is an over payment.
For `tx6`, it will be broadcasted at `rate=10`. It carries market fee rate.

image


### 3. What would happen if the reason for pending CCTXs (`1~6`) is that TSS is not working (network issue, bug, etc.)?

In this case, the `zetacore` will bumps fee for all six CCTXs until it hits the max rate of `latest * 300% = 30`. So when TSS issue gets resolved, outbound `tx1, tx2, tx3, tx4, tx5, tx6` will be broadcasted at `rate=30`, even when latest rate is `10`. In this case, TSS overpays `(30 - originalGasPrice) * gasUsed` for each CCTX.

TSS not working is one of the edge cases, and it can be any other issues that can halt/delay outbound.

image


### 4. A question about `GasStabilityPool` is: what exactly are we trying to control through it?
It seems we positioned `GasStabilityPool` as a mandatory middle man. When CCTXs are pending for long time, it burns gas and hopes `zetaclient` to resolve the outbound according to its guidance.

When `GasStabilityPool` runs out of funds, we'll need to manually funds it in order to unblock stuck outbound. So, in what scenarios (when ran out of funds) we should NOT fund `GasStabilityPool`? Like, if we manually look at the Ethereum gas price an it is surprisingly high or something else?

kingpinXD:
> 
> ### 4. A question about `GasStabilityPool` is: what exactly are we trying to control through it?
> It seems we positioned `GasStabilityPool` as a mandatory middle man. When CCTXs are pending for long time, it burns gas and hopes `zetaclient` to resolve the outbound according to its guidance. We say `hopes` because what's happening in external chain is invisible to `zetacore`.
> 
> When `GasStabilityPool` runs out of funds, we'll need to manually funds it in order to unblock stuck outbound. So, in what scenarios (when ran out of funds) we should NOT fund `GasStabilityPool`? Like, if we manually look at the Ethereum gas price an it is surprisingly high or something else?

IMO, the following is the use for a gas stability pool at a high level.
 It should be used to account for any profit the TSS makes over time; funds get added to it whenever outbound uses fewer funds than what the user paid for. It would maintain an accounting for this number so that we don't automatically start paying out more than what the TSS owns ( we can manually override by adding to the pool )

ws4charlie:
@kingpinXD 

What I see is that the current implementation funds `(EffectiveGasLimit - EffectiveGasUsed) * EffectiveGasPrice` back to the `GasStabilityPool`. If we want to track a number of surplus (or deficit), we need

1. Add `(EffectiveGasUsed * EffectiveGasPrice) - (OriginalGasLimitByUser * OriginalGasPriceByUser)`, it can be either a positive or negative number.
2. The  balance of`GasStabilityPool` can be either positive (surplus) or negative (deficit).
3. The current accounting above is incorrect.

Accounting is still a separate thing, it just visualizes the system and provides some hints. If the balance becomes negative, then we could increase the charge (by some multiplier) from users; if the balance grows, we could consider lower the charge (by same multiplier).

If the balance becomes negative, it doesn't mean we should stop bumping fees, as long as the current gas price isn't suspiciously high. So what we should control are 3 parameters:

1. `gasPriceMultiplierChargeUser = 1.2`, this collects surplus from users on `withdraw`.
2. `gasPriceMultiplierSendOutbound = 1.1`, this helps finality (> market fee) and also makes profit (< 1.2)
3. `gasPriceCap = some configurable number`, this protect against accidental high gas price.

```
if gasLimit > 0 && gasUsed > 0 && !gasPrice.IsZero() {
	if gasLimit > gasUsed {
		remainingGas := gasLimit - gasUsed
		remainingFees := math.NewUint(remainingGas).Mul(gasPrice).BigInt()

		// We fund the stability pool with a portion of the remaining fees.
		remainingFees = percentOf(remainingFees, RemainingFeesToStabilityPoolPercent)

		// Fund the gas stability pool.
		if err := k.fungibleKeeper.FundGasStabilityPool(ctx, chainID, remainingFees); err != nil {
			return err
                 }
                 ...
	}
}
```

lumtis:
Again the gas stability pool mechanism is up to be optimized. The overall vision of the mechanism is to have an automated way, that doesn't require supervision, to unblock stuck cctx. On EVM if a nonce is long pending, we make the assumption that it is stuck because of gas price as the blockchain doesn't have access to the external world, we also iterate next cctxs for time gain but this is a part to be optimized as well, tracked by an issue.

I think the fact we iterate multiple CCTX even the non-stuck ones for Bitcoin for now is a detail, we can easily subsidized the cost.

My original question where I don't think I had an answer yet is who pay for the fee bump?

> The TSS actually doesn't infer any cost for the withdraw in RBF tx, it uses the updated fee rate (the GasPriorityFee field) to bump the RBF tx.

This doesn't answer the question, where does the fund come from to update the fee rate? If the fee rate is increase, the cost is higher for the outbound, who pay for this difference of cost?

ws4charlie:
> This doesn't answer the question, where does the fund come from to update the fee rate? If the fee rate is increase, the cost is higher for the outbound, who pay for this difference of cost?

In the RBF implementation, the users (who have ever withdrawn funds) will pays for it and Zetachain is going to make profit under the current multiplier `2.0`.

Zetachain charge withdraw fee at multiplier `2.0` and pays Bitcoin network gas fee at multiplier `1.5` here: https://github.com/zeta-chain/node/blob/feat-bitcoin-RBF-zetaclient-implementation/zetaclient/chains/bitcoin/signer/outbound_data.go#L83

lumtis:
>In the RBF implementation, the users (who have ever withdrawn funds) will pays for it and Zetachain is going to make profit under the current multiplier 2.0. We never make profit since mainnet launch even if we keep charging EVM chain gas fee at multiplier 1.2 and Bitcoin chain at 2.0.
Zetachain charge withdraw fee at multiplier 2.0 and pays Bitcoin network gas fee at multiplier 1.5 here: https://github.com/zeta-chain/node/blob/feat-bitcoin-RBF-zetaclient-implementation/zetaclient/chains/bitcoin/signer/outbound_data.go#L83

Ok, so the user pay 33% surplus on the initial outbound fee. What if the increase for RBF is more than this 33% surplus? Is it a impossible scenario?

lumtis:
And why did we decided to have this 33% surplus in the first place instead of the user paying only the actual fee?

ws4charlie:
> Ok, so the user pay 33% surplus on the initial outbound fee. What if the increase for RBF is more than this 33% surplus? Is it a impossible scenario?
User pays `2.0 * gasPrice`, we `use 1.5 * gasPrice`, so we make 25% profit from user as an insurance. These two parameters are configurable so we can either increase profit, or lower it to allow users enjoying lower fee. The benefit isn't focusing on the profit one single CCTX, it it's a crowdfunding.

> And why did we decided to have this 33% surplus in the first place instead of the user paying only the actual fee?
Paying actual fee means we charge on multiplier `1.0` and TSS send tx on multiplier `1.0`, it is the ideal case (revenue-expenditure balances) in a statistical sense.

lumtis:
> Where is the percentage 33% coming from?, maybe I misunderstood.

If user pays 2.0x, and TSS use 1.5x, this is a 33% surplus

>User pays 2.0 * gasPrice * gasLimit, we use 1.5 * gasPrice * gasLimit, so we make 25% profit from user as an insurance. These two parameters are configurable so we can either increase profit, or lower it to allow users enjoying lower fee.

You can't make assumption that this covers the funds to increase the CCTX, if a CCTX fee is increase, this increase needs to be covered as ZRC20 at the protocol level, there is no way around.

ws4charlie:
Here is the new commit to burn additional fees from gas stability pool: https://github.com/zeta-chain/node/pull/3377/commits/a509d8835e03632006160b394e483dd6407768d7, @lumtis  could you pls check if this looks better?

proto/zetachain/zetacore/observer/crosschain_flags.proto
```diff
@@ -19,6 +19,10 @@ message GasPriceIncreaseFlags {
   // Maximum number of pending crosschain transactions to check for gas price
   // increase
   uint32 maxPendingCctxs = 5;
+
+  // Retry interval for Bitcoin chain pending cctx gas price increase
```
lumtis, line 23: Why using a different vallue compared to other chain? block time?

x/crosschain/keeper/abci.go
```diff
@@ -175,3 +204,81 @@ func CheckAndUpdateCctxGasPrice(
...
+// IsCCTXGasPriceUpdateSupported checks if the given chain supports gas price update
+func IsCCTXGasPriceUpdateSupported(chainID int64, additionalChains []chains.Chain) bool {
+	switch {
+	case chains.IsZetaChain(chainID, additionalChains):
+		return false
+	case chains.IsEVMChain(chainID, additionalChains),
+		chains.IsBitcoinChain(chainID, additionalChains):
+		return true
+	default:
+		return false
```
lumtis, line 282:
```suggestion
func IsCCTXGasPriceUpdateSupported(chainID int64, additionalChains []chains.Chain) bool {
	switch {
	case chains.IsEVMChain(chainID, additionalChains),
		chains.IsBitcoinChain(chainID, additionalChains):
		return true
	default:
		return false
```

x/crosschain/keeper/abci.go
```diff
@@ -175,3 +204,81 @@ func CheckAndUpdateCctxGasPrice(
...
+	if math.NewUint(currentGasPrice).GTE(newGasPrice) {
+		return math.ZeroUint(), math.ZeroUint(), nil
+	}
+
+	// compute gas price increase
+	gasPriceIncrease = newGasPrice.SubUint64(currentGasPrice)
+
+	// withdraw additional fees from the gas stability pool
+	chainID := cctx.GetCurrentOutboundParam().ReceiverChainId
+	gasLimit := math.NewUint(cctx.GetCurrentOutboundParam().CallOptions.GasLimit)
```
lumtis, line 248: We don't use this paramter with Bitcoin, don't we?

x/crosschain/keeper/abci.go
```diff
@@ -102,7 +107,7 @@ IterateChains:
 
 // CheckAndUpdateCctxGasPrice checks if the retry interval is reached and updates the gas price if so
```
kingpinXD, line 108:
```suggestion
// CheckAndUpdateCCTXGasPrice checks if the retry interval is reached and updates the gas price if so
```

proto/zetachain/zetacore/observer/crosschain_flags.proto
ws4charlie: It will take 30 minutes for a BTC outbound to finalize.

x/crosschain/keeper/abci.go
```diff
@@ -19,20 +19,20 @@ const (
...
 	k Keeper,
 	cctx types.CrossChainTx,
 	flags observertypes.GasPriceIncreaseFlags,
 ) (math.Uint, math.Uint, error)
 
-// IterateAndUpdateCctxGasPrice iterates through all cctx and updates the gas price if pending for too long
+// IterateAndUpdateCCTXGasPrice iterates through all cctx and updates the gas price if pending for too long
 // The function returns the number of cctxs updated and the gas price increase flags used
-func (k Keeper) IterateAndUpdateCctxGasPrice(
+func (k Keeper) IterateAndUpdateCCTXGasPrice(
```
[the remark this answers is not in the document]
ws4charlie: Creating sub methods is good. For example, having a `hasGasPriceUpdateEpochArrived` to validate and `updateCCTXGasPriceForChain` to list pending CCTXs and update.

x/crosschain/keeper/abci.go
ws4charlie: Yeah. `Zetaclient` is not using it.

It is only used to charge withdrawer fee. It is a static number `gasLimit == 177`, meaning each withdrawal costs `177vB` in Bitcoin transaction size, and we charge withdraw fee as `177vB * gasPrice`. When `gasPrice` increases, it charges extra `177vB * extraGasPrice`

x/crosschain/keeper/abci.go
ws4charlie: The `chains.IsEVMChain` also returns `true` on ZetaChain as ZetaChain has `chain.Vm == Vm_evm`.

x/crosschain/keeper/abci.go
ws4charlie: updated https://github.com/zeta-chain/node/pull/3377/commits/2ccee4c00e6e2f72b5353b87dcdb59647e1cb99b

lumtis:
Any idea why the simulation tests fail?

x/crosschain/types/message_vote_gas_price.go
```diff
@@ -53,5 +54,8 @@ func (msg *MsgVoteGasPrice) ValidateBasic() error {
 	if msg.ChainId < 0 {
 		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidChainID, "chain id (%d)", msg.ChainId)
 	}
+	if chains.IsBitcoinChain(msg.ChainId, []chains.Chain{}) && msg.PriorityFee != 0 {
```
lumtis, line 57: Can we add comment why this check specifically for Bitcoin

x/observer/client/cli/tx_update_gas_price_increase_flags.go
```diff
@@ -43,6 +43,10 @@ func CmdUpdateGasPriceIncreaseFlags() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			retryIntervalBTC, err := strconv.ParseInt(args[5], 10, 64)
+			if err != nil {
+				return err
+			}
```
lumtis, line 49:
```suggestion
			
			retryIntervalBTC, err := strconv.ParseInt(args[5], 10, 64)
			if err != nil {
				return err
			}
			
```

ws4charlie:
> Any idea why the simulation tests fail?

This was caused by gas price validation change. Updated.

x/crosschain/types/message_vote_gas_price.go
ws4charlie: description added: https://github.com/zeta-chain/node/pull/3377/commits/98a7472ac6d9abc750b72d89de04c70e93496a32

lumtis (approved):

PR #3378, opened 2025-01-18
fix: improve solana tx broadcasting in e2e tests

skosito:
# Description

Improve broadcasting of solana txs in e2e test. This was tested on mainnet and contains couple of fixes, I tried 5 consecutive spl deposits and all worked without tx dropping.

For context, tx was dropped on mainnet and the biggest issue is that blockhash expired sometimes between tx signing and broadcasting, therefore changed to use confirmed commitment level, which is fine, especially for tests, since this should reduce that interval.

Other fixes:
- add cp limit and price so tx is better packed
- skip preflight for tests, as simulation is not needed for tests, and it speeds up broadcast
- add preflight commitment with max retry - this is the tip i found, no matter that preflight is skipped, it is still affecting automatic retry
- manual re-broadcast (this might be overkill as showed with latest testing, but still keeping it just in case, also would be ok to remove it for now imo, but leaving it in PR just in case)

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

Based on the comprehensive summary, here are the release notes:

- **Improvements**
  - Enhanced Solana blockchain interaction by updating transaction commitment levels
  - Improved transaction handling and compute budget management
  - Added more robust error logging for transaction broadcasts

- **Technical Updates**
  - Updated account information retrieval methods
  - Modified balance checking mechanisms across multiple test scenarios
  - Introduced more granular transaction confirmation strategies

- **Performance**
  - Added compute unit limits and price settings for transactions
  - Optimized transaction broadcasting logic

These changes aim to improve the reliability and precision of blockchain interactions in our testing and deployment processes.

### DIFF
--- a/e2e/runner/solana.go
+++ b/e2e/runner/solana.go
@@ -8,6 +8,7 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/gagliardetto/solana-go"
 	associatedtokenaccount "github.com/gagliardetto/solana-go/programs/associated-token-account"
+	computebudget "github.com/gagliardetto/solana-go/programs/compute-budget"
 	"github.com/gagliardetto/solana-go/programs/system"
 	"github.com/gagliardetto/solana-go/programs/token"
 	"github.com/gagliardetto/solana-go/rpc"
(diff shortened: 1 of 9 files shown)


lumtis (approved):

e2e/runner/solana.go
```diff
@@ -313,39 +328,82 @@ func (r *E2ERunner) DeploySPL(privateKey *solana.PrivateKey, whitelist bool) *so
...
+
+// BroadcastTxSync broadcasts a transaction and waits for it to be finalized
+func (r *E2ERunner) BroadcastTxSync(tx *solana.Transaction) (solana.Signature, *rpc.GetTransactionResult) {
+	r.Logger.Info("Broadcast start")
+	start := time.Now()
+	timeout := 2 * time.Minute // Expires after 2 mins
+	sig, out, isConfirmed := r.BroadcastTxSyncOnce(tx)
 	for {
-		require.False(r, time.Since(start) > timeout, "waiting solana tx timeout")
+		require.False(r, time.Since(start) > timeout, "solana tx timeout")
```
kingpinXD, line 397: Can we use the retry package for this instead?

brewmaster012 (approved):

PR #3353, opened 2025-01-15
feat: add liquidity cap parameter to ZRC20 creation

lumtis:
# Description

- Add liquidityCap paramter for whitelistERC20 and deployZRC20 messages

## Summary by CodeRabbit

Based on the comprehensive summary, here are the release notes:

## Release Notes

- **New Features**
  - Added liquidity cap parameter for ZRC20 token creation
  - Enhanced token deployment with more precise liquidity management
  - Expanded CLI commands to support liquidity cap specification

- **Improvements**
  - Updated transaction message structures to include liquidity cap
  - Improved validation for token deployment and whitelisting processes
  - Added support for more granular control over token liquidity

- **Technical Updates**
  - Integrated `cosmossdk.io/math` library for precise numeric handling
  - Updated multiple modules to support new liquidity cap functionality

PR #3387, opened 2025-01-21
feat: allow nil liquidity cap

gartnera:
Allow nil/null liquidity cap on https://github.com/zeta-chain/node/pull/3353 to allow upgrade tests to pass


x/fungible/types/message_deploy_fungible_coin_zrc20.go
```diff
@@ -69,9 +69,6 @@ func (msg *MsgDeployFungibleCoinZRC20) ValidateBasic() error {
 	if msg.Decimals > 77 {
 		return cosmoserrors.Wrapf(sdkerrors.ErrInvalidRequest, "decimals must be less than 78")
 	}
-	if msg.LiquidityCap.IsNil() {
```
lumtis, line 72: Not sure we should remove this check? Seems still relevant to me
`liquidityCap` itself can be nil, then default would be set, but we should still prevent using the value inside `liquidityCap` to be nil

x/fungible/keeper/evm_test.go
```diff
@@ -408,6 +410,32 @@ func TestKeeper_DeployZRC20Contract(t *testing.T) {
 		require.NotNil(t, newBalance)
 		require.Equal(t, amount.Int64(), newBalance.Int64())
 	})
+
+	t.Run("can deploy the zrc20 contract with default liquidity cap", func(t *testing.T) {
```
swift1337, line 414: What would happen if we have the default cap (e.g. $1000 USDC) and someone tries to deploy $2000?

swift1337 (approved):
LGTM

x/fungible/types/message_deploy_fungible_coin_zrc20.go
gartnera: I'm think `ValidateBasic()` gets called in cosmos side code so we wouldn't have the ability to set the default value before this gets called.

The default value logic lives in `x/fungible/keeper/evm.go` right now.

x/fungible/keeper/evm_test.go
gartnera: Your deposit gets reverted if it will exceed the liquidity cap.

x/fungible/types/message_deploy_fungible_coin_zrc20.go
gartnera: Ah ok I understand what you're saying. Restored the inner nil check and test: https://github.com/zeta-chain/node/pull/3387/commits/b3e52cf358b2328fead04684715a7f1e19a78b15

lumtis (approved):

PR #3406, opened 2025-01-23
refactor(e2e): use context cancellation rather than exit/panic

gartnera:
Use context cancellation rather than a hard exit in the e2e tests to improve debugging of e2e hangs. We often see the e2e tests exit because of the block production timeout. That's not great since we can't tell which test is hanging. I think it's often just one that end up taking a long time.

- cancel the context in both the block production monitor and the top level timeout goroutine
- handle context cancellation in `RunE2ETest`
- catch failure (panic) in individual test so that we can log duration
- route os signals down to context cancellation. This ensure that the current hung test is logged if we hit the github actions level timeout.
- make test messages a bit cleaner
- use cause style cancellation so we can indicate why the context is being cancelled

Example where I set test timeout to 5 minutes:

```
btc_deposit  | ✅ completed - bitcoin_std_memo_inscribed_deposit_and_call (duration 9.57673786s)
btc_deposit  | ⏳ running   - crosschain_swap
setup        | ❌ bitcoin tests failed: context cancelled in crosschain_swap after 22.417662125s
setup        | ❌ e2e tests failed after 5m0.004704899s
e2e failed
```


cmd/zetae2e/local/local.go
```diff
@@ -147,7 +144,19 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 	}
 
 	// initialize context
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, legacyCancel := context.WithTimeoutCause(context.Background(), TestTimeout, ErrTopLevelTimeout)
+	defer legacyCancel()
+	ctx, cancel := context.WithCancelCause(ctx)
+
```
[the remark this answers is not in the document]
gartnera: The timeout cause and cancellation cause are separate things. WithTimeoutCause's cancel does not take a reason.

cmd/zetae2e/setup_bitcoin.go
```diff
@@ -46,7 +46,7 @@ func runSetupBitcoin(_ *cobra.Command, args []string) error {
 		logger,
 	)
 	if err != nil {
-		cancel()
+		cancel(err)
```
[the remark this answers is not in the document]
gartnera: it is type context.CancelCauseFunc

cmd/zetae2e/local/local.go
```diff
@@ -187,7 +196,7 @@
 	// monitor block production to ensure we fail fast if there are consensus failures
 	// this is not run in an errgroup since only returning an error will not exit immediately
 	// this needs to be early to quickly detect consensus failure during genesis
-	go monitorBlockProductionExit(ctx, conf)
+	go monitorBlockProductionExit(ctx, cancel, conf)
```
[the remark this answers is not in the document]
gartnera: The wrapper function already handles this. But I probably should rename it `monitorBlockProductionCancel`

e2e/runner/run.go
```diff
@@ -17,21 +21,41 @@ func (r *E2ERunner) RunE2ETests(e2eTests []E2ETest) (err error) {
...
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				errChan <- fmt.Errorf("panic: %v", r)
+			}
+			close(errChan)
+		}()
+		e2eTest.E2ETest(r, args)
+		errChan <- nil
+	}()
```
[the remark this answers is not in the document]
gartnera: stack trace is pretty pointless in this context imo

cmd/zetae2e/balances.go
```diff
@@ -46,7 +46,7 @@ func runBalances(cmd *cobra.Command, args []string) error {
 	logger := runner.NewLogger(false, color.FgHiCyan, "")
 
 	// initialize context
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancelCause(context.Background())
```
swift1337, line 49: nice, I wasn't aware of this function

cmd/zetae2e/run.go
```diff
@@ -99,8 +99,8 @@ func runE2ETest(cmd *cobra.Command, args []string) error {
 	}
 
 	// initialize context
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	ctx, cancel := context.WithCancelCause(context.Background())
```
swift1337, line 102: do we need to have a `cancel` here at all?

e2e/runner/run.go
```diff
@@ -17,21 +21,41 @@ func (r *E2ERunner) RunE2ETests(e2eTests []E2ETest) (err error) {
...
+		return fmt.Errorf("context cancelled in %s after %s", e2eTest.Name, time.Since(startTime))
+	}
 
 	// check zrc20 balance vs. supply
 	if checkAccounting {
 		r.CheckZRC20BalanceAndSupply()
 	}
 
-	r.Logger.Print("✅ completed in %s - %s", time.Since(startTime), e2eTest.Description)
+	r.Logger.Print("✅ completed - %s (duration %s)", e2eTest.Name, time.Since(startTime))
```
swift1337, line 58: nit:  `duration` might be redundant 
```suggestion
	r.Logger.Print("✅ completed - %s (%s)", e2eTest.Name, time.Since(startTime))
```

swift1337 (approved):
LGTM

cmd/zetae2e/run.go
gartnera: Well it is referenced by `zetae2econfig.RunnerFromConfig` but I find this usage kinda suspect. Really we should just run everything in an `errgroup.WithContext()` which will cancel the context if anything in the group returns the error. But that's a larger refactoring.

cmd/zetae2e/local/local.go
```diff
@@ -147,7 +144,19 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 	}
 
 	// initialize context
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, legacyCancel := context.WithTimeoutCause(context.Background(), TestTimeout, ErrTopLevelTimeout)
```
ws4charlie, line 147: Are `timeoutCancel` and `systemCancel` more explicit names?

gartnera: I can call it timeoutCancel it doesn't really matter to me

ws4charlie (approved):
looks good. minor comment

lumtis (approved):

PR #3429, opened 2025-01-28
fix(zetaclient): only broadcast solana transaction if expected nonce

gartnera:
We shouldn't try to broadcast a solana transaction unless the pda nonce is expected. When we get N concurrent transactions we know that the broadcast will fail `if pda.Nonce < nonce`.

Remove the unconditional sleep so we broadcast the first transaction as fast as possible to get the nonce moving faster.

This is a quick minimal fix. We really should have a queue of outbound transactions to broadcast so we only attempt to submit one at a time.

WARN: Performance tests show significant degradation so this isn't working correctly at least on localnet.

### DIFF
--- a/zetaclient/chains/solana/signer/signer.go
+++ b/zetaclient/chains/solana/signer/signer.go
@@ -208,7 +208,9 @@ func (signer *Signer) broadcastOutbound(
 	// to tolerate tx nonce mismatch with PDA nonce or unknown RPC error
 	backOff := broadcastBackoff
 	for i := 0; i < broadcastRetries; i++ {
-		time.Sleep(backOff)
+		if i > 0 {
+			time.Sleep(backOff)
+		}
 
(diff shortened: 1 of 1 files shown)


CryptoFewka:
Is there any kind of bound sanity bound on the backoff? Would the backoff reset if zetaclientd is restarted?

gartnera:
> Is there any kind of bound sanity bound on the backoff? Would the backoff reset if zetaclientd is restarted?
Existing code:
```
// try broacasting tx with increasing backoff (1s, 2s, 4s, 8s, 16s, 32s, 64s)
```

> Would the backoff reset if zetaclientd is restarted

If just one zetaclient restarts, it will try to resign which will fail since the other zetaclients would not be trying to sign. So it would just be stuck trying to sign repeatedly and not even make it to broadcast .

PR #3433, opened 2025-01-29
feat(zetaclient): process first solana transaction faster

gartnera:
Apply a slight optimization to solana processing so that we initiate the first pending outbound transaction immediately rather than waiting for the `OutboundScheduleInterval`. `OutboundScheduleInterval` is 15 on testnet and mainnet so we could end up pointlessly waiting around in quite a few cases. And we really can't reduce that parameter since otherwise TSS will be spammed.

There should be little to no change in the performance test results in localnet since we set `OutboundScheduleInterval: 2`

Mark as consensus breaking since this would need to be deployed with coordination on testnet/mainnet to avoid signer desync.

Conflicts with https://github.com/zeta-chain/node/pull/3438

## Summary by CodeRabbit

- **New Features**
	- Enhanced cross-chain transaction (CCTX) processing logic to improve responsiveness and handling of newly created transactions
	- Introduced more flexible scheduling mechanism for transaction processing

- **Improvements**
	- Updated transaction processing conditions to allow more immediate handling of new cross-chain transactions
	- Refined keysign scheduling strategy

### DIFF
--- a/zetaclient/chains/solana/solana.go
+++ b/zetaclient/chains/solana/solana.go
@@ -141,9 +141,10 @@ func (s *Solana) scheduleCCTX(ctx context.Context) error {
 		zetaHeight = uint64(zetaBlock.Block.Height)
 
 		// #nosec G115 positive
-		interval          = uint64(s.observer.ChainParams().OutboundScheduleInterval)
-		scheduleLookahead = s.observer.ChainParams().OutboundScheduleLookahead
-		scheduleLookback  = uint64(float64(scheduleLookahead) * outboundLookbackFactor)
+		interval           = uint64(s.observer.ChainParams().OutboundScheduleInterval)
+		scheduleLookahead  = s.observer.ChainParams().OutboundScheduleLookahead
+		scheduleLookback   = uint64(float64(scheduleLookahead) * outboundLookbackFactor)
+		needsProcessingCtr = 0
 	)
 
 	cctxList, _, err := s.observer.ZetacoreClient().ListPendingCCTX(ctx, chainID)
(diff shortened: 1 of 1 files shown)


zetaclient/orchestrator/orchestrator.go
```diff
@@ -477,13 +480,23 @@ func (oc *Orchestrator) ScheduleCctxSolana(
...
 
+		shouldFastProcess := needsProcessingCtr == 0 && inboundParams.ObservedExternalHeight == zetaHeight
+		needsProcessingCtr++
+
 		// noop
 		if solSigner.IsOutboundActive(outboundID) {
 			continue
 		}
 
+		if shouldFastProcess {
```
ws4charlie, line 491: Not 100% sure about my understanding. Do you mean `Is this CCTX the first one created in the last Zeta block and it is also the first eligible cctx for a keysign`? Maybe we could make the meaning explicit like:

```
processedCCTXs := 0 // ahead of for loop

... ...

// schedule newly created cctx right away, no need to wait for next interval
// 1. schedule new cctx only when there is no other older pending cctx to process.
// 2. schedule the very first cctx (there can be multiple) created in the last Zeta block.
isCCTXNewlyCreated := inboundParams.ObservedExternalHeight == zetaHeight
shouldProcessNewCCTX := isCCTXNewlyCreated && processedCCTXs == 0
if shouldProcessNewCCTX {
    oc.logger.Info().Msgf("ScheduleCctxSolana: initiating fast process for %s", outboundID)
}

keysignIntervalArrived := nonce%interval == zetaHeight%interval

// schedule a TSS keysign
if shouldProcessNewCCTX || keysignIntervalArrived {
    go signer.TryProcessOutbound(
	ctx,
	cctx,
	observer,
	oc.zetacoreClient,
	zetaHeight,
    )
    processedCCTXs++
}
```

gartnera: Yes I like your comments, I will update.

However, the `processedCCTXs++` needs to be where it is since we should increment it even if there is any background processing happening

ws4charlie: Sounds good.

ws4charlie (approved):
it looks good

lumtis:
Curious, why using `CONSENSUS_BREAKING_ACK` if it's a ZetaClient change?

gartnera:
> Curious, why using `CONSENSUS_BREAKING_ACK` if it's a ZetaClient change?

Because if you try to run an old version of zetaclient and this one together they will not agree on when to sign the first transaction.

I'm trying to make it clear that I don't think this is safe to backport directly to the release branch without some additional coordination mechanism.

lumtis (approved):

PR #3448, opened 2025-01-31
refactor: consolidate version normalization logic

gartnera:
Deduplicate the version `ensurePrefix` logic into the constant package.


pkg/constant/version.go
lumtis, line 1: Could we add unit tests

gartnera: Yes I sorry I forgot to commit.

lumtis (approved):

kingpinXD (approved):

Issue #1345, opened 2023-10-25
Implement `RemoveInboundTracker`


lumtis:
We currently have the message `RemoveFromOutTxTracker` that allows to manually remove a out tx tracker by an admin group 1.
https://github.com/zeta-chain/node/blob/9028fe3e5ba17d01d7cb58e23594d7e559718278/x/crosschain/keeper/keeper_out_tx_tracker.go#L157

Would it make sense to have a similar mechanism for InTx tracker ?

cc @kingpinXD

kingpinXD:
It would be good, but I feel it might be lower on our priority list .
As opposed to outtxtrackers created for every outbound, the inTx trackers would be created manually or by a separate process that scans external chains looking for missed trackers. 
something like this : https://github.com/zeta-chain/node/blob/add-xcheck-cmd/cmd/xcheck/main.go 

We could add more stringent checks before creating the tracker first. What do you think?

lumtis:
Ok, thanks for the context, it sounds good to let this issue aside for the time being.

brewmaster012:
I was thinking about the other direction; 
since adding trackers is relatively easy, we can set expiration of the intx tracker (or periodically purge them). 
This also is needed for better spam (is the checking enough to ward off all spams?) control, via requiring intx tracker submitter to post some ZETA as bond.

kingpinXD:
Trackers with expiry would mean 
1.  We only query non-expired trackers 
2. We still need a mechanism to clean up expired trackers to prevent these trackers from increasing the size of the state .
This cleanup can be 
-  BeginBlock logic. 
-  A trasnaction.

I dont think adding more logic in the begin block is preferable , as it is harder to test and not under active control.

If we decide to go with the tx route I guess , implementing the original msg to delete the tracker is much simpler than having an expiry mechanism and deleting expired tracker

kingpinXD:
Confirminf as this issue is quite onl 
RemoveFromOutTxTracker has been removed and replaced by MsgAddOutboundTracker
	"/zetachain.zetacore.crosschain.MsgAddOutboundTracker",
	"/zetachain.zetacore.crosschain.MsgRemoveOutboundTracker",

All the admin messages related to trackers are under Emergency group now, I think we should add the new MsgRemoveInboundTracker under the same group

PR #3469, opened 2025-02-05
feat: add new a new msg to remove inbound trackers

kingpinXD:
# Description

Closes https://github.com/zeta-chain/node/issues/1345
 
# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality to remove inbound trackers under emergency conditions via a new CLI command.
- **Documentation**
  - Updated user documentation and API references to include details about the new removal command.
- **Tests**
  - Added tests to ensure proper authorization and behavior when removing inbound trackers, including handling non-existent cases.


x/crosschain/types/message_remove_inbound_tracker.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return []sdk.AccAddress{creator}
+}
+
+func (msg *MsgRemoveInboundTracker) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgRemoveInboundTracker) ValidateBasic() error {
```
kingpinXD, line 42: I think its okay to be be not stringent about validations 
- This message is sent by policy address 
- If the hash or chainID is wrong, it does nothing

x/authority/migrations/v3/migrate.go
```diff
@@ -0,0 +1,21 @@
...
+	SetAuthorizationList(ctx sdk.Context, list types.AuthorizationList)
+}
+
+// MigrateStore migrates the authority module state from the consensus version 2 to 3
+func MigrateStore(
+	ctx sdk.Context,
+	keeper authorityKeeper,
+) error {
+	// It is okay to not validate here, as the authorization list is fixed and will not change
+	keeper.SetAuthorizationList(ctx, types.DefaultAuthorizationsList())
```
lumtis, line 19: Not a good approach because we might already have authorization different from default, why ntot using addAuthorization?

lumtis (changes requested):

x/authority/migrations/v3/migrate.go
kingpinXD: I was considering having the DefaultAuthorizationsList mapped to a document we maintain , so that we are always confident on what the authorization list is. 
I can still modify the logic to adding if that seems better

lumtis: I think it's proper using AddAuthorization

lumtis: @kingpinXD seems we're still overwriting the list

lumtis (approved):

ws4charlie (approved):
LGTM

PR #3482, opened 2025-02-05
fix: fix evm healthcheck to reduce log spam

skosito:
# Description

on testnet there is a delay between latest block and block by number queries, resulting in big log spam

add retry to health check for block by number query so errors are removed

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Enhanced the health check process with a robust retry mechanism that makes multiple attempts with brief delays, improving resilience and reducing transient errors.

### DIFF
--- a/cmd/zetatool/config/config.go
+++ b/cmd/zetatool/config/config.go
@@ -18,7 +18,7 @@ const (
 
 func TestnetConfig() *Config {
 	return &Config{
-		ZetaChainRPC: "https://zetachain-testnet-grpc.itrocket.net:443",
+		ZetaChainRPC: "https://zetachain-testnet-rpc.itrocket.net:443",
 		EthereumRPC:  "https://ethereum-sepolia-rpc.publicnode.com",
 		ZetaChainID:  101,
 		BtcUser:      "",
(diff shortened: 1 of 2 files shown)


zetaclient/chains/evm/client/client.go
gartnera, line 1: I think I prefer the `c.HeaderByNumber(ctx, nil)` option you described in slack. Is simpler and fails more obviously.

zetaclient/chains/evm/client/client.go
```diff
@@ -94,15 +93,15 @@ func (c *Client) IsTxConfirmed(ctx context.Context, txHash string, confirmations
 // HealthCheck asserts RPC health. Returns the latest block time in UTC.
 func (c *Client) HealthCheck(ctx context.Context) (time.Time, error) {
 	// query latest block number
-	bn, err := c.BlockNumber(ctx)
+	_, err := c.BlockNumber(ctx)
 	if err != nil {
-		return time.Time{}, errors.Wrap(err, "unable to get block number")
+		return time.Time{}, errors.Wrap(err, "unable to get latest block number")
 	}
```
gartnera, line 99: just remove this?

gartnera (approved):

ws4charlie (approved):

Issue #3474, opened 2025-02-05
Add option to initialize SUI environment in local E2E tests


lumtis:
Same as Solana or TON
Adding a `--test-sui` flag in the local E2E test run, this setup necessary environment for SUI:

- Deploy the gateway in local SUI
- Deploy a token representing fungible asset
- Whitelist the token
- Integrate the chain on ZetaChain

Issue #3318, opened 2024-12-19
Add support for SUI (base integration with DepositAndCall)


lumtis:
**Motivation**

The SUI blockchain is rapidly gaining recognition in the ecosystem. https://sui.io/
Move is considered as a serious alt- EVM contender with its resource-oriented paradigm https://blog.sui.io/move-origins-sam-blackshear/

**Solution**

SUI seems to support recovering messages signed with ECDSA signature scheme: https://www.mystenlabs.com/blog/cryptography-in-sui-cross-chain-signature-verification
The solution can roughly be similar to Solana and TON when TSS congregate to sign a message with EVM private key, the message is nested in a SUI transaction and recover in a gateway contract.

**Fungible Asset Support**

SUI fungible assets uses a `SUI::Coin` type. All fungible assets, including the SUI gas token uses this object type. Therefore supporting any fungible asset might be trivial once SUI token integration is completed.
https://docs.sui.io/standards/coin

**Calling SUI Contract**

No research done yet on this part.

CharlieMc0:
I think we should tackle integrations in this order. 

- Create SUI Gateway Smart Contract(s) @brewmaster012 
- Integrate Into Developer LocalNet @brewmaster012 @fadeev 
- Integrate Into Protocol LocalNet 
- Integrate Into Testnet 
- Integrate Into Mainnet

lumtis:
yes, although

> Integrate Into Protocol LocalNet

This is the biggest and most tracked issues on the development side will be for this part

brewmaster012:
> SUI seems to support recovering messages signed with ECDSA signature scheme: https://www.mystenlabs.com/blog/cryptography-in-sui-cross-chain-signature-verification
> The solution can roughly be similar to Solana and TON when TSS congregate to sign a message with EVM private key, the message is nested in a SUI transaction and recover in a gateway contract.

It's actually easier than that, as SUI natively supports secp256k1 ECDSA signature, so there is no need for contract level verification. 

In fact they natively support multi-sig, that's good target for outbound speed optimization to bypass slow ECDSA TSS.

brewmaster012:
> I think we should tackle integrations in this order.
> 
> * Create SUI Gateway Smart Contract(s) [@brewmaster012](https://github.com/brewmaster012)
> * Integrate Into Developer LocalNet [@brewmaster012](https://github.com/brewmaster012) [@fadeev](https://github.com/fadeev)
> * Integrate Into Protocol LocalNet
> * 
[...]

Issue #3471, opened 2025-02-05
Initialize a `pkg/contracts/sui` package for parsing Deposit and DepositAndCall events


lumtis:
The package should contains utility function that allows to parse the Deposit and DepositAndCall events emitted by the gateway.

Logic related to sui contract parsing should be in this pacakge and abstracted from the ZetaClient

PR #3485, opened 2025-02-05
feat(e2e): setup sui

gartnera:
TODO:

- [x] add make target and e2e job in matrix
- [x] fund deployer account via faucet
- [x] connect to sui local validator in e2e tests
- [x] deploy contract from e2e side
- [ ] ~store contract address in config~

This can be merged if/when someone else want's to take over this process. Just let me know and I'll hand it over.

Signer and publish logic derived from [protocol-contracts-sui](https://github.com/zeta-chain/protocol-contracts-sui/tree/main/localtest) with some refactoring and improved test coverage.

Update: I think this is a good start and first PR, I think we should do the rest in separate PRs

Notes:
- There aren't any good Sui golang client libraries that fully support secp256k1, so you have to do the work yourself :/

Related to https://github.com/zeta-chain/node/issues/3318 and https://github.com/zeta-chain/node/issues/3474 and https://github.com/zeta-chain/node/pull/3465

## Summary by CodeRabbit

- **New Features**
  - Integrated Sui blockchain support with conditional test execution based on pull request labels and event types.
  - Introduced a new command-line flag and Makefile target to initiate Sui-specific tests and setup, streamlining local testing workflows.
  - Enabled Sui gateway deployment with encoded binary assets for smoother operations.

- **Tests**
  - Added tests to validate Sui cryptographic functions and address generation, ensuring reliable integration.

- **Chores**
  - Updated dependencies to include the Sui SDK and related libraries.


e2e/utils/sui/signer.go
lumtis, line 1: Can maybe be put in the package https://github.com/zeta-chain/node/issues/3471?

gartnera: Yeah I expect some of this logic to be refactored to pkg eventually. But a lot of this logic (and gateway binary) is not needed by zetaclient and would just bloat the binary size.

e2e/utils/sui/signer.go
```diff
@@ -0,0 +1,146 @@
...
+			Curve: secp256k1.S256(),
+			X:     privKey.PubKey().X(),
+			Y:     privKey.PubKey().Y(),
+		},
+		D: new(big.Int).SetBytes(secret),
+	}
+
+	return &SignerSecp256k1{
+		privkey: ecdsaPrivKey,
+	}
```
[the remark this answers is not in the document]
gartnera: We need to use https://pkg.go.dev/github.com/ethereum/go-ethereum@v1.13.15/crypto#Sign for `[R || S || V]` signature format which only takes an `ecdsa.PrivateKey`? Or is there a decent way to do this with `ecdsa.PrivateKey`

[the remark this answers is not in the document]
gartnera: So we only actually need R and S, see full sui requirements for any other readers:

> An accepted ECDSA secp256k1 and secp256r1 signature must follow:
>
> 1. The internal hash used by ECDSA must be SHA256 [SHA-2](https://en.wikipedia.org/wiki/SHA-2) hash of the transaction data. Sui uses SHA256 because it is supported by [Apple](https://developer.apple.com/forums/thread/89619), HSMs, and [cloud](https://developer.apple.com/forums/thread/89619), and it is widely adopted by [Bitcoin](https://en.bitcoin.it/wiki/Elliptic_Curve_Digital_Signature_Algorithm).
> 2. The signature must be of length 64 bytes in the form of [r, s] where the first 32 bytes are r, the second 32 bytes are s.
> 3. The r value can be between 0x1 and 0xFFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364140 (inclusive).
> 4. The s value must be in the lower half of the curve order. If the signature is too high, convert it to a lower s according to [BIP-0062](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#low-s-values-in-signatures) with the corresponding curve orders using order - s. For secp256k1, the curve order is 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141. For secp256r1, the curve order is 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551 defined in [Standards for Efficient Cryptography](https://secg.org/SEC2-Ver-1.0.pdf).
> 5. Ideally, the signature must be generated with deterministic nonce according to [RFC6979](https://www.rfc-editor.org/rfc/rfc6979).

So theoretically we can just do:

```
sig := decred_ecdsa.SignCompact(&priv, hash, false)
sig = sig[1:]
```

To get just `S || V` format?

e2e/utils/sui/signer.go
ws4charlie: yeah, `pkg/contracts/sui` is the final place to go

ws4charlie (approved):
looks good overall.

e2e/runner/setup_sui.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/block-vision/sui-go-sdk/models"
+	"github.com/block-vision/sui-go-sdk/sui"
+	"github.com/stretchr/testify/require"
+
+	sui_utils "github.com/zeta-chain/node/e2e/utils/sui"
+)
+
+func (r *E2ERunner) SetupSui(faucetURL string) {
+	r.Logger.Print("⚙️ initializing gateway program on Sui")
```
lumtis, line 14:
```suggestion
	r.Logger.Print("⚙️ initializing gateway package on Sui")
```

e2e/runner/setup_sui.go
```diff
@@ -0,0 +1,54 @@
...
+		Signature: []string{signature},
+		Options: models.SuiTransactionBlockOptions{
+			ShowEffects:        true,
+			ShowBalanceChanges: true,
+			ShowEvents:         true,
+		},
+		RequestType: "WaitForLocalExecution",
+	})
+	require.NoError(r, err)
+	fmt.Println(resp)
```
lumtis, line 51: Necessary? Let's use logger if so

lumtis (approved):
utACK

PR #3486, opened 2025-02-06
feat: add Sui chain info

lumtis:


lumtis:
Closing, I will open another one not changing ZetaChain local id as it appear non trivial and not necessary

PR #3507, opened 2025-02-10
fix: fix from field in eth receipt rpc method

skosito:
# Description

ethermint PR: https://github.com/zeta-chain/ethermint/pull/206

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **Chores**
  - Updated a key dependency to a newer version, offering improved stability and access to the latest enhancements.
  
- **Bug Fixes**
  - Enhanced error handling during transaction processing to ensure a more robust and reliable experience.

### DIFF
--- a/rpc/backend/utils.go
+++ b/rpc/backend/utils.go
@@ -110,6 +110,7 @@ func (b *Backend) getAccountNonce(
 			sender, err := ethMsg.GetSender(b.chainID)
 			if err != nil {
 				b.logger.Debug("failed to parse from field", "hash", ethMsg.Hash, "error", err.Error())
+				continue
 			}
 			if sender == accAddr {
 				nonce++


CryptoFewka:
Shouldn't this have some kind of a change log entry?

rpc/backend/utils.go
```diff
@@ -107,7 +107,10 @@ func (b *Backend) getAccountNonce(
 				break
 			}
 
-			sender := ethMsg.GetSender()
+			sender, err := ethMsg.GetSender(b.chainID)
+			if err != nil {
+				b.logger.Debug("failed to parse from field", "hash", ethMsg.Hash, "error", err.Error())
+			}
 			if sender == accAddr {
```
lumtis, line 114: Wouldn't sender be undefined/unpredictable if err != nil here? If it is fine not incrementing the nonce in this case, then it looks good to me but maybe returning nonce directly above is cleaner

swift1337 (approved):

Issue #3443, opened 2025-01-30
Add Solana Upgrade tests


lumtis:
**Describe the Issue**
Add a E2E command that run Solana tests after an upgrade such as the regular upgrade tests

The test should also simulate upgrade of the Sonala program when needed across different version (this will be needed for example when adding support for withdrawAndCall)
This is how it is done for the EVM gateway as a reference: https://github.com/zeta-chain/node/blob/develop/e2e/runner/upgrade.go

kingpinXD:
https://github.com/zeta-chain/node/issues/3535

Added some details based on my research for the last few days. Overall, I think we can just use the CLI to upgrade, it's a simple and straight-forward process, we can add some verification logic in the e2e test to verify that the contract was upgraded, which should be good enough for a test.
I

Issue #3535, opened 2025-02-14
`Solana` : Deploy and Upgrade contracts


kingpinXD:
# Deploying and Upgrading contracts on Solana

### Background 
- Deploying and upgrading contracts on solana is pretty similar. The major difference is that when deploying a contract the program ID is generated but an upgrade requires the program ID to be passed in.
- The program ID is generated from the program's keypair. The local net program is `94U5AHQMKkV5txNJ17QPXWoh474PheGou6cNP2FEuL1d`.
- Both deploy and upgrade can be done using the same command `deploy`

### Possible options 
- Deploy using the Solana CLI (Which is being used right now in the node repo)
  - Pros: 
    - Simple and easy to use
    - No need to write any code
    - Can be done in a single command `solana program deploy gateway-upgrade.so --program-id 94U5AHQMKkV5txNJ17QPXWoh474PheGou6cNP2FEuL1d`
  - Cons:
    - It uses the CLI, which might not be the best option to deploy programmatically

- Deploy using Go 
  - We can replicate the deployment logic in go and use that to deploy the program
  - The library `"github.com/gagliardetto/solana-go"` does have a deploy function but that did not work .
  - Deployment steps 
  - Create and initialize a buffer 
  - Break program data into chunks. And write each chunk into the buffer
  - Finally deploy the program using the buffer , Example code here , primarily for documentation purposes 
    https://gist.github.com/kingpinXD/98cedfb565ab88e1d2d2c42b76c519d9
  
  - Pros 
    - Can be used to deploy programmatically
  - Cons
    - We would have to write and maintain this code. I don't think maintaining would be too difficult as an instruction set that interacts with solana is pretty simple, but the documentation around it seems to be scarce, so debugging any issues might consume substantial amount of time.
    - The code would need non-trivial improvement. We write the data in chunks, we cannot write sequentially as this would take a lot of time. We would need to implement some kind of queuing mechanism for writing and verifying chunks, and retrying specific chunks if they fail.

### Recommendations
  - In my opinion, we should continue using the Solana CLI to deploy and upgrade contracts. It seems to be the most recommended way to deploy contracts on solana.
  - The CLI itself is stable and easy to interact with.

lumtis:
Looks no longer relevant

PR #3536, opened 2025-02-14
test(`e2e`): solana upgrade contract e2e test

kingpinXD:
# Description

Related PR from [protocol-contracts-solana](https://github.com/zeta-chain/protocol-contracts-solana)
https://github.com/zeta-chain/protocol-contracts-solana/pull/81

Refer this issue for more context : https://github.com/zeta-chain/node/issues/3535

Closes : https://github.com/zeta-chain/node/issues/3443

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced a new command for executing end-to-end tests focused on contract upgrades.
  - Enhanced blockchain integration with support for SOL/SPL withdrawals, deposits, and an advanced abort workflow.
  - Upgraded the local testing environment to streamline Solana contract upgrades through automated setup and deployment.

- **Bug Fixes**
  - Resolved issues in withdrawal amount verification to improve transaction reliability.

- **Tests**
  - Added enhanced verification steps to confirm successful Solana contract upgrades during testing.


Makefile
```diff
@@ -274,6 +274,11 @@ start-e2e-test: e2e-images
 	@echo "--> Starting e2e test"
 	cd contrib/localnet/ && $(DOCKER_COMPOSE) up -d
 
+start-e2e-test-upgrade-contracts: e2e-images
+	@echo "--> Starting e2e test"
+	export E2E_ARGS="--upgrade-contracts --test-solana" && \
+	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile solana -f docker-compose-upgrade-contracts.yml up -d
+
```
gartnera, line 281: How hard would it be to do this as part of the existing upgrade tests?

contrib/localnet/solana/start-solana-upgrade.sh
gartnera, line 1: Is there a reason why we shouldn't just do this on every run (even non-upgrade)?

Makefile
kingpinXD: It should not be difficult to do that 
We need to execute these two commands on the solana container .

```
# Get program ID from gateway keypair
GATEWAY_PROGRAM_ID=$(solana-keygen pubkey gateway-keypair.json)
# upgrade to the gateway-upgrade program using the program ID from keypair
solana program deploy gateway_upgrade.so --program-id "$GATEWAY_PROGRAM_ID"
```
I wanted to keep the flow similar to the existing contract upgrades for evm , but I think it will be a good idea to run on every upgrade .
Would also be good to get thoughts from the rest of the team, and  I can make the changes.

kingpinXD: Thinking about it more, I think we should do both. I don't really see the evm contracts being upgraded, so it would be good just to keep this command separate

contrib/localnet/solana/start-solana-upgrade.sh
gartnera, line 1: The downside to this approach is that we don't actually have an opportunity to run the tests with the old contract.

testutil/helpers.go
```diff
@@ -26,6 +26,12 @@ func NoError(err error) {
 	}
 }
 
+func True(b bool) {
```
lumtis, line 29: Why do we need this?

pkg/contracts/solana/pda.go
```diff
@@ -16,4 +16,15 @@ type PdaInfo struct {
...
+	// DepositPaused is the flag to indicate if the deposit is paused
+	DepositPaused bool
+}
+
+// PdaInfoUpgraded represents the PDA for the gateway program
+
+type PdaInfoUpgraded struct {
+	PdaInfo
+	// Upgraded is the flag to indicate if the PDA is upgraded
+	Upgraded bool
```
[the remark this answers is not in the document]
kingpinXD: I intilize the contract after the upgrade. So, the new field has the updated value.

We would need a realloc to migrate state when modifying the PDA. 

But Imo, that might be good to do in a follow-up pr.

contrib/localnet/solana/start-solana-upgrade.sh
kingpinXD: Modified to run it at the end of every solana test run, and removed the need to supply the flag. 
The upgrade itself does not take very long and should be fine to run for every test

pkg/contracts/solana/pda.go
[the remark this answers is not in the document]
kingpinXD: 12d20f343519abea04d271321019443587a127f3

e2e/runner/contract_upgrade_solana.go
```diff
@@ -0,0 +1,128 @@
...
+	})
+	require.NoError(r, err)
+
+	// deserialize the PDA info
+	pdaDataBefore := solanacontracts.PdaInfo{}
+	err = borsh.Deserialize(&pdaDataBefore, pdaInfo.Bytes())
+	require.NoError(r, err)
+
+	if err := triggerSolanaUpgrade(); err != nil {
+		return false
```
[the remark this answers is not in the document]
kingpinXD: We have require.True function in the primary function already. It's better to use require only at a higher level. I can print this error out, though, so that debugging is easier if it fails

e2e/runner/contract_upgrade_solana.go
```diff
@@ -0,0 +1,128 @@
...
+	// Return first 8 bytes
+	return hash[:8]
+}
+
+// triggerSolanaUpgrade triggers the Solana upgrade by creating a file `execute-update` on the Solana container
+// The shell script on the Solana container will remove the file after completing the upgrade
+// Refer: contrib/localnet/solana/start-solana.sh
+func triggerSolanaUpgrade() error {
+	// Create the execute-update file on Solana container
+	createCmd := exec.Command("ssh", "root@solana", "touch", "/data/execute-update")
```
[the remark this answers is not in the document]
kingpinXD: We can, but I found this option to be more stable. 
However, I did not spend much time investigating why direct command execution sometimes failed, as this worked fine.
I can look into the option more in the future , if you think there is a definite advantage

contrib/localnet/orchestrator/start-zetae2e.sh
```diff
@@ -10,6 +10,8 @@
 # Trap signals and forward to children
 trap 'kill -- -$$' SIGINT SIGTERM
 
+service ssh start
```
gartnera, line 13: We do `/usr/sbin/sshd` on the other scripts/services can you do the same here?

contrib/localnet/solana/Dockerfile
```diff
@@ -1,14 +1,27 @@
...
 COPY ./connected-keypair.json .
 COPY ./connected_spl.so .
 COPY ./connected_spl-keypair.json .
+COPY ./gateway_upgrade.so .
+
+# Setup SSH server
+# Expose port 22 for SSH access
+EXPOSE 22
+# Install and configure SSH server
+RUN apt update && apt install -y openssh-server
```
gartnera, line 23: Please install this before copying in the scripts and data so you don't have to reinstall this everytime something changes.

cmd/zetae2e/local/local.go
```diff
@@ -492,6 +491,13 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 
 	logger.Print("✅ e2e tests completed in %s", time.Since(testStartTime).String())
 
+	if testSolana {
+		require.True(
+			deployerRunner,
+			deployerRunner.VerifySolanaContractsUpgrade(conf.AdditionalAccounts.UserSolana.SolanaPrivateKey.String()),
+		)
+	}
+
```
gartnera, line 500: So we don't run any tests after the upgrade?

contrib/localnet/solana/start-solana.sh
```diff
@@ -11,9 +12,29 @@ sleep 5
...
+
+echo "Gateway program ID: $GATEWAY_PROGRAM_ID"
+echo "Starting upgrade loop"
+# Execute upgrade when execute-update file is found.
+# This file is created by the orchestrator when trying to upgrade the program
+while true; do
+    if [ -f "/data/execute-update" ]; then
+        echo "Found execute-update file, performing upgrade"
+        solana program deploy gateway_upgrade.so --program-id "$GATEWAY_PROGRAM_ID"
+        rm /data/execute-update
```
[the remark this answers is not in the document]
ws4charlie: I want to understand a bit. So this script will pause here until the gateway program upgrading is done on the other side? The empty file `/data/execute-update` is a signal to notify upgrading is complete?

cmd/zetae2e/local/local.go
kingpinXD: We run it at the end of a test; that should be enough to test everything we need to assert.
I am not sure how node binary upgrades and program upgrade are related , lmk if there is a specific condition you are thinking about

contrib/localnet/solana/start-solana.sh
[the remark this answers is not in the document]
kingpinXD: @skosito  https://github.com/zeta-chain/node/pull/3536/files/2ec8c1ba2adf20258cfcbac3d78672f333f50d85#r1968008460 , similar reason 

@ws4charlie  yeah

cmd/zetae2e/local/local.go
gartnera: We don't seem to do anything to validate that the new program actually works?

kingpinXD: No, but we assert the state logic.
Testing the contract itself should be a separate test; this test is for upgrading and making sure the upgrade works.

I think a better plan would be just to include solana tests into our upgrade test , which we should be doing anyway , which would be able to test the new contract , but that's out of scope for this PR

gartnera (approved):

Issue #3521, opened 2025-02-12
Separate high risk params in chain_params from lower risk operational params


CharlieMc0:
**Describe the Issue**

The `/zetachain.zetacore.observer.MsgUpdateChainParams` message is to permissive for the operational group to control. Specifically the ability to update contract addresses should be tightly controlled by the admin group. 

current chain_params:
 ``` 
{
  "chain_params": {
    "chain_id": "1",
    "confirmation_count": "14",
    "gas_price_ticker": "300",
    "inbound_ticker": "15",
    "outbound_ticker": "15",
    "watch_utxo_ticker": "0",
    "zeta_token_contract_address": "0xf091867EC603A6628eD83D274E835539D82e9cc8",
    "connector_contract_address": "0x000007Cf399229b2f5A4D043F20E90C9C98B7C6a",
    "erc20_custody_contract_address": "0x0Bad40D9e9C369f2223c835E108f43a45fd223B5",
    "outbound_schedule_interval": "64",
    "outbound_schedule_lookahead": "60",
    "ballot_threshold": "0.660000000000000000",
    "min_observer_delegation": "10000000000000000000.000000000000000000",
    "is_supported": true,
    "gateway_address": "0x48B9AACC350b20147001f88821d31731Ba4C30ed"
  }
}
```

**Expected Outcome**
A description of what outcome you are seeking.

Break up the chain_params so that contract related changes are restricted to groupAdmin and observer/Signer/TSS params are configurable by the operations group. 

Proposed Params Structure: 
```
{
  "chain_params": {
    "chain_id": "1",
    "is_supported": true,
    "ballot_threshold": "0.660000000000000000",
    "min_observer_delegation": "10000000000000000000.000000000000000000",
    "zeta_token_contract_address": "0xf091867EC603A6628eD83D274E835539D82e9cc8",
    "connector_contract_address": "0x000007Cf399229b2f5A4D043F20E90C9C98B7C6a",
    "erc20_custody_contract_address": "0x0Bad40D9e9C369f2223c835E108f43a45fd223B5",
    "gateway_address": "0x48B9AACC350b20147001f88821d31731Ba4C30ed",
    "observer_signer_params": {
      "confirmation_count": "14",
      "gas_price_ticker": "300",
      "inbound_ticker": "15",
      "outbound_ticker": "15",
      "watch_utxo_ticker": "0",
      "outbound_schedule_interval": "64",
      "outbound_schedule_lookahead": "60"
    }
  }
}
```

[...]

PR #3538, opened 2025-02-17
feat(observer): implement `MsgUpdateOperationalChainParams`

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/3521

Add `MsgUpdateOperationalChainParams`

Set `MsgUpdateChainParams` to admin (it's currently not in the migration script, my thinking is that we already plan to migrate it manually as we're setting up the operational group?) [EDIT: actually necessary for making the upgrade test work anyway]

## Summary by CodeRabbit

- **New Features**
  - Enhanced chain parameter configuration with advanced confirmation and operational settings.
  - Introduced a new tracking command and the ability to remove inbound trackers for streamlined management.

- **Bug Fixes**
  - Resolved issues with transaction scheduling and operational updates for improved reliability.

- **Documentation**
  - Updated API and user guides to reflect new operational update capabilities.

- **Tests**
  - Expanded test coverage to validate the enhanced chain parameter functionalities.


proto/zetachain/zetacore/observer/tx.proto
```diff
@@ -69,6 +72,21 @@ message MsgUpdateChainParams {
 
 message MsgUpdateChainParamsResponse {}
 
+message MsgUpdateOperationalChainParams {
+  option (cosmos.msg.v1.signer) = "creator";
```
swift1337, line 76: Should we narrow the scope of "legacy" `MsgUpdateChainParams`?

x/observer/types/chain_params.go
```diff
@@ -18,6 +17,23 @@ var (
 	DefaultBallotThreshold       = sdkmath.LegacyMustNewDecFromStr("0.66")
 )
 
+// Validate checks that the ConfirmationParams is valid
+func (cp ConfirmationParams) Validate() error {
+	if cp.SafeInboundCount == 0 {
```
swift1337, line 22: consider using `switch {}`

x/observer/types/message_update_operational_chain_params.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	cosmoserrors "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+const TypeMsgUpdateOperationalChainParams = "update_operational_chain_params"
+
+var _ sdk.Msg = &MsgUpdateChainParams{}
```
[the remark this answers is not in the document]
swift1337: agree

swift1337 (approved):
Generally LGTM, but would be great to ensure that we drop as much old code as possible w/o performing a migration

proto/zetachain/zetacore/observer/tx.proto
lumtis: `MsgUpdateChainParams` is not legacy, it will still be used for updating the sensitive data, I found it simpler to keep the message as is for now to avoid make more code change, it has been used numerous times so it is battle tested

e2e/txserver/zeta_tx_server.go
```diff
@@ -305,6 +305,22 @@ type intoAny interface {
 	AsAny() *codectypes.Any
 }
 
+// UpdateChainParams updates the chain params for the given chain ID
+// It takes into account that the required policy get updated in v28 operational -> admin
```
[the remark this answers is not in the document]
lumtis: > is there an issue to clean this up post v28?

Yeah we can create an issue for this

> also, should we check if err is policy related before retrying?

I think this is unnecessary overhead, this function is generally not supposed to fail at all if it does then the retry will fail as well

x/observer/types/chain_params.go
```diff
@@ -46,109 +61,90 @@ func (cpl ChainParamsList) Validate() error {
...
 	}
-	if params.OutboundScheduleLookahead == 0 || params.OutboundScheduleLookahead > 500 { // 500 cctxs
-		return errorsmod.Wrapf(
-			sdkerrors.ErrInvalidRequest,
+	if cp.OutboundScheduleLookahead == 0 || cp.OutboundScheduleLookahead > 500 { // 500 cctxs
+		return fmt.Errorf(
+
 			"OutboundScheduleLookahead %d out of range",
-			params.OutboundScheduleLookahead,
+			cp.OutboundScheduleLookahead,
```
kingpinXD, line 107: I feel like we can remove the upper limits for these values, they are set by a policy after all. 
Removing might simply the validation logic. 
Does not need to be done in this pr though ,

PR #3548, opened 2025-02-18
feat(zetacore)!: ensure cctx list is sorted by creation time

gartnera:
The `CctxAll()` RPC is quite useless on live networks since the results are ordered by `tss` then `chainID`. 

~Add an index on the `created_timestamp` field.~ Add an counter index which is incremented when a new CCTX is created. When listing all CCTX, use this index so that the CCTX are ordered by when they were created. You do have to specify `pagination.reversed=true` right now, but I could also invert counter field the sort descending by default.

This allows you to see the most recent CCTXs first.

Without this, it's quite annoying to calculate CCTX latency in the exporter and other places.

TODO:

- [ ] migration. please advise if this is needed.
- [x] test coverage
- [x] do we need a new API endpoint? Or put this behind a separate `sort_created_timestamp=true` parameter.

I could also introduce a `ListRecentCctx(lookbackRange)` which just uses the nonces to look for recent CCTXs. I'm mostly just interested in having a way lookup recent CCTX. It would look very similar to `ListPendingCctx()`. Update: you can't do that because it would miss inbounds.

Also:
- set max page size to 1000 on this endpoint
- add `unordered` parameter to get the old behavior

## Summary by CodeRabbit

- **New Features**
	- Introduced a timestamp-based indexing mechanism for cross-chain transactions to support more efficient and reliable data retrieval and pagination.

- **Refactor**
	- Enhanced transaction storage and query handling with improved variable naming and error processing to ensure smoother cross-chain transaction management.


x/crosschain/keeper/cctx.go
```diff
@@ -93,6 +93,10 @@ func (k Keeper) setNonceToCCTX(
 	}
 }
 
+const (
+	TSIndexKey = "cctx-ts-"
+)
+
 // SetCrossChainTx set a specific cctx in the store from its index
 func (k Keeper) SetCrossChainTx(ctx sdk.Context, cctx types.CrossChainTx) {
```
brewmaster012, line 101: consider leaving a comment here for the ordering

x/crosschain/types/keys.go
```diff
@@ -15,6 +15,9 @@ const (
 	// StoreKey defines the primary module store key
 	StoreKey = ModuleName
 
+	// StoreKey defines the primary module store key
+	StoreKeyCreatedAtIndex = ModuleName + "-createdat"
```
brewmaster012, line 19: where is this used?

brewmaster012 (approved):
LGTM

x/crosschain/keeper/cctx.go
lumtis, line 101: We should want to set it at cctx creation? Should not be in this method but in place that create the cctx for the first time

x/crosschain/keeper/cctx.go
```diff
@@ -103,7 +107,20 @@ func (k Keeper) SetCrossChainTx(ctx sdk.Context, cctx types.CrossChainTx) {
...
+	tsIndexStore := prefix.NewStore(ctx.KVStore(k.storeKey), tsIndexPrefix)
+	cctxIndex, err := cctx.GetCCTXIndexBytes()
+	if err != nil {
+		k.Logger(ctx).Error("unable to GetCCTXIndexBytes", "err", err)
+		return
+	}
+	// must use big endian so most significant bytes are first for sortability
+	createdAtBytes := sdk.Uint64ToBigEndian(uint64(cctx.CctxStatus.CreatedTimestamp))
+	indexKey := append(createdAtBytes, cctxIndex[:]...)
+	tsIndexStore.Set(indexKey, cctxIndexPrefix)
```
lumtis, line 123: Can be moved in a sub method

lumtis, line 123: This method doesn't account for collision, if two cctx are created in the ame block I believe it would overwrite?

x/crosschain/keeper/cctx.go
gartnera: Right now we use `SetCrossChainTx()` for both create and update. I could separate the codepaths but that would increase the size of this PR quite a bit.

But it might be worth it to avoid updating the index ever iteration.

gartnera:
> We already have the timestamp creation in the cctx data, wouldn't be cleaner to use a auto incremental ID instead? Would fix the collision issue

Yeah I think it would be cleaner, it would definitely make pagination easier. But it does make the migration logic a bit harder. 

We could also just skip the migration and say this only applies to new data.

x/crosschain/types/keys.go
gartnera: Removed/updated

x/crosschain/keeper/cctx.go
```diff
@@ -93,6 +93,44 @@ func (k Keeper) setNonceToCCTX(
 	}
 }
 
+// GetCctxCounter retrieves the current counter value
```
kingpinXD, line 96: We should move these counter functions to a separate file.

x/crosschain/keeper/cctx.go
```diff
@@ -93,6 +93,44 @@ func (k Keeper) setNonceToCCTX(
...
+func (k Keeper) GetCctxCounter(ctx sdk.Context) uint64 {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.CounterValueKey))
+	storedCounter := store.Get([]byte(types.CounterValueKey))
```
kingpinXD, line 99: Would be good to have a unit test for a condition when the value is not there

x/crosschain/keeper/cctx.go
kingpinXD: I agree separating create and update paths would be a decent chunk of refactor, I don't know if it's worth it for just this value.
The cctx is created using the `NewCCTX` function, and is but only saved to the state after validation. An alternative could be to add the counter to cctx object itself.
I think all the options have pros and cons and we should decide on the one that is the simplest to test ,

gartnera:
> Would need a migration to inititlize the counter

Why? It will just be zero when first read? It will then immediately be incremented to 1.

swift1337:
> > Would need a migration to inititlize the counter
> 
> Why? It will just be zero when first read? It will then immediately be incremented to 1.

Shouldn't it be `len(cctx)` in genesis migration?

gartnera:
> > > Would need a migration to inititlize the counter
> > 
> > 
> > Why? It will just be zero when first read? It will then immediately be incremented to 1.
> 
> Shouldn't it be `len(cctx)` in genesis migration?

imo no since old CCTX will not be included in this index.

x/crosschain/keeper/cctx_counter.go
```diff
@@ -0,0 +1,61 @@
...
+
+// getNextCctxCounter retrieves and increments the counter for ordering
+func (k Keeper) getNextCctxCounter(ctx sdk.Context) uint64 {
+	storedCounter := k.GetCctxCounter(ctx)
+	nextCounter := storedCounter + 1
+	k.SetCctxCounter(ctx, nextCounter)
+	return nextCounter
+}
+
+// setCctxCounterIndex sets the CCTX in the counter index
```
kingpinXD, line 44:
```suggestion
// setCctxCounterIndex sets a new CCTX in the counter index
```
or just anything which makes it clear this should only be updated once when the cctx is created

x/crosschain/keeper/cctx_counter.go
```diff
@@ -0,0 +1,61 @@
...
+// SetCctxCounter updates the current counter value
+//
+// This should only be used in setCctxCounterIndex and in state import
+func (k Keeper) SetCctxCounter(ctx sdk.Context, val uint64) {
+	store := k.getCounterValueStore(ctx)
+	store.Set([]byte(types.CounterValueKey), sdk.Uint64ToBigEndian(val))
+}
+
+// getNextCctxCounter retrieves and increments the counter for ordering
+func (k Keeper) getNextCctxCounter(ctx sdk.Context) uint64 {
```
kingpinXD, line 37: I see that this function ( and some other provate functions in this file ) are already covered under the unit tests for `SetCrossChainTx`
But still consider adding some unit tests for these functions

gartnera: So to test this we'd have to export the function right? Since the tests are in a separate package?

kingpinXD: We have a file to do that, `export_private_functions_test` .

kingpinXD (approved):

Issue #3511, opened 2025-02-10
Early ballot deletion can result in missed inbound observation, mitigation needed


ws4charlie:
**Describe the incident**
A missed inbound was detected on Feb. 9th in Base chain mainnet. The inbound was [this deposit](https://basescan.org/tx/0x66b59ad844404e91faa9587a3061e2f7af36f7a7a1a0afaca3a2efd811bc9463) ([cctx](https://zetachain-mainnet-api.itrocket.net/zeta-chain/crosschain/inboundHashToCctxData/0x66b59ad844404e91faa9587a3061e2f7af36f7a7a1a0afaca3a2efd811bc9463))

All the `zetaclients` had posted their inbound votes successfully, but there was no CCTX registered in `zetacore`.

How it was resolved:
An inbound tracker was manually created and submitted to `zetacore`, and the CCTX was then finalized right away: [the CCTX](https://zetachain.blockpi.network/lcd/v1/public/zeta-chain/crosschain/inboundHashToCctxData/0x66b59ad844404e91faa9587a3061e2f7af36f7a7a1a0afaca3a2efd811bc9463)

The root cause:
One of the votes posted by zetaclient was this: [the inbound vote](https://explorer.zetachain.com/cosmos/tx/E7FAA84651B83ED1EFDC00437C3781C3CA8140F54332A8F59059E1A1E9D41486)

A pending ballot was supposed to be created in chain state before `v27`, but the ballot was not found during the incident. So we guess that the newly installed ballot deletion logic had removed it (nothing else deletes ballot).

Basically, a ballot, once created as pending, takes at least `6/9` votes to finalize. If a pending ballot (e.g. got 5/9 votes) is not able to get enough votes within `BallotMaturityBlocks`, it will be deleted and all existing votes were lost. Late votes may also come in and re-create the ballot, but they won't be able to finalize it due to the lack of voters.

In general, an inbound ballot could stay in pending status for longer time (than `BallotMaturityBlocks`)  due to any type of delay (RPC latency, zetaclient offline, etc.)

<details>
<summary>ballot status around finalization</summary>

### 6971932

```
ballot_identifier: 0x66ef65e8c1dc5008effc66d5ff561e4d4aa1e0fe201b1bd3f8ba4097b3378d50
ballot_status: BallotInProgress
observation_type: InboundTx
voters:
- vote_type: NotYetVoted
  voter_address: zeta1p0uwsq4naus5r4l7l744upy0k8ezzj84mn40nf
[...]

PR #3553, opened 2025-02-19
refactor: add a new buffer blocks param to delay deletion of pending ballots

kingpinXD:
# Description
Closes : https://github.com/zeta-chain/node/issues/3555

Params after update on localnet
```
params:
  ballot_maturity_blocks: "100"
  block_reward_amount: "9620949074074074074.074070733466756687"
  observer_emission_percentage: "00.25"
  observer_slash_amount: "100000000000000000"
  pending_ballots_buffer_blocks: "144000"
  tss_signer_emission_percentage: "00.25"
  validator_emission_percentage: "00.50"
  ```

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Added a new parameter to delay the removal of pending ballots, enhancing ballot management.
  - Improved reward distribution by distinguishing between finalized and pending ballots.

- **Refactor**
  - Streamlined ballot clearing operations and centralized consensus version management.
  - Enhanced parameter validation and error handling for more robust operations.

- **Tests**
  - Expanded test coverage for ballot management, rewards distribution, and migration processes to ensure system stability.

- **Chores**
  - Introduced a migration path to update store parameters smoothly and updated related API documentation.

### CODE REVIEW

File: x/observer/types/ballot.go [hunk 1]
```diff
@@ -90,28 +94,35 @@ func CreateVotes(listSize int) []VoteType {
 // BuildRewardsDistribution builds the rewards distribution map for the ballot
 // It returns the total rewards units which account for the observer block rewards
 func (m Ballot) BuildRewardsDistribution(rewardsMap map[string]int64) int64 {
+	// If the ballot is in progress, return 0, we do not want to distribute rewards for in progress ballots
```
kingpinXD, line 97: The current function should replicate the older logic exactly, without any changes.  Please keep that in mind when you look it over. 
The older logic did not explicitly define a Condition for `BallotStatus_BallotInProgress` but behaved the same way as none of the cases matched.

x/observer/keeper/ballot.go
```diff
@@ -82,28 +82,58 @@ func (k Keeper) AddBallotToList(ctx sdk.Context, ballot types.Ballot) {
...
+		logBallotDeletion(ctx, ballot)
+		ballotsDeleted++
+	}
+
+	// If all ballots are finalized and deleted, then delete the list of ballots.
+	// This would avoid a second iteration at the buffered maturity height
+	// If there are a few ballots remaining we would delete the list at the buffered maturity height
+	if ballotsDeleted == len(maturedBallots.BallotsIndexList) {
+		k.DeleteBallotListForHeight(ctx, maturedBallotsHeight)
+	}
```
kingpinXD, line 114: We can consider updating the ballotListforHeight instead to remove deleted ballots, but that would involve setting the ballotList back to the store. This option looks cleaner, in my opinion

x/emissions/types/keys.go
```diff
@@ -51,5 +51,14 @@ var (
 
 	// BallotMaturityBlocks is amount of blocks needed for ballot to mature
 	// by default is set to 100
-	BallotMaturityBlocks = 100
+	BallotMaturityBlocks = 100 // approximately 9-10 minutes
+	// PendingBallotsBufferBlocks is a buffer number of blocks
+	//(in addition to BallotMaturityBlocks)
+	// that we use only for pending ballots before deleting them
+	PendingBallotsBufferBlocks = SecondsToBlocks(60 * 60 * 24 * 10) // 10 days
```
kingpinXD, line 58: Decided to use 10 days instead of 2 weeks, imo 10 days seems to be a good enough time frame

proto/zetachain/zetacore/emissions/params.proto
```diff
@@ -27,6 +28,7 @@ message Params {
     (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
     (gogoproto.nullable) = false
   ];
+  int64 pending_ballots_buffer_blocks = 12;
```
ws4charlie, line 31: nit: would you like to rename it as `pending_ballots_deletion_buffer_blocks`?

x/emissions/abci.go
```diff
@@ -143,8 +149,14 @@ func DistributeObserverRewards(
 	// Processing Step 2: Emit the observer emissions
 	keeper.EmitObserverEmissions(ctx, finalDistributionList)
 
-	// Processing Step 3: Delete all matured ballots and the ballot list
-	emissionsKeeper.GetObserverKeeper().ClearMaturedBallotsAndBallotList(ctx, maturityBlocks)
+	// Processing Step 3a: Delete all finalized ballots at `maturityBlocksForFinalizedBallots` height
+	// This step optionally deletes the `BallotListForHeight` if all ballots are finalized and deleted
```
ws4charlie, line 153: `// This step optionally deletes the `BallotListForHeight` if all ballots are finalized`?

ws4charlie (approved):
looks good overall!

lumtis:
This should close https://github.com/zeta-chain/node/issues/3511?

kingpinXD:
> This should close #3511?

Yeah it should fix it , 
I mentioned this PR on the issue already. But wanted to keep it open until we can verify that this change fixes the issue on mainnet.

x/emissions/abci.go
ws4charlie: okay, this step does deletion as well.

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,50 @@
...
+			updatedParams.BlockRewardAmount = params.BlockRewardAmount
+		}
+
+		if params.ObserverSlashAmount.GTE(sdkmath.ZeroInt()) {
+			updatedParams.ObserverSlashAmount = params.ObserverSlashAmount
+		}
+		if params.BallotMaturityBlocks > 0 {
+			updatedParams.BallotMaturityBlocks = params.BallotMaturityBlocks
+		}
+	}
```
lumtis, line 47: Why do we need all of this?
Cna't we should fetch the value and set the pendingBallotDeletion param to the default, am I missing something?

x/emissions/types/keys.go
```diff
@@ -51,5 +51,14 @@ var (
...
 )
+
+// SecondsToBlocks converts seconds to blocks.It assumes that block time is 6 seconds
+func SecondsToBlocks(secs int64) int64 {
```
lumtis, line 62: We don't need this function, let's compute value manually and put the number of blocks in the constant directly

x/emissions/types/params.go
```diff
@@ -125,6 +129,19 @@ func validateBallotMaturityBlocks(i interface{}) error {
...
 }
 
+func validatePendingBallotsBufferBlocks(i interface{}) error {
+	v, ok := i.(int64)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// pending ballots buffer blocks can be 0 to delete pending ballots immediately
+	if v < 0 {
```
lumtis, line 139: We should prevent using 0 as well

x/observer/keeper/ballot.go
```diff
@@ -82,28 +82,58 @@ func (k Keeper) AddBallotToList(ctx sdk.Context, ballot types.Ballot) {
...
 
-// ClearMaturedBallotsAndBallotList deletes all matured ballots and the list of ballots for a given height.
-// It also emits an event for each ballot deleted.
-func (k Keeper) ClearMaturedBallotsAndBallotList(ctx sdk.Context, maturityBlocksParam int64) {
-	maturedBallotsHeight := getMaturedBallotHeight(ctx, maturityBlocksParam)
-
-	// Fetch all the matured ballots, return if no matured ballots are found
-	// For the current implementation, this should never happen as ClearMaturedBallotsAndBallotList is only called after the Distribution of the rewards,
-	//	which means that there are matured ballots to be deleted
+// ClearFinalizedMaturedBallots deletes all matured and finalized ballots for a given height.
```
lumtis, line 85: Instead of two functions, it seems to me that we could have just one taking a bool in parameter, `deleteNonFinalize`

x/emissions/migrations/v5/migrate.go
kingpinXD: Its not mandatory, but the SetParams, calls validate before saving and can return an error. I just want to make sure the validate does not fail .
We don't expect any of the values to be invalid, to be fair, but I don't see a problem with keeping this logic

x/emissions/types/keys.go
kingpinXD: Imo this makes it more readable , we normally discuss things like this in terms of `time` and not `blocks`

x/emissions/types/params.go
kingpinXD: okay changed to 1 .
But 0 would allow us to replicate the current behaviour if we ever need to

x/emissions/migrations/v3/migrate.go
```diff
@@ -1,50 +0,0 @@
-package v3
```
kingpinXD, line 1: Remove the v3 migration script as it was not compatible with this change 
Rest of the historical migration scripts would be removed as part of this https://github.com/zeta-chain/node/issues/3585

x/emissions/types/keys.go
lumtis: You just need to say in comment how it has been computed, it's straightforward to check

lumtis (commented):
LGTM but I think we should remove SecondsToBlocks, let's not add new functions when not necessary

lumtis (approved):

Issue #3549, opened 2025-02-18
Athens: ZetaClient has difficulty keeping up with observing Arbitrum Sepolia


CryptoFewka:
**Describe the Bug**
Our Observers have trouble keeping up with Arbitrum Sepolia. Specifically, we have an increasing value from subtracting  `zetachain_exporter_rpc_block_height` for Arbitrum Sepolia from each of multiple validator's `zetaclient_last_scanned_block_number` for Arbitrum Sepolia. This indicates that the gap between the blockchain's highest block and the validator's highest scanned block is increasing.

**To Reproduce**
Investigate our Athens validators in ap-southeast-1 with this query in grafana:
```
max by(job) (scalar(max(zetachain_exporter_rpc_block_height{chain="421614"}))- max by(job) (zetaclient_last_scanned_block_number{chain="arbitrum_sepolia",job=~"validator0-ap-southeast-1-testnet|validator1-ap-southeast-1-testnet"}))
```

**Expected Behavior**
This value should remain steady and flat at the number of blocks we require for confirmation on Arbitrum Sepolia

**Screenshots**
Continuing to fall behind in ap-southeast-1:
<img width="817" alt="Image" src="https://github.com/user-attachments/assets/dcf3cf66-c2c3-4f16-a532-c8efcfd23e5a" />
Fell behind but caught back up in us-east-1:
<img width="814" alt="Image" src="https://github.com/user-attachments/assets/c07252d9-ed6a-46c7-9003-31807af4b11a" />

**Environment (please complete the following information):**
N/A

gartnera:
Minimal solution implemented here, https://github.com/zeta-chain/node/pull/3554

Should probably be moved to a chain parameter.

Issue #3525, opened 2025-02-13
Optimize RPC use from the Observer Nodes


CharlieMc0:
**Is your feature request related to a problem? Please describe.**

Observer nodes make frequent RPC calls. The volume of RPC calls can be very expensive for Observer operators. They need to run an RPC node of their own with high spec to handle the load or rely on an RPC platform that charges based on volume or compute units. This can get quite expensive for operators. 

**Describe the solution you'd like**
Review ZetaClient to see if there are optimization that can be made to reduce the frequency of RPC calls. 

**Motivation and Context**
Reduce costs for Observer node operators.

lumtis:
We should have some way to monitor all the RPC calls and being able to produce a report after local testing to set the part of the code that can be optimized.

CryptoFewka:
> We should have some way to monitor all the RPC calls and being able to produce a report after local testing to set the part of the code that can be optimized.
I would prefer this to be done by ZetaClient which would maintain and expose counters for external calls by method, chain, and success. 

If need be we can accomplish this by placing a proxy that captures and counts these in between the local test and the internet. 

I would expect it to take the same amount of development time to create either solution, but the teams doing the work would be different. One group may be more preferable than the other due to resource constraints.

morde08:
Since v29 has gone live on Zetachain Mainnet our Zetaclient RPC usage has gone down by roughly 70% showing that the optimizations have worked to reduce RPC usage.

![Image](https://github.com/user-attachments/assets/1b9b33d2-e3d5-4a3d-ba4f-afd4eada35fd)

PR #3570, opened 2025-02-21
refactor(zetaclient): make EVM getLogs call once

gartnera:
Make the `getLogs` call once then filter the results for each event type. Now instead of making 3 `getLogs` calls we only make 1!

Update: also combine the v1 `getLogs` calls. Is it possible to deploy on a chain without zeta connector having a valid address?

Related to https://github.com/zeta-chain/node/issues/3549 https://github.com/zeta-chain/node/issues/2493 https://github.com/zeta-chain/node/issues/3525

## Summary by CodeRabbit

- **Refactor**
  - Consolidated multiple steps in event log retrieval into a single, streamlined process for enhanced efficiency.
  - Updated the processing of deposit and call events with improved error handling and more reliable validation.
  - Adjusted the initialization workflow to better support unified log management.

### DIFF
--- a/zetaclient/chains/evm/observer/inbound.go
+++ b/zetaclient/chains/evm/observer/inbound.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 
 	sdkmath "cosmossdk.io/math"
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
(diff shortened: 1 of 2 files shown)


swift1337 (approved):
LGTM, nice optimization!

ws4charlie (approved):
good point!

Issue #3573, opened 2025-02-24
Log in goroutine after TestOrchestratorV2/updates_app_context has completed


gartnera:
```
PASS
panic: Log in goroutine after TestOrchestratorV2/updates_app_context has completed: {"level":"info","module":"scheduler","task.name":"update_metrics","task.group":"orchestrator","task.type":"block_ticker","worker.name":"unknown","message":"Background task completed"}
	 [recovered]
[...]

PR #3574, opened 2025-02-24
fix(testlog): fix panics after test finished

swift1337:
Closes https://github.com/zeta-chain/node/issues/3573 🤞

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the logging process in testing utilities for a more consistent and clear output.
  - Enhanced error reporting during log operations, ensuring better diagnostic feedback during tests.


gartnera (approved):
Not exactly sure what the fix is here but we can give it a shot

swift1337:
> Not exactly sure what the fix is here but we can give it a shot

the fix is that prevously testlog had `io.Multiwriter(...)` that wrote to buffer AND `zerolog.NewTestWriter(t)`. And we were checking for panic only in the buffer part. Now we write logs sequentially in the same routine and thus can recover from panic

PR #3577, opened 2025-02-24
fix(e2e): only write address book once

gartnera:
When running in persistent mode, we shouldn't try to write out the address book file a second time. When you restart the container, the other zetaclient containers may not be up yet and the dns resolution performed by this command may fail.

## Summary by CodeRabbit

- **Chores**
	- Improved the startup process by adding a safety check to ensure critical initialization data isn’t overwritten unnecessarily, enhancing system stability.


lumtis (approved):
LGTM

PR #3578, opened 2025-02-25
feat: add disable_tss_block_scan parameter

gartnera:
Add observer parameter to skip actions which require scanning individual blocks.

Set this by default on localnet. Remove legacy deposit usage in normal tests. Set `skip_block_scan = false` when running the legacy tests. This should give us full coverage of the combinations of this parameter.

Migrate the observer params so that the parameter is set correctly for all chains. We will immediately turn off block scanning for everything on testnet except ethereum. We will announce deprecation timeline on mainnet and process via normal group proposals.

Closes https://github.com/zeta-chain/node/issues/3549

Closes https://github.com/zeta-chain/node/issues/3579

Related to https://github.com/zeta-chain/node/issues/3525

## Summary by CodeRabbit

- **New Features**
  - Enhanced token deposit functionality with upgraded deployment methods for Ether and ERC20 tokens.
  - Introduced a configurable option to control block scanning behavior.

- **Documentation**
  - Updated API and operational parameter documentation to reflect the new block scanning option.

- **Tests & Chores**
  - Added migration routines and expanded test coverage to support updated chain parameters and configuration settings.


proto/zetachain/zetacore/observer/params.proto
```diff
@@ -34,6 +34,10 @@ message ChainParams {
 
   // Advanced confirmation parameters for chain to support fast observation
   ConfirmationParams confirmation_params = 18;
+
+  // Skip actions that require scanning the contents of each block.
+  // The main thing this disables is transfers directly to the TSS address.
+  bool skip_block_scan = 19;
```
lumtis, line 40: Wouldn't disable better wording than skip?

Skip sounds like skipping an optional step in a process for me, while this disable a feature by itself

x/observer/migrations/v11/migrate.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type observerKeeper interface {
+	GetChainParamsList(ctx sdk.Context) (val types.ChainParamsList, found bool)
+	SetChainParamsList(ctx sdk.Context, chainParams types.ChainParamsList)
+}
+
+var skipBlockScanChainIDs = []int64{
+	// localnet
+	chains.GoerliLocalnet.ChainId,
```
lumtis, line 20: I suggest to keep it for the local evm because parsing all blocks will still be supported for at least one evm chains on each live network

proto/zetachain/zetacore/observer/params.proto
gartnera: Updated

e2e/runner/evm.go
```diff
@@ -38,6 +41,24 @@ func (r *E2ERunner) ETHDeposit(
 	return tx
 }
 
+// DepositEther sends Ethers into ZEVM using V2 protocol contracts
```
ws4charlie, line 44:
```suggestion
// DepositEtherDeployer sends Ethers into ZEVM using V2 protocol contracts
```

x/observer/keeper/migrator.go
```diff
@@ -61,3 +62,8 @@ func (m Migrator) Migrate8to9(ctx sdk.Context) error {
 func (m Migrator) Migrate9to10(ctx sdk.Context) error {
 	return v10.MigrateStore(ctx, m.observerKeeper)
 }
+
+// Migrate10to11 migrates the store from consensus version 10 to 11
+func (m Migrator) Migrate10to11(ctx sdk.Context) error {
```
ws4charlie, line 67: Will this go to the release to be cut next Monday? If it is, you could just reuse the existing `Migrate9to10`, otherwise just ignore this comment.

ws4charlie (commented):
overall looks good. Left minor comment

x/observer/keeper/migrator.go
gartnera: I'd prefer to keep the migrations separate as the `Migrate9to10` has already run on developnet. These migrations are very light so I don't think we need to worry about it.

If we were doing a pass over all CCTX or something heavy like that I think it would be worth considering.

docs/spec/observer/messages.md
```diff
@@ -201,6 +201,7 @@ message MsgUpdateOperationalChainParams {
 	int64 outbound_schedule_interval = 7;
 	int64 outbound_schedule_lookahead = 8;
 	ConfirmationParams confirmation_params = 9;
+	bool disable_block_scan = 10;
```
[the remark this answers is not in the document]
gartnera: We are also skipping `filterTSSOutbound`. Doesn't really matter to me.

x/observer/keeper/migrator.go
ws4charlie: okay.

docs/spec/observer/messages.md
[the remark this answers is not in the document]
ws4charlie: I think `filterTSSOutbound` is a pure supplement to outbound trackers. This can be moved out to `outbound.go` in future and just use cached blocks (without RPC call).

x/observer/keeper/migrator.go
kingpinXD: I agree with the original comment; It would be better to reuse the existing migration if  possible. 
Assuming developnet can be reset easily ?

If not, it would be great to atleast manually check that both the migrations are running and all the values are correctly set

proto/zetachain/zetacore/observer/params.proto
```diff
@@ -34,6 +34,10 @@ message ChainParams {
 
   // Advanced confirmation parameters for chain to support fast observation
   ConfirmationParams confirmation_params = 18;
+
+  // Skip actions that require scanning the contents of each block.
+  // The main thing this disables is transfers directly to the TSS address.
+  bool disable_block_scan = 19;
```
kingpinXD, line 40: Would disable_tss_inbound , be more appropriate. 
Will we use this param for any other checks in the future?

x/observer/migrations/v11/migrate.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// testnets
+	chains.AvalancheTestnet.ChainId,
+	chains.ArbitrumSepolia.ChainId,
+	chains.BscTestnet.ChainId,
+	chains.BaseSepolia.ChainId,
+	chains.Amoy.ChainId,
+
+	// mainnets
+	// the other mainnets will be deprecated later
```
kingpinXD, line 30: Does this need to be tracked via an issue ?

x/observer/migrations/v11/migrate_test.go
```diff
@@ -0,0 +1,57 @@
...
+		// migrate the store
+		err := v11.MigrateStore(ctx, *k)
+		require.NoError(t, err)
+
+		// ensure we still have same number of chain params after migration
+		newChainParams, found := k.GetChainParamsList(ctx)
+		require.True(t, found)
+		require.Len(t, newChainParams.ChainParams, len(oldChainParams.ChainParams))
+
+		for _, params := range newChainParams.ChainParams {
```
kingpinXD, line 48: NIT: Maybe add a check to make other values are identical

docs/spec/observer/messages.md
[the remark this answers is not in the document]
gartnera: You'd need to make sure you only run it after you've already run `ob.observeTSSReceive` or they would race populating the block cache.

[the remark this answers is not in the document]
ws4charlie: okay. looks like separating the two is difficult.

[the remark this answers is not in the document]
kingpinXD: imo if the parameter is only affecting TSS (inbound and outbound), it should have the keyword `TSS` in the name.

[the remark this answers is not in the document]
ws4charlie: I agree, having `tss` as part of the name makes its purpose explicit.

x/observer/migrations/v11/migrate.go
gartnera: Issue linked

x/observer/keeper/migrator.go
gartnera: I can reset it but it takes some time

docs/spec/observer/messages.md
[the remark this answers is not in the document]
gartnera: Updated to `disable_tss_block_scan`

e2e/utils/zetacore.go
```diff
@@ -81,7 +81,15 @@ func WaitCctxsMinedByInboundHash(
 		var cctxs []*crosschaintypes.CrossChainTx
 
 		timedOut := time.Since(startTime) > timeout
-		require.False(t, timedOut, "waiting cctx timeout, cctx not mined, inbound hash: %s", inboundHash)
+		if timedOut {
+			require.False(t, timedOut, "waiting cctx timeout, cctx not mined, inbound hash: %s", inboundHash)
```
ws4charlie, line 85: when `timedOut == true` does the program execution terminate right on this line, before `return cctxs`?

e2e/utils/zetacore.go
```diff
@@ -81,7 +81,15 @@ func WaitCctxsMinedByInboundHash(
...
+			return cctxs
+		}
+
+		if ctx.Err() != nil {
+			require.NoError(t, ctx.Err())
```
ws4charlie, line 90: similar question, does the execution stop right on this line?

e2e/utils/zetacore.go
gartnera: No execution does not stop, that was why I wanted to add the return.

I think we should actually `panic()` in the `func (r *E2ERunner) FailNow() {` to halt execution of the test immediately.

ws4charlie: Agree, fail the test fast will be better in future. I'm was curious.

gartnera: Updated to panic in d11c53476 so that the test exits faster.

lumtis (approved):
LGTM

ws4charlie (approved):

kingpinXD (approved):
lgtm

x/observer/migrations/v11/migrate_test.go
gartnera: bc50d5896

PR #3589, opened 2025-02-25
chore: remove gov v1beta1 endpoints from docs

gartnera:
Remove all `/cosmos/gov/v1beta1/` endpoints from the openapi docs. These endpoints have been broken since the sdk v0.47 upgrade. Users should just use the `/cosmos/gov/v1/` endpoints which actually work.

Also restore missing upgrade module docs.


lumtis (approved):
LGTM

swift1337 (approved):

Issue #2966, opened 2024-10-03
Develop mechanism to refund unused tokens in connected chain call


lumtis:
**Describe the Issue**
In connected chain calls, the gateway give approval of the asset to the target contract, for unused funds the amount is sent back to the custody.

This amount should be refunded to the sender instead.

**Considered solution**

Add a remaining value in the CCTX outbound data. This value is refunded back to the sender when the outbound is processed.

kingpinXD:
Listing down the changes needed so as to to avoid changes during implementation
Using a withdraw and call as example (v2)


a) Smart contract changes 
- The ERC20Custody , calls the executeWithERC20

```
        // Transfer the tokens to the Gateway contract
        IERC20(token).safeTransfer(address(gateway), amount);

        // Forward the call to the Gateway contract
        gateway.executeWithERC20(token, to, amount, data);

        emit WithdrawAndCall(token, to, amount, data);
    }
```
The executeWithERC20 retunrs the remaining balance to the AssetHandler
```
  // Transfer any remaining tokens back to the custody/connector contract
        uint256 remainingBalance = IERC20(token).balanceOf(address(this));
        if (remainingBalance > 0) {
            transferToAssetHandler(token, remainingBalance);
        }
```

To refund this amount to the original sender we would need to add this amount to the `WithdrawAndCall` event 
This can be done either by 
1. Calcualating balance of custody before and after the call 
2. Returning the remaining balance from `executeWithERC20`.

I am leaning more towards  option 2, its more straightforward as we are already calcuating the remaining balance .



b) Zetaclient changes 
The `parseOutboundReceivedValue` already uses this event , we would be able to parse the new value and broadcast it back to zetacore when posting outbound Vote 


c) Zetacore Changes 
- The VoteOutbound message should include the new field 
- We should send these tokens nack to the caller ( Handle if the the trasfer fails )
- We can add this value to the CCTX as well , to track , However the transfer can be done in the same block as finalization for the outbound vote.

kingpinXD:
Listing down the changes needed to to avoid changes during implementation.

Using a withdraw and call as example (v2)
## Smart contract changes

The ERC20Custody , calls the executeWithERC20
```solidity
IERC20(token).safeTransfer(address(gateway), amount);
        // Forward the call to the Gateway contract
        gateway.executeWithERC20(token, to, amount, data);
        emit WithdrawAndCall(token, to, amount, data);
    }
```
```solidity
// Transfer the tokens to the Gateway contract
[...]

PR #3619, opened 2025-03-03
Feat: WIP, support for fast outbound confirmation for EVM/Bitcoin chain

ws4charlie:
# Description

This PR is still in the progress.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


zetaclient/chains/evm/signer/outbound_tracker_reporter.go
```diff
@@ -68,6 +69,15 @@ func (signer *Signer) reportToOutboundTracker(
 				continue
 			}
 
+			// stop if the cctx is already finalized
```
ws4charlie, line 72: I saw a lots unnecessary outbound tracker submissions in E2E tests. I tracked down the root cause: 

the outbound ballot finalizes so fast in local network and the tracker reporter is still submitting tracker without checking the CCTX's status. This will not happen on mainnet due to the slow finality.

e2e/runner/zevm.go
```diff
@@ -23,6 +24,24 @@ import (
...
+	revertOptions gatewayzevm.RevertOptions,
+) *ethtypes.Transaction {
+	tx, err := r.GatewayZEVM.Withdraw(
+		r.ZEVMAuth,
+		[]byte(receiver.EncodeAddress()),
+		amount,
+		r.BTCZRC20Addr,
+		revertOptions,
+	)
+	require.NoError(r, err)
```
swift1337, line 40: should we also wait for the receipt?

e2e/runner/zevm.go
```diff
@@ -23,6 +24,24 @@ import (
...
+
+	return tx
```
swift1337, line 42: should we also wait for the receipt and successful execution?

e2e/utils/zetacore.go
```diff
@@ -192,6 +208,53 @@ func WaitCCTXMinedByIndex(
 	}
 }
 
+// WaitOutboundTracker wait for outbound tracker to be filled with 'hashCount' hashes
+func WaitOutboundTracker(
+	ctx context.Context,
+	client crosschaintypes.QueryClient,
+	chainID int64,
+	nonce uint64,
+	hashCount int,
```
swift1337, line 217: why do we need this `hashCount` param?

x/observer/types/chain_params.go
```diff
@@ -171,6 +171,12 @@ func (cp ChainParams) InboundConfirmationFast() uint64 {
 	return cp.ConfirmationParams.SafeInboundCount
 }
 
+// IsOutboundFastConfirmationEnabled returns true if fast outbound confirmation is enabled.
+func (cp ChainParams) IsOutboundFastConfirmationEnabled() bool {
```
swift1337, line 175: I don't understand the benefit for fast outbounds. It adds complexity to the codebase.

zetaclient/chains/base/observer.go
```diff
@@ -425,6 +420,48 @@ func (ob *Observer) PostVoteInbound(
...
+		logs.FieldConfirmationMode: msg.ConfirmationMode.String(),
+		logs.FieldBallot:           msg.Digest(),
+	}
+
+	// post vote to zetacore
+	zetaHash, _, err := ob.ZetacoreClient().PostVoteOutbound(ctx, gasLimit, retryGasLimit, msg)
+
+	switch {
+	case err != nil:
+		ob.logger.Inbound.Error().Err(err).Fields(lf).Msg("posted outbound vote: failed")
```
swift1337, line 452: no need to log this as it should be logged by the caller

swift1337 (changes requested):
Let's chat about the importance of this feature. 

I'm not sure how it would benefit users in general, but I can see it being useful in a few rare situations where we need to perform something in ZEVM after confirming an outbound. 

Let me know what you think. cc @lumtis

lumtis:
> Let's chat about the importance of this feature.
> 
> I'm not sure how it would benefit users in general, but I can see it being useful in a few rare situations where we need to perform something in ZEVM after confirming an outbound.
> 
> Let me know what you think. cc @lumtis

Exactly, this has benefits once you have logic to handle response from the cctx.

There will be benefits once we integrate https://github.com/zeta-chain/node/issues/2966, user will have funds refunded faster. Although this feature is mostly benefitial for crosschain call that doesn't concern fast observation for now.
Even in the future, with a hook like `onAcknowledgement` the fast observation might be disable because you couldn't recover from double spend.

In general the benefits will be minimal in all cases. If this PR is not too complex let's still move forward and merge it, you can disable the fast move by setting the same value as safe.

e2e/utils/zetacore.go
ws4charlie: `hashCount > 1` can happen on outbound replacement (gas price increase), so I wanted to make it general.

e2e/runner/zevm.go
ws4charlie: I make it consistent. The ZEVM call usually fails right away unless ZetaChain halts.

x/observer/types/chain_params.go
ws4charlie: It just finalizes CCTX faster. For bitcoin, it saves `10~20` minutes. For EVM chains, Ethereum saves up to `3` minutes. For polygon, it saves up to `5` minutes. We can decide which chain to apply if it's beneficial.

zetaclient/chains/base/observer.go
ws4charlie: indeed. Removed https://github.com/zeta-chain/node/pull/3619/commits/cec606e69cebbac9569ea5c862e45fc796ef2a25

zetaclient/chains/bitcoin/observer/outbound.go
```diff
@@ -168,53 +166,63 @@ func (ob *Observer) VoteOutboundIfConfirmed(ctx context.Context, cctx *crosschai
...
+	cctxIndex string,
+	outboundHash string,
+	blockHeight uint64,
+	amountInSat *big.Int,
+	nonce uint64,
+	coinType coin.CoinType,
+) *crosschaintypes.MsgVoteOutbound {
+	const (
 		// not used with Bitcoin
+		outboundGasUsed  = 0
```
swift1337, line 197: I think `outboundGasUsed` & `outboundGasPrice` should be provided from on-chain data to topup BTC stability pool and partially refund BTC.BTC if we spent less. cc @lumtis

zetaclient/chains/bitcoin/observer/outbound.go
```diff
@@ -168,53 +166,63 @@ func (ob *Observer) VoteOutboundIfConfirmed(ctx context.Context, cctx *crosschai
...
+		outboundStatus   = chains.ReceiveStatus_success // Bitcoin outbound never fails
+	)
+
+	// determine confirmation mode
+	confirmationMode := crosschaintypes.ConfirmationMode_FAST
+	if ob.IsBlockConfirmedForOutboundSafe(blockHeight) {
+		confirmationMode = crosschaintypes.ConfirmationMode_SAFE
+	}
+
+	msg := crosschaintypes.NewMsgVoteOutbound(
```
swift1337, line 209:
```suggestion
return crosschaintypes.NewMsgVoteOutbound(
```

zetaclient/chains/evm/observer/outbound.go
```diff
@@ -213,17 +140,60 @@ func (ob *Observer) VoteOutboundIfConfirmed(
...
+	nonce uint64,
+	coinType coin.CoinType,
+) *crosschaintypes.MsgVoteOutbound {
+	// determine confirmation mode
+	confirmationMode := crosschaintypes.ConfirmationMode_FAST
+	if ob.IsBlockConfirmedForOutboundSafe(receipt.BlockNumber.Uint64()) {
+		confirmationMode = crosschaintypes.ConfirmationMode_SAFE
+	}
+
+	msg := crosschaintypes.NewMsgVoteOutbound(
```
swift1337, line 178:
```suggestion
return crosschaintypes.NewMsgVoteOutbound(
```

swift1337 (changes requested):

zetaclient/chains/bitcoin/observer/outbound.go
ws4charlie: updated https://github.com/zeta-chain/node/pull/3619/commits/093590de94c43e1d9da63e9fc4167fef09ec1135

zetaclient/chains/evm/observer/outbound.go
ws4charlie: updated https://github.com/zeta-chain/node/pull/3619/commits/093590de94c43e1d9da63e9fc4167fef09ec1135

zetaclient/chains/bitcoin/observer/outbound.go
ws4charlie: The reason is that Bitcoin miner always takes 100% gas fee contained in the outbound, and there is no concept of `gasLimit/gasUsed` in the Bitcoin protocol

swift1337 (approved):

PR #3629, opened 2025-03-04
refactor: reduce protocol fee for ZETA transfer

lumtis:
# Description

Reduce from 2 to 0.5 ZETA

### DIFF
--- a/x/crosschain/types/keys.go
+++ b/x/crosschain/types/keys.go
@@ -23,7 +23,7 @@ const (
 	// we keep the current value for backward compatibility
 	MemStoreKey = "mem_metacore"
 
-	ProtocolFee = 2000000000000000000
+	ProtocolFee = 500000000000000000
 
 	// CCTXIndexLength is the length of a crosschain transaction index
 	CCTXIndexLength          = 66


x/crosschain/types/keys.go
```diff
@@ -23,7 +23,7 @@ const (
 	// we keep the current value for backward compatibility
 	MemStoreKey = "mem_metacore"
 
-	ProtocolFee = 2000000000000000000
+	ProtocolFee = 500000000000000000
```
swift1337, line 26: Let's create an issue for making this a gov proposal

swift1337 (approved):

x/crosschain/types/keys.go
lumtis: Should be in the smart contract yes, created this issue https://github.com/zeta-chain/protocol-contracts/issues/464

lumtis, line 26:
```suggestion
	ProtocolFee = 250000000000000000
```

CharlieMc0 (approved):

x/crosschain/types/keys.go
```diff
@@ -23,7 +23,7 @@ const (
 	// we keep the current value for backward compatibility
 	MemStoreKey = "mem_metacore"
 
-	ProtocolFee = 2000000000000000000
+	ProtocolFee = 250000000000000000
```
kingpinXD, line 26: should this be 
500000000000000000 , according to the description

kingpinXD: okay I see there are a few comments around this already. Maybe update the description based on the final decision

Issue #3383, opened 2025-01-21
Improve Outbound performance


lumtis:
This EPIC is aimed to track all issues to improve the performance of the TSS and outbound signing more generally:
- keysigning delay
- keysigning failure rate
- p2p connectivity

PR #3633, opened 2025-03-04
refactor(zetaclient): wait for expected nonce and defer relayer signing

gartnera:
There is no reason to even try to submit the transaction until the nonce is in the expected range. Update: that would not be true if we enabled skipping the preflight checks (which may be a better solution than this).

If we defer relayer transaction signing until we're closer to actually being able to relay the transaction, we can avoid starting the two minute `recentBlockhash` timer ([ref](https://solana.com/docs/core/transactions#recent-blockhash)).

We need to continue doing TSS signing immediately to avoid desynchronization.

TODO:
- [ ] deduplicate `signer.client.GetAccountInfoWithOpts` calls. We should only make 1 call even if we have N pending transactions
- [ ] inject the fallbackTx as `txGetterT` and resolve it only when actually needed. Although I don't exactly understand why we actually would want to do this. Testing for the `NonceMismatch` will have a lot of false positives since we see it during normal operations.

Note, I have another older version of this logic that make the relayer more like a queue. But that has a bunch of concurrency overhead that could be a bit of maintenance burden.

Related to https://github.com/zeta-chain/node/issues/3383

### Performance Test Comparison

We see 10-20% reductions in latency since we are spending less time in the relayer backoff loop

#### [Before](https://github.com/zeta-chain/node/actions/runs/13647314383/job/38148522148) (nightly run)

```
perf_sol_wit | Latency report:
perf_sol_wit | min:  42.59
perf_sol_wit | max:  649.80
perf_sol_wit | mean: 330.21
perf_sol_wit | std:  182.90
perf_sol_wit | p50:  314.24
perf_sol_wit | p75:  480.37
perf_sol_wit | p90:  576.56
perf_sol_wit | p95:  616.17
```

#### [After](https://github.com/zeta-chain/node/actions/runs/13664327834/job/38202479261?pr=3633) (this PR)

```
perf_sol_wit | Latency report:
perf_sol_wit | min:  26.58
perf_sol_wit | max:  568.70
perf_sol_wit | mean: 296.78
perf_sol_wit | std:  159.03
perf_sol_wit | p50:  296.59
perf_sol_wit | p75:  425.42
perf_sol_wit | p90:  512.57
perf_sol_wit | p95:  531.25
```

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced retrieval of account details by allowing additional query options.
  
- **Refactor**
  - Improved outbound transaction handling by deferring transaction creation, which streamlines error management and processing efficiency.

### DIFF
--- a/zetaclient/chains/solana/signer/signer.go
+++ b/zetaclient/chains/solana/signer/signer.go
@@ -42,6 +42,8 @@ const (
 	SolanaMaxComputeBudget = 1_400_000
 )
 
+type txGetterT func() (*solana.Transaction, error)
+
 // Signer deals with signing Solana transactions and implements the ChainSigner interface
 type Signer struct {
 	*base.Signer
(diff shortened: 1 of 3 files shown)


zetaclient/chains/solana/signer/signer.go
```diff
@@ -209,6 +211,56 @@ func (signer *Signer) TryProcessOutbound(
 	// set relayer balance metrics
 	signer.SetRelayerBalanceMetrics(ctx)
 
+	for range 30 {
```
lumtis, line 214: Why this value? Can we use a constant?

lumtis, line 214: Could we maybe consider a sub function? This function starts getting very complex

gartnera: No significance, just picked a value

zetaclient/chains/solana/signer/signer.go
```diff
@@ -334,12 +357,59 @@ func (signer *Signer) broadcastOutbound(
...
+) bool {
+	for range 30 {
+		if ctx.Err() != nil {
+			return false
+		}
+		sleepDuration := time.Second
+		pdaInfo, err := signer.client.GetAccountInfoWithOpts(
+			ctx,
+			signer.pda,
+			&rpc.GetAccountInfoOpts{Commitment: rpc.CommitmentProcessed},
```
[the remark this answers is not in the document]
gartnera: It's enough so that `signer.client.SendTransactionWithOpts()` will no longer throw errors. Maybe I should call it `isTransactionSendable()`

If a transaction is dropped, we'll have to wait for the next retry interval. At that point, the account info should have reverted back to the previous value.

zetaclient/chains/solana/signer/signer.go
```diff
@@ -213,6 +215,27 @@ func (signer *Signer) TryProcessOutbound(
 	// set relayer balance metrics
 	signer.SetRelayerBalanceMetrics(ctx)
 
+	isNonceRelayable := signer.waitNonceRelayable(ctx, nonce, logger)
```
ws4charlie, line 218: If you really want a matching Solana `pending nonce`, you can do it right [here](https://github.com/zeta-chain/node/blob/solana-pda-nonce-preflight-v2/zetaclient/chains/solana/signer/signer.go#L323) very simply in the tx broadcaster logic, and we're already waiting for up to `64` seconds in there.

gartnera: Note that I'm using confirmation:processed not the default confirmation:finalized which that code is using. I think I'd actually need to update this code a bit to not give up when `nonce > pda.Nonce` when `confirmation:processed` as the nonce could go backwards on a failure

ws4charlie: You can just switch the commitment level to `CommitmentProcessed` to give it a try. We must face a harsh reality: there is no accurate way to tell the exact pending nonce in an off-chain program (here is `zetaclientd`).

- zetaclient may connect to a load balancer, which wraps a group of endpoints.
- each RPC endpoint have their own Solana chain state which may not be in good sync.

The only thing we can do is to retry and optimize the timing, leaving the Solana endpoint itself resolve all other conflicts.

gartnera: I both of those are very valid concerns but I think we could also try to solve those at the infrastructure level by selecting a higher quality RPC provider.

gartnera: Do note that this pda nonce check loop happens before the relayer transaction signing so that we can avoid starting the two minute recent blockhash timer until we're in a valid range.

Issue #3635, opened 2025-03-05
`zetacore` `zetaclient` : removing unnessary inbound trackers


kingpinXD:
We do not have a mechanism to remove unnecessary inbound trackers until v29 is released, which containes `MsgRemoveInboundTracker`

Once added, the trackers force the Zetaclients to keep posting votes.
The votes cannot be finalized as the corresponding cctx has already been finalized. 

This iss can be used to track the network until the new Msg can be used.

If needed, we can consider a zetaliclent-only fix to not vote on these trackers

PR #3636, opened 2025-03-05
fix: check if cctx has already been created before casting inbound vote

kingpinXD:
# Description
Closes : https://github.com/zeta-chain/node/issues/3635

The pr adds a check to stop zetaclient from voting on unnecessary ballots.

For e2e testing, its not straightforward to add a test , as we only print to logs when not casting votes.This has been tested manually by creating a tracker with the details from a confirmed CCTX and checking zetalient logs.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/base/observer.go
+++ b/zetaclient/chains/base/observer.go
@@ -406,6 +406,22 @@ func (ob *Observer) PostVoteInbound(
 		logs.FieldCoinType: coinType.String(),
 	}
 
+	cctxIndex := msg.Digest()
+
+	// The cctx is created after the inbound ballot is finalized
+	// if the cctx already exists, we should still vote if the ballot is present
+	// if the cctx exists but the ballot does not exist, we can still try voting if the ballot is present
+	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
+	if err == nil {
+		// The cctx exists we should still vote if the ballot is present
+		_, err = ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
+		if err != nil {
+			// Query for ballot failed, the ballot does not exist we can return
+			ob.logger.Inbound.Warn().Err(err).Fields(lf).Msg("inbound detected: cctx exists but no ballot")
+			return "", nil
+		}
+	}
+
 	// make sure the message is valid to avoid unnecessary retries
 	if err := msg.ValidateBasic(); err != nil {
 		ob.logger.Inbound.Warn().Err(err).Fields(lf).Msg("invalid inbound vote message")
--- a/zetaclient/chains/base/observer.go
+++ b/zetaclient/chains/base/observer.go
@@ -414,15 +414,14 @@ func (ob *Observer) PostVoteInbound(
 	// 2. if the cctx exists but the ballot does not exist, we do not need to vote
 	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
 	if err == nil {
-			// The cctx exists we should still vote if the ballot is present
-			_, err = ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
-			if err != nil {
-				// Verify ballot is not found
-				if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
-					// Query for ballot failed, the ballot does not exist we can return
-					ob.logger.Inbound.Info().Fields(lf).Msg("inbound detected: cctx exists but the ballot does not")
-					return cctxIndex, nil
-				}
+		// The cctx exists we should still vote if the ballot is present
+		_, ballotErr := ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
+		if ballotErr != nil {
+			// Verify ballot is not found
+			if st, ok := status.FromError(ballotErr); ok && st.Code() == codes.NotFound {
+				// Query for ballot failed, the ballot does not exist we can return
+				ob.logger.Inbound.Info().Fields(lf).Msg("inbound detected: cctx exists but the ballot does not")
+				return cctxIndex, nil
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


zetaclient/chains/base/observer.go
```diff
@@ -406,6 +406,22 @@ func (ob *Observer) PostVoteInbound(
...
+
+	// The cctx is created after the inbound ballot is finalized
+	// if the cctx already exists, we should still vote if the ballot is present
+	// if the cctx exists but the ballot does not exist, we can still try voting if the ballot is present
+	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
+	if err == nil {
+		// The cctx exists we should still vote if the ballot is present
+		_, err = ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
+		if err != nil {
+			// Query for ballot failed, the ballot does not exist we can return
```
kingpinXD, line 419: We can also check if this operator has voted or not ,if needed before trying to vote

zetaclient/chains/base/observer.go
```diff
@@ -406,6 +406,22 @@ func (ob *Observer) PostVoteInbound(
...
+	cctxIndex := msg.Digest()
+
+	// The cctx is created after the inbound ballot is finalized
+	// 1. if the cctx already exists, we could try voting if the ballot is present
+	// 2. if the cctx exists but the ballot does not exist, we do not need to vote
+	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
+	if err == nil {
+		// The cctx exists we should still vote if the ballot is present
+		_, err = ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
+		if err != nil {
```
ws4charlie, line 418: I think distinguishing `NotFound` from other errors (e.g. RPC failure) is better. Other errors don't mean the ballot deletion already happened, and the ballot may still exist and is open for vote?

brewmaster012:
how is this tested?

kingpinXD:
> how is this tested?

Just made a draft now and tested on localnet manually; however I'll try to add some tests before creating a final pr ( Would need to resolve the errors around unit tests first )

zetaclient/chains/base/observer.go
```diff
@@ -413,12 +414,15 @@ func (ob *Observer) PostVoteInbound(
...
 	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
 	if err == nil {
-		// The cctx exists we should still vote if the ballot is present
-		_, err = ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
-		if err != nil {
-			// Query for ballot failed, the ballot does not exist we can return
-			ob.logger.Inbound.Info().Fields(lf).Msg("inbound detected: cctx exists but the ballot does not")
-			return cctxIndex, nil
+		// Verify ballot is not found
+		if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
```
ws4charlie: shouldn't this error code check be performed on the error returned by `GetBallotByID` below?

ws4charlie (approved):
lgtm

zetaclient/chains/base/observer.go
```diff
@@ -406,6 +408,25 @@ func (ob *Observer) PostVoteInbound(
...
 
+	cctxIndex := msg.Digest()
+
+	// The cctx is created after the inbound ballot is finalized
+	// 1. if the cctx already exists, we could try voting if the ballot is present
+	// 2. if the cctx exists but the ballot does not exist, we do not need to vote
+	_, err := ob.ZetacoreClient().GetCctxByHash(ctx, cctxIndex)
+	if err == nil {
+		// The cctx exists we should still vote if the ballot is present
+		_, ballotErr := ob.ZetacoreClient().GetBallotByID(ctx, cctxIndex)
```
gartnera, line 419: Can we skip this check entirely if the cctx is already finalized? Or do we want to try and vote on all cctx for auditing purposes

zetaclient/chains/base/observer.go
```diff
@@ -406,6 +408,25 @@ func (ob *Observer) PostVoteInbound(
...
+		if ballotErr != nil {
+			// Verify ballot is not found
+			if st, ok := status.FromError(ballotErr); ok && st.Code() == codes.NotFound {
+				// Query for ballot failed, the ballot does not exist we can return
+				ob.logger.Inbound.Info().Fields(lf).Msg("inbound detected: cctx exists but the ballot does not")
+				return cctxIndex, nil
+			}
```
gartnera, line 426: This logic seems inverted? If the ballot isn't found don't we need to create it? I don't really know what I'm talking about here.

zetaclient/chains/base/observer.go
brewmaster012: cctx is created when enough votes comes in, so no m0re is needed in principle right?

brewmaster012: i'm thinking maybe just check cctx existence: 
exists -> no vote
non existing -> vote

wouldn't this be simpler?

gartnera: Yeah I think that make sense to me

kingpinXD: no we still need to allow votes if the ballot exists . The ballot gets finalized at 6/9 votes for mainnet for example , we should still accept the votes from the last 3 observers if the ballot is still present

zetaclient/chains/base/observer.go
kingpinXD: If the ballot isn't found, it means that the ballot has been deleted.
Since we first check that the cctx exists, it guarantees that the ballot was created and deleted.

if the cctx is not found it would still continue on the regular flow of voting and creating a callot

gartnera (approved):

PR #3667, opened 2025-03-06
fix: increase gas token withdraw gas limit for EVM chain

ws4charlie:
# Description

Background:

The error below has been blocking Arbitrum gas token withdraw.
image

Not all EVM chains use the same `21000` gas limit for gas token transfers, e.g. `Arbitrum` could uses `24000+` gas for ETH transfer. The PR increases the static gas limit `21000 -> 25000` to make it working for all EVM chains.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


morde08 (approved):
LGTM

PR #3676, opened 2025-03-07
feat(e2e): add --test-filter parameter

gartnera:
Now you can do this to focus on the one test you're working on rather than having to comment and uncomment code to control which tests to run.

```
export E2E_ARGS="--test-filter stress_solana_withdraw"
make start-e2e-performance-test
```

The other test routines will still be started but none of the tests will actually run.

## Summary by CodeRabbit

- **New Features**
  - Introduced a command-line option that accepts a regular expression filter, allowing users to run only tests matching specific patterns.
  - Improved test execution by integrating targeted filtering throughout the testing process for enhanced efficiency.
  - This update empowers users to precisely control which tests run, streamlining test workflows.

### DIFF
--- a/e2e/runner/e2etest.go
+++ b/e2e/runner/e2etest.go
@@ -103,6 +103,9 @@ func (r *E2ERunner) GetE2ETestsToRunByConfig(
 		if !found {
 			return nil, fmt.Errorf("e2e test %s not found", testSpec.Name)
 		}
+		if r.TestFilter != nil && !r.TestFilter.MatchString(e2eTest.Name) {
+			continue
+		}
 		e2eTestToRun := E2ETest{
 			Name:           e2eTest.Name,
 			Description:    e2eTest.Description,
--- a/cmd/zetae2e/local/local.go
+++ b/cmd/zetae2e/local/local.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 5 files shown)


gartnera:
> nice, but would be great if it can accept multiple tests, for example for withdraws to work you often need deposit tests first etc

It's a regex so something like `--test-filter 'eth_deposit|eth_withdraw'` or `--test-filter ^eth.*` should work.

e2e/runner/e2etest.go
```diff
@@ -103,6 +103,9 @@ func (r *E2ERunner) GetE2ETestsToRunByConfig(
 		if !found {
 			return nil, fmt.Errorf("e2e test %s not found", testSpec.Name)
 		}
+		if r.TestFilter != nil && !r.TestFilter.MatchString(e2eTest.Name) {
```
kingpinXD, line 106: Consider using []string for TestFilterer , to run multiple tests.
We can add filter multiple name using a regex but a comma seprated list might be easier to use

kingpinXD (commented):
This is really useful , it would be great to document the usage with examples somewhere , the README under zetae2e looks like agood place , but anywhere else you think appropriate is good too.

e2e/runner/e2etest.go
kingpinXD: Actually I think its fine , something like this seems to work 
r.TestFilter = regexp.MustCompile(`^(stress_solana_withdraw|stress_solana_deposit)$`)

Maybe just documenting the usage should be enough.

lumtis (approved):
Good idea

PR #3680, opened 2025-03-10
fix(zetaclient): evm chains: use gasLimit from CCTX

skosito:
# Description

Hardcoding evm gas limit for simple transfers doesn't work anymore due to variable arbitrum behavior.
So we can increase gas limit in arbitrum gas zrc20, and use that instead. We don't need minimum gas limit check anymore, we can just use gas limit provided in callOptions.gasLimit.

For erc20s zrc20s it is 100k already and for gas tokens 21k, with exception of arbitrum where we will bump to 100k on smart contract level.

For contracts it would be caller responsibility to put adequate value.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **Refactor**
  - Improved the management of transaction gas limits by removing static defaults and streamlining validation to focus on capping excessive values.
  
- **Bug Fixes**
  - Updated warning messages to accurately indicate when the gas limit exceeds acceptable thresholds.
  
- **Tests**
  - Revised test cases to align with the new dynamic gas limit determination.

### DIFF
--- a/zetaclient/chains/evm/signer/gas.go
+++ b/zetaclient/chains/evm/signer/gas.go
@@ -10,6 +10,7 @@ import (
 	"github.com/zeta-chain/node/x/crosschain/types"
 )
 
+// maxGasLimit is the maximum gas limit cap for EVM chain outbound to prevent excessive gas
 const maxGasLimit = 2_500_000
 
 // Gas represents gas parameters for EVM transactions.


zetaclient/chains/evm/signer/gas.go
```diff
@@ -10,10 +10,7 @@ import (
 	"github.com/zeta-chain/node/x/crosschain/types"
 )
 
-const (
-	minGasLimit = 100_000
-	maxGasLimit = 1_000_000
-)
+const maxGasLimit = 2_500_000
```
ws4charlie, line 13: would be great if you could put some comment here, for example: `maxGasLimit` is the maximum gas limit cap for EVM chain outbound to prevent excessive gas

ws4charlie (approved):
lgtm. Minor comment

PR #3694, opened 2025-03-11
fix(e2e): localnet persistence

gartnera:
- fix makefile subshell syntax. `make clear-localnet-persistence` now actually works
- add bitcoin persistence. this is needed now that we only setup the bitcoin wallet once.
- fix solana persistence. we need to prevent immediate pruning of the ledger or transactions will be pruned immediately on restart.

## Summary by CodeRabbit

- **New Features**
  - Introduced a Bitcoin service with a dedicated data persistence volume for enhanced data management.
  
- **Chores**
  - Refined local environment commands to improve the reliability of persistent volume cleanup.
  - Optimized the blockchain testing setup by adjusting the validator startup to better manage ledger size.

### DIFF
--- a/contrib/localnet/solana/start-solana.sh
+++ b/contrib/localnet/solana/start-solana.sh
@@ -6,7 +6,7 @@ solana-keygen new -o /root/.config/solana/id.json --no-bip39-passphrase
 
 solana config set --url localhost
 echo "starting solana test validator..."
-solana-test-validator &
+solana-test-validator --limit-ledger-size 50000000 &
 
 sleep 5
 # airdrop to e2e sol account
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ stop-localnet:
 
 # delete any volume ending in persist
 clear-localnet-persistence:
-	docker volume rm $(docker volume ls -qf "label=localnet=true")
+	$(DOCKER) volume rm $$($(DOCKER) volume ls -qf "label=localnet=true")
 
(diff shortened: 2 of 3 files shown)


lumtis (approved):

contrib/localnet/solana/start-solana.sh
```diff
@@ -6,7 +6,7 @@ solana-keygen new -o /root/.config/solana/id.json --no-bip39-passphrase
 
 solana config set --url localhost
 echo "starting solana test validator..."
-solana-test-validator &
+solana-test-validator --limit-ledger-size 50000000 &
```
kingpinXD, line 9: Imo this value can be lowered if needed `50000000` , `1000000` should be enough even for the upgrade tests

kingpinXD (approved):

contrib/localnet/solana/start-solana.sh
gartnera: I think this is the minimum default value on mainnet. I was thinking it might be nice to have similar behavior.

kingpinXD: okay, that works, primary reason I mentioned the smaller number is to try and make the solana container a bit lighter.

gartnera: Also it only seems to prune on restart so this is really only relevant to users utilizing the persistence feature.

You can totally just reset the persistence state whenever you want to as well.

PR #3708, opened 2025-03-13
feat(zetaclient): solana lead relayer

gartnera:
Allow a solana relayer to designate itself the lead relayer. The lead relayer will skip all preflight checks when submitting transactions and assume it's transactions will always succeed. So long as you are submitting your transactions to the same RPC server, it should be safe to submit them as fast as possible so long as they are in order. This is a simple demo for CI and feedback but it will require a few things:

- [x] lead relayer must submit transactions serially.
    - [ ] do not increment backoff counter while waiting for next transaction
    - [ ] mod nonce scheduling leads to higher nonce getting scheduled before lower nonce
- [ ] lead relayer should configure small priority fee. [ref](https://docs.helius.dev/guides/sending-transactions-on-solana#setting-the-right-priority-fee)
- [x] non-lead relayer should ideally avoid submitting the transaction the first time to allow lead relayer to take priority and avoid conflicts
    - [ ] replace static 2 second wait in current implementation. we should be able to determine if this is the first try based on observed block number and current block number.
- [ ] the lead relayer solana RPC must provide sticky load balancing and should only switch nodes on a failover
- [ ] the api call to load the pda nonce should be deduplicated
- [ ] connect a localnet to a live network to validate this actually works

[Initial performance tests](https://github.com/zeta-chain/node/actions/runs/13840342673/job/38726156880?pr=3708) show 15-20% performance gain even without these checks.

Note: when you disable preflight checks some relayer funds will be burned if your transaction fails. This is why it's important that the preflight checks are normally enabled.

With most requirements minimally satisfied, we see a 50-60% [performance increase](https://github.com/zeta-chain/node/actions/runs/13844320397/job/38739256146) which makes solana withdrawals as fast as ethereum 🎉 

```
perf_sol_wit | Latency report:
perf_sol_wit | min:  18.70
perf_sol_wit | max:  23.11
perf_sol_wit | mean: 20.42
perf_sol_wit | std:  0.95
perf_sol_wit | p50:  20.18
perf_sol_wit | p75:  20.86
perf_sol_wit | p90:  21.79
perf_sol_wit | p95:  22.44
perf_sol_wit | all SOL withdrawals completed
perf_sol_wit | ✅ completed - stress_solana_withdraw (2m29.606121s)
perf_sol_wit | 🍾 solana withdraw performance test completed in 3m1.240450272s
```

Note: we can increase the schedule lookahead because we have signature caching now so TSS won't be spammed.

Related to https://github.com/zeta-chain/node/pull/3633

### DIFF
--- a/zetaclient/chains/solana/signer/signer.go
+++ b/zetaclient/chains/solana/signer/signer.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"cosmossdk.io/errors"
@@ -59,7 +60,8 @@ type Signer struct {
 	// pda is the program derived address of the gateway program
 	pda solana.PublicKey
 
-	isLeadRelayer bool
+	isLeadRelayer     bool
+	maxProcessedNonce atomic.Uint64
 }
 
 // New Signer constructor.
@@ -287,6 +289,16 @@ func (signer *Signer) broadcastOutbound(
 		logs.FieldTx:     tx.Signatures[0].String(),
 	}
 
+	if !signer.isLeadRelayer {
+		time.Sleep(time.Second * 2)
+	}
+
+	maxProcessedNonce := signer.maxProcessedNonce.Load()
+	if signer.isLeadRelayer && nonce <= maxProcessedNonce && nonce != 0 {
+		logger.Info().Msg("lead relayer will only process transaction once")
+		return
+	}
+
 	// try broacasting tx with increasing backoff (1s, 2s, 4s, 8s, 16s, 32s, 64s)
 	// to tolerate tx nonce mismatch with PDA nonce or unknown RPC error
 	backOff := broadcastBackoff
@@ -303,8 +315,21 @@ func (signer *Signer) broadcastOutbound(
 				logger.Error().Err(err).Fields(lf).Msgf("unable to deserialize PDA info")
 			} else if pda.Nonce > nonce {
 				logger.Info().Err(err).Fields(lf).Msgf("PDA nonce %d is greater than outbound nonce, stop retrying", pda.Nonce)
+				if nonce > signer.maxProcessedNonce.Load() {
+					signer.maxProcessedNonce.Store(nonce)
+				}
 				break
 			}
+			if signer.maxProcessedNonce.Load() == 0 {
+				signer.maxProcessedNonce.Store(pda.Nonce)
+			}
+		}
+
+		maxProcessedNonce := signer.maxProcessedNonce.Load()
+		if nonce > maxProcessedNonce+1 && maxProcessedNonce != 0 {
+			// not incrementing backoff is intentional
+			logger.Info().Fields(lf).Uint64("max_processed_nonce", maxProcessedNonce).Msg("nonce > maxProcessedNonce+1")
+			continue
 		}
 
 		// broadcast the signed tx to the Solana network with preflight check
@@ -318,6 +343,9 @@ func (signer *Signer) broadcastOutbound(
 			// This reduces the number of "failed" txs due to repeated broadcast attempts.
 			rpc.TransactionOpts{PreflightCommitment: rpc.CommitmentProcessed, SkipPreflight: signer.isLeadRelayer},
 		)
+		// allow next transactions to be processed even if there was an error
+		// but we should only increment this if it's actually the new max value
+		signer.maxProcessedNonce.CompareAndSwap(nonce-1, nonce)
 		if err != nil {
 			// in case it is not failure due to nonce mismatch, replace tx with fallback tx
 			// probably need a better way to do this, but currently this is the only error to tolerate like this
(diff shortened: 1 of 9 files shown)


zetaclient/chains/solana/signer/signer.go
```diff
@@ -312,8 +341,11 @@ func (signer *Signer) broadcastOutbound(
 			// Commitment "processed" will simulate tx against more recent state
 			// thus fails faster once a tx is already broadcasted and processed by the cluster.
 			// This reduces the number of "failed" txs due to repeated broadcast attempts.
-			rpc.TransactionOpts{PreflightCommitment: rpc.CommitmentProcessed},
+			rpc.TransactionOpts{PreflightCommitment: rpc.CommitmentProcessed, SkipPreflight: signer.isLeadRelayer},
```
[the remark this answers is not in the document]
gartnera: > i know its in draft mode, but just wanted to suggest to maybe try to call simulate rpc endpoint manually if skipping preflight checks, curious to see improvements, and that way we can catch failing tx during simulation, i read it should be quicker than preflight

https://github.com/zeta-chain/node/pull/3633 (polling the pda nonce waiting for it to be in range) seems to accomplish approximately the same thing. I think that would be slightly better too since you could just do it once for N transactions.

> on maxRetries tx opts, i found multiple opinions, some suggest it to not set it, and retry manually, which we basically doing with zetaclient, but i also saw a post somewhere setting maxRetries to 1 or 2, so one more thing to test i guess

>> Maximum number of times for the RPC node to retry sending the transaction to the leader. If this parameter is not provided, the RPC node will retry the transaction until it is finalized or until the blockhash expires.

That's interesting so in that case the RPC node would actually be the one retrying the transaction. I think it might make sense to set to 1-2 on the lead relayer. I wonder if the RPC is held open while it is being retried or if that only happens in the background?

I would assume if the preflight check fails this retry would probably be ignored anyway.

zetaclient/chains/solana/solana.go
```diff
@@ -179,7 +179,7 @@ func (s *Solana) scheduleCCTX(ctx context.Context) error {
 		// 1. schedule the very first cctx (there can be multiple) created in the last Zeta block.
 		// 2. schedule new cctx only when there is no other older cctx to process
 		isCCTXNewlyCreated := inboundParams.ObservedExternalHeight == zetaHeight
-		shouldProcessCCTXImmedately := isCCTXNewlyCreated && needsProcessingCtr == 0
+		shouldProcessCCTXImmedately := isCCTXNewlyCreated
```
ws4charlie, line 182: is `needsProcessingCtr` read nowhere? should remove?

zetaclient/chains/solana/signer/signer.go
```diff
@@ -299,8 +315,21 @@ func (signer *Signer) broadcastOutbound(
 				logger.Error().Err(err).Fields(lf).Msgf("unable to deserialize PDA info")
 			} else if pda.Nonce > nonce {
 				logger.Info().Err(err).Fields(lf).Msgf("PDA nonce %d is greater than outbound nonce, stop retrying", pda.Nonce)
+				if nonce > signer.maxProcessedNonce.Load() {
```
ws4charlie, line 318: The name `maxProcessedNonce` is unclear to me. Does it mean the `max nonce processed only by me` or `the gateway pending nonce`, similar to Ethereum pending nonce. I think giving a consistent name helps understanding, Bitcoin chain also maintains its own pending nonce in zetaclient.

zetaclient/chains/solana/signer/signer.go
```diff
@@ -299,8 +315,21 @@ func (signer *Signer) broadcastOutbound(
...
+					signer.maxProcessedNonce.Store(nonce)
+				}
 				break
 			}
+			if signer.maxProcessedNonce.Load() == 0 {
+				signer.maxProcessedNonce.Store(pda.Nonce)
```
ws4charlie, line 324: `pda` is a invalid value when program execution comes here. Ideally, it would be good if we have `maxProcessedNonce` updating logic in one single place (including initialization and update), like

```
if nonce >= signer.maxProcessedNonce.Load() {
	signer.maxProcessedNonce.Store(nonce)
}
```

zetaclient/chains/solana/signer/signer.go
```diff
@@ -299,8 +315,21 @@ func (signer *Signer) broadcastOutbound(
...
+			}
+		}
+
+		maxProcessedNonce := signer.maxProcessedNonce.Load()
+		if nonce > maxProcessedNonce+1 && maxProcessedNonce != 0 {
```
ws4charlie, line 329: If the increasing backoff is not what you want, you could try `retry.DoWithBackoff` with constant backoff parameter, like [here](https://github.com/zeta-chain/node/blob/solana-lead-relayer/zetaclient/chains/bitcoin/signer/signer.go#L159). Not sure what this refinement is intended for, would be good to put a piece description for easy understanding, same for above ones.

zetaclient/chains/solana/signer/signer.go
gartnera: > max nonce processed only by me

yes that is want I mean

I should call it `maxRelayedNonce`

ws4charlie:
This PR will be simplified and replaced by https://github.com/zeta-chain/node/pull/3940
@skosito @lumtis

PR #3717, opened 2025-03-14
fix: fix solana withdraw and call outbounds

skosito:
# Description

Every zetaclient is doing key sign and  then relayer sign, even though they might not have relayer key. This was causing panics from the start of solana outbounds, but not causing outbounds to fail because panics were recovered and relayer broadcasted it.

But now we might have 2 key signs, tx and fallback tx, so if there is panic after key sign for tx, fallback tx keysign will always fail.

This PR is simplest fix for this, for next milestone we can refactor it to be a bit more cleaner and less code duplication, but just skipping relayer sign should be enough if key is not set up.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 
 ### Fixes
 * [3711](https://github.com/zeta-chain/node/pull/3711) - fix TON call_data parsing
+* [3717](https://github.com/zeta-chain/node/pull/3717) - fix solana withdraw and call panic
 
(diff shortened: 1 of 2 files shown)


gartnera (approved):

ws4charlie (approved):
LGTM

Issue #3723, opened 2025-03-17
bug: a low gas limit in ERC20 revert outbound will block whole outbound


ws4charlie:
**Describe the Bug**
A recent `zetaclientd` hotfix removed the `100K` minimum gas limit check and the gas limit field in the CCTX was used instead. This caused a situation when a `ERC20` revert outbound has a low gas limit `21000` and blocks outbound due to the error `intrinsic gas too low`.

The blocked CCTX in BASE was [this one](https://zetachain.blockpi.network/lcd/v1/public/zeta-chain/crosschain/cctx/8453/25785).

![Image](https://github.com/user-attachments/assets/494ae804-aa5b-4a68-9fab-0619a19038d7)

To unblock the outbound without `zetacore` change, we added back the `100K` gas limit check only for ERC20 token in a [zetaclient-only hotfix](https://github.com/zeta-chain/node/pull/3719/files).

**Expected Behavior**

I think a proper gas limit should be used for ERC20 revert outbound in order to both charge correct gas fee from end user and avoid blocked outbound.

![Image](https://github.com/user-attachments/assets/98c155fa-5b36-41a2-b984-dcfed4773dd0)

lumtis:
There are two issues:

- If a revert happens, either gas limit in revertOptions should be used, otherwise the default gas limit of the ZRC20 (should have been 100000 for this cctx)
- Outbounds shouldn't be blocked because of low gas limit, if this happens we should have a mechanism to revert the transaction

This issue looks more outlining the first problem, I will create an issue for the second one

CharlieMc0:
Can this be closed? @ws4charlie

kingpinXD:
We should, overwrite the limit if the one provided is too low when updating the revertOptions for the cctx

lumtis:
The initial issue with the wrong gas limit set was fixed, [Outbound with gas limit too low should not block outbounds #3725](https://github.com/zeta-chain/node/issues/3725) is the followup

PR #3726, opened 2025-03-17
fix: add min gas limit for non-gas outbound transactions (develop rebase)

lumtis:
# Description

Add to upstream backported fix: https://github.com/zeta-chain/node/pull/3719 

Add a TODO to address the underlying issue

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction reliability by auto-adjusting gas limits when they fall below a minimum threshold. This change ensures smoother transaction processing and provides a warning when the gas limit is too low.

### DIFF
--- a/zetaclient/chains/evm/signer/gas.go
+++ b/zetaclient/chains/evm/signer/gas.go
@@ -15,7 +15,7 @@ import (
 const maxGasLimit = 2_500_000
 
 const gasTransferGasLimit = 21_000
-const nonGasMinGasLimit = 100_000
+const contractCallMinGasLimit = 100_000
 
 // Gas represents gas parameters for EVM transactions.
 //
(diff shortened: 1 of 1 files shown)


zetaclient/chains/evm/signer/gas.go
```diff
@@ -73,6 +77,17 @@ func gasFromCCTX(cctx *types.CrossChainTx, logger zerolog.Logger) (Gas, error) {
 			Uint64("cctx.initial_gas_limit", params.CallOptions.GasLimit).
 			Uint64("cctx.gas_limit", limit).
 			Msgf("Gas limit is too high; Setting to the maximum (%d)", maxGasLimit)
+	} else if limit == gasTransferGasLimit && (params.CoinType != coin.CoinType_Gas || (cctx.IsCurrentOutboundRevert() && cctx.RevertOptions.CallOnRevert)) {
```
kingpinXD, line 80: Just wondering who would be paying for this increased limit, as we would have only charged the user `gasTransferGasLimit` when creating the revert outbound?

lumtis: Yes, there is an underpayment until we address https://github.com/zeta-chain/node/issues/3723 but it has been the actual behavior since the beginning until we patching with the cctx data.

Right now it is a bit mitigated in the sense it strictly check for 21k, so the user will still have to pay a fee but we need to complete this before adding the refund PR

PR #3756, opened 2025-03-25
feat: solana revert options

skosito:
# Description

solana PR: https://github.com/zeta-chain/protocol-contracts-solana/pull/95

This PR is mainly about parsing revert options, actual usage needs execute revert functions in solana gateway, and i will add more unit tests when that is ready, since there will be real inbounds and outbounds in localnet i can use in tests.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced configurable revert options for Solana transactions, enhancing the robustness and clarity of transaction handling.
  - Extended support for SOL-specific revert options in cross-chain vote messaging, allowing for detailed control over transaction reversion scenarios.


x/crosschain/types/revert_options.go
```diff
@@ -49,6 +50,21 @@ func NewRevertOptionsFromEVM(revertOptions gatewayevm.RevertOptions) RevertOptio
 	}
 }
 
+// NewRevertOptionsFromSOL initializes a new RevertOptions from a solana.RevertOptions
+func NewRevertOptionsFromSOL(revertOptions solana.RevertOptions) RevertOptions {
```
lumtis, line 54: As per the Solana pr, I suggest to add onAbort address parsing here.

By adding it, the onAbort workflow should be automatically supported for Solana

lumtis (commented):
- Let's add a note in the changelogs this is a breaking change for Solana deposit, and the gateway must be upgrade after node upgrade
- Can we modify the depositAndCall revert E2E test to revert funds (no onRever call) to a specifc address?

lumtis (approved):

x/crosschain/types/cctx.go
```diff
@@ -153,12 +153,16 @@ func (m *CrossChainTx) AddRevertOutbound(gasLimit uint64) error {
...
 	if m.ProtocolContractVersion == ProtocolContractVersion_V2 {
 		if chains.IsBitcoinChain(m.InboundParams.SenderChainId, []chains.Chain{}) {
 			if m.RevertOptions.RevertAddress != "" {
 				revertReceiver = m.RevertOptions.RevertAddress
 			}
+		} else if chains.IsSolanaChain(m.InboundParams.SenderChainId, []chains.Chain{}) {
+			revertAddress, valid := m.RevertOptions.GetSOLRevertAddress()
+			if valid {
+				revertReceiver = revertAddress.String()
+			}
```
kingpinXD, line 165: Consider refactoring to use a switch statement

x/crosschain/types/revert_options.go
```diff
@@ -76,6 +94,13 @@ func (r RevertOptions) GetEVMRevertAddress() (ethcommon.Address, bool) {
 	return addr, !crypto.IsEmptyAddress(addr)
 }
 
+// GetSOLRevertAddress returns the SOL revert address
+// if the revert address is not a valid address, it returns false
+func (r RevertOptions) GetSOLRevertAddress() (solana.PublicKey, bool) {
```
kingpinXD, line 99: We have the same pattern for other chains, but we should consider propagating this to the user. 
If the revert ends up on the sender address instead of the revert address, it might not be super clear for a dApp user. 
Wdyt? Maybe we should set this error into `error_revert,` which would be overwritten if the revert fails, but if the revert is successful, it would still be available

x/crosschain/types/cctx.go
```diff
@@ -153,12 +153,16 @@ func (m *CrossChainTx) AddRevertOutbound(gasLimit uint64) error {
 
 	// in protocol contract V2, developers can specify a specific address to receive the revert
 	// if not specified, the sender address is used
-	// note: this option is current only support for EVM type chains
 	if m.ProtocolContractVersion == ProtocolContractVersion_V2 {
 		if chains.IsBitcoinChain(m.InboundParams.SenderChainId, []chains.Chain{}) {
 			if m.RevertOptions.RevertAddress != "" {
 				revertReceiver = m.RevertOptions.RevertAddress
 			}
+		} else if chains.IsSolanaChain(m.InboundParams.SenderChainId, []chains.Chain{}) {
```
kingpinXD, line 161: we can add a test case under `Test_SetRevertOutboundValues` maybe

e2e/e2etests/test_solana_deposit_and_call_revert.go
```diff
@@ -24,12 +27,26 @@ func TestSolanaDepositAndCallRevert(r *runner.E2ERunner, args []string) {
 
 	// execute the deposit transaction
 	data := []byte("hello reverter")
-	sig := r.SOLDepositAndCall(nil, reverterAddr, depositAmount, data)
+	revertAddressPrivateKey, err := solana.NewRandomPrivateKey()
```
kingpinXD, line 30: NIT : maybe a revert option with incorrect address would be easy to add . 
Up to you , its more like good to have and not needed

kingpinXD (approved):

Issue #3755, opened 2025-03-25
Add a function to check if a Sui object is shared or not


lumtis:
One of the core principle of Sui withdraw and call is to allow only to provide shared object in `on_call` blockchain call.

This check is crititcal to prevent a user for withdrawing the withdrawCap.

A function shouldallow to run this check.

It can the be used in the withdrawAndCall signature. If one object is not shared, the withdrawAndCall should be reverted.

PR #3766, opened 2025-03-26
feat: check withdrawAndCall objects are shared

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/3755

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced validations for Sui interactions that ensure only objects with valid shared or immutable ownership are processed, reducing potential errors.
- **Tests**
  - Expanded test coverage now validates various ownership scenarios—including empty, valid shared, mixed, and invalid cases—to improve overall reliability.


zetaclient/chains/sui/client/client.go
```diff
@@ -245,3 +249,52 @@ func parseRPCResponse[T any](raw []byte) (T, error) {
...
+
+	// should always be the case, we add this check as a extra safety measure to ensure an object is not skipped
+	if len(res) != len(objectIDs) {
+		return fmt.Errorf("expected %d objects, but got %d", len(objectIDs), len(res))
+	}
+
+	return containsOwnedObject(res)
+}
+
+func containsOwnedObject(res []*models.SuiObjectResponse) error {
```
ws4charlie, line 277: Returning a `bool` is more natural for function name like `containsOwnedObject`. Is `ensureNoOwnedObjects` or `ensureSharedOrImmutableObjects` a more accurate name if it returns an error?

swift1337 (approved):

ws4charlie (approved):
looks good overall.

zetaclient/chains/sui/client/client.go
swift1337: IMO the error is better because that way, we know the root cause with more details.

ws4charlie: Didn't figure out the purpose of the function from its name as the name doesn't match what it does. lol

lumtis: Renamed with the check prefix terminology to amplify it's not just returning a boolean

PR #3770, opened 2025-03-27
fix: improve fallback tx error handling

skosito:
# Description

extended solana examples a bit to enable e2e test https://github.com/zeta-chain/protocol-contracts-solana/pull/97

idea is to check error logs for `Program <PROGRAM_ID> invoked` logs, and to check if some program is invoked after gateway

- if only gateway is invoked, we just skip NonceMismatch errors, anything else is requires fallback (eg: token transfer, regular transfer, etc)
- if something else is invoked after gateway, fallback is needed, error msg from that program is not considered, so it can be anything, including NonceMismatch - e2e test is extended with this case
- if connected program calls back gateway, it is reentrancy and it will fail, but gateway invoke might appear again, so this should cover that scenario as well

we probably should look into other solana golang libraries, this one just gives error string so we must do some parsing to figure out on our own

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **Documentation**
  - Updated the changelog to reflect improved fallback transaction error handling.

- **Bug Fixes**
  - Enhanced the error handling in transactions to trigger fallback processing under additional conditions for improved reliability.

- **Tests**
  - Refined test conditions to validate more specific revert triggers.
  - Added tests to verify accurate detection of error conditions based on invocation sequences.


pkg/contracts/solana/instruction.go
```diff
@@ -582,3 +583,30 @@ func DecodeExecuteMsg(msgbz []byte) (ExecuteMsg, error) {
 
 	return msg, nil
 }
+
+// ProgramInvokedAfterTargetInErrStr checks if any program (other than the specified one)
+// was invoked after the given program in the error string logs.
+func ProgramInvokedAfterTargetInErrStr(errStr, targetProgram string) bool {
```
gartnera, line 589: Rather than pure string matching maybe we could type cast it to `*jsonrpc.RPCError` and look at the `Data["logs"]` attribute?

pkg/contracts/solana/instruction_test.go
```diff
@@ -78,3 +78,120 @@ func Test_RecoverSigner(t *testing.T) {
...
+			 (string) (len=10) "returnData": (interface {}) <nil>,
+			 (string) (len=13) "unitsConsumed": (json.Number) (len=5) "13987"
+			}
+		   })`
+
+		invoked := contracts.ProgramInvokedAfterTargetInErrStr(errStr, "94U5AHQMKkV5txNJ17QPXWoh474PheGou6cNP2FEuL1d")
+		require.False(t, invoked)
+	})
+
+	t.Run("program invoked after gateway", func(t *testing.T) {
```
ws4charlie, line 118:
```suggestion
	t.Run("target program invoked after gateway", func(t *testing.T) {
```

pkg/contracts/solana/instruction.go
ws4charlie: want to confirm: the gateway will be passed to `targetProgram` and we're not checking receiver program in here?

gartnera: https://github.com/gagliardetto/solana-go/blob/main/rpc/jsonrpc/jsonrpc.go#L214 is the definitive type of the error returned. You can see it has the structure I mentioned.

lumtis (approved):
Wondering, do we have a case where we produce a NonceMismatch from signers in E2E tests?

Is it something that could be reproduced?

pkg/contracts/solana/instruction.go
gartnera: There will probably have to still be some string parsing but it would be nice to be as targeted as possible.

ws4charlie (approved):
looks good

gartnera (approved):

ws4charlie (approved):

PR #3777, opened 2025-03-28
refactor: use SignBatch keysign for solana outbound tx and fallback tx

skosito:
# Description

Instead of having 2 consecutive key signs for solana tx and fallback tx, use SignBatch. 

Since this will be needed for execute revert and so on, this should make it a bit less boilerplate in signer code.

Also, locally it seems there is slight performance improvement (~2s less for e2e test solana_withdraw_and_call and ~1 for spl_withdraw_and_call), and ~15s for overall test set, but this needs more measurement, it is just based on couple of runs.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Enabled batch processing for signing Solana outbound transactions, allowing both primary and fallback flows to be handled concurrently.

- **Refactor**
  - Streamlined transaction handling for Solana operations, consolidating processes for both standard and SPL transactions to enhance execution reliability and efficiency.


lumtis (approved):

ws4charlie (approved):
looking good

Issue #3779, opened 2025-03-31
Athens3 testnet halted


gartnera:
```
panic: more validators than maxValidators found

goroutine 1 [running]:
github.com/cosmos/cosmos-sdk/x/staking/keeper.Keeper.GetLastValidators({{0x4b20c00, 0xc001daa058}, {0x4b86170, 0xc001d075a0}, {0x4b6f9f0, 0xc000242140}, {0x7fc414629708, 0xc000ba3408}, {0x4b85e50, 0xc002290570}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.12/x/staking/keeper/validator.go:431 +0x658
github.com/cosmos/cosmos-sdk/x/staking/keeper.Keeper.TrackHistoricalInfo({{0x4b20c00, 0xc001daa058}, {0x4b86170, 0xc001d075a0}, {0x4b6f9f0, 0xc000242140}, {0x7fc414629708, 0xc000ba3408}, {0x4b85e50, 0xc002290570}, ...}, ...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.12/x/staking/keeper/historical_info.go:120 +0x348
github.com/cosmos/cosmos-sdk/x/staking/keeper.(*Keeper).BeginBlocker(0x378?, {0x4b647e8?, 0xc0006f2008?})
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.12/x/staking/keeper/abci.go:16 +0x13e
github.com/cosmos/cosmos-sdk/x/staking.AppModule.BeginBlock(...)
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.12/x/staking/module.go:181
github.com/cosmos/cosmos-sdk/types/module.(*Manager).BeginBlock(_, {{0x4b646d0, 0x70c8ec0}, {0x4b86a30, 0xc002675040}, {{0x0, 0x0}, {0xc001d300f0, 0xd}, 0x939c86, ...}, ...})
        /root/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.12/types/module/module.go:778 +0x188
github.com/zeta-chain/node/app.(*App).BeginBlocker(...)
        /go/src/github.com/zeta-chain/node/app/app.go:867
github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).beginBlock(0xc0016fbb08, 0xc000f12900?)
[...]

PR #3780, opened 2025-03-31
fix(v29): force max val number

lumtis:
Related to https://github.com/zeta-chain/node/issues/3779

This should not be backported to develop (imo @gartnera)


x/crosschain/module.go
```diff
@@ -179,6 +179,24 @@ func (am AppModule) BeginBlock(c context.Context) error {
 	// error is logged in the function
 	am.keeper.IterateAndUpdateCctxGasPrice(ctx, supportedChains, keeper.CheckAndUpdateCctxGasPrice)
 
+	// PATCH - v29
+	// check the staking param and force 100 for maxValidator
+	stakingParams, err := am.keeper.GetStakingKeeper().GetParams(ctx)
+	if err != nil {
+		// avoid returning error here to halt the chain
+		am.keeper.Logger(ctx).Error("failed to get staking params", "error", err)
+	} else {
```
kingpinXD, line 188: NIT: consider returning early instead of an `if` `else` logic.

kingpinXD (approved):
lgtm , minor comments

gartnera (approved):
Tested on my personal athens rpc node:

```
7:29PM INF Replay last block using real app module=consensus server=node
7:29PM INF max validator different than 100 maxValidators=25 module=x/crosschain
7:29PM INF staking params updated maxValidators=100 module=x/crosschain
```

PR #3795, opened 2025-04-09
refactor: add solana program with re-organished code

kingpinXD:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


kingpinXD:
> wondering about gateway.so reduction in size, is it reduced by 86% or (-56.4 KB (86%)? but in any case do you know where reduction is coming from? solana e2e tests passing meaning all should be good

I think it might be due to the toolchain config , I think we might have been using a different version of rust before , its not fixed to 1.81.0

lumtis (approved):

Issue #3798, opened 2025-04-14
Add Sui withdraw address validation


lumtis:
**Describe the Issue**
Add validation for Sui address at this location: https://github.com/zeta-chain/node/blob/370b1d55a0a5c3db6eeb52d2f8ab68cdace1abc0/x/crosschain/keeper/evm_hooks.go#L359

PR #3802, opened 2025-04-14
fix: prevent withdraw with invalid Sui address

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/3798

Longer term solution for https://github.com/zeta-chain/node/pull/3801

Prevent the withdraw from the smart contract function directly

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented withdrawal attempts using invalid Sui addresses.
  - Improved error messages for withdrawal operations, providing clearer feedback when an unsupported address is entered across multiple chains.


ws4charlie (approved):
looks good

PR #3820, opened 2025-04-22
fix: set retry gas limit if outbound is successful

skosito:
# Description

Currently if outbound is successful retry gas limit will be 0, meaning there won't be retry for voting if 500k is not enough.
This introduces retry gas limit 1M for all observers.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


lumtis (approved):
@skosito let's create a PR for `develop` as well

kingpinXD (approved):

PR #3823, opened 2025-04-24
fix: use increment nonce if prepareExecuteTx fails

skosito:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/solana/signer/signer.go
+++ b/zetaclient/chains/solana/signer/signer.go
@@ -152,19 +152,20 @@ func (signer *Signer) TryProcessOutbound(
 
 	case coin.CoinType_Gas:
 		if cctx.IsWithdrawAndCall() {
-			executeTxGetter, err := signer.prepareExecuteTx(ctx, cctx, height, logger)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


zetaclient/chains/solana/signer/signer.go
```diff
@@ -152,19 +152,20 @@ func (signer *Signer) TryProcessOutbound(
...
 			if err != nil {
 				logger.Error().Err(err).Msgf("TryProcessOutbound: Fail to sign increment_nonce outbound")
 				return
 			}
-
-			txGetter = executeTxGetter
 			fallbackTxGetter = incrementNonceTxGetter
+
+			executeTxGetter, err := signer.prepareExecuteTx(ctx, cctx, height, logger)
+			if err != nil {
```
kingpinXD, line 163: Wondering if we should check the error, 
Fro example ,  If the `prepareExecuteTx` fails at 
```
	message, err := hex.DecodeString(cctx.RelayedMessage)
	if err != nil {
		return nil, errors.Wrapf(err, "decodeString %s error", cctx.RelayedMessage)
	}
```

Would we continue with the same logic, for `err!=nil`?

PR #3836, opened 2025-04-30
chore: versions docs file

fadeev:
## Summary by CodeRabbit

- **New Features**
  - Introduced a centralized version tracking file displaying version, commit, and timestamp information for multiple components.


versions.json
```diff
@@ -0,0 +1,29 @@
+[
+  {
+    "node": {
```
swift1337, line 3: What's that? Why do we need this file?

fadeev: We need a table that answers the question "which version of EVM/Sui/Solana/etc. protocol contracts are running with this version of the node". So that when the node upgrades on testnet to version X, we need to know to which versions of protocol contracts localnet/toolkit/etc. need to be updated.

fadeev: We can have an actual table, but this is easier to fetch programmatically and render in the docs (compared to markdown).

swift1337: Let's keep this out of the repo. I see these issues:
- It's not clear why node repo needs to keep track of all other repos that we have
- It's not clear who's responsible for keeping this up to date
- It's not clear how to automate this

fadeev: We can just import all protocol contracts repos in go.mod (like we are doing right now [with EVM protocol contracts](https://github.com/zeta-chain/node/blob/4c51775d71c8e313303dc4db125ac57acf793aad/go.mod#L53)) and keep track of it this way. We are not using Solana/Sui/TON contracts directly in the repo though.

fadeev: @swift1337 where would you store this info? Markdown table in the docs?

lumtis: I think a MD file in the node repo would work, the table might grow so I would keep it out of the readme.md, just provide a link to it.

lumtis: >It's not clear why node repo needs to keep track of all other repos that we have

node repo is the central point of the protocol architecture (renaming it to zetachain has been considered but involves to many implications regarding broken links) and determine depedencies with all protocol contracts repos so this looks the right place to me to store this table

> It's not clear who's responsible for keeping this up to date

> It's not clear how to automate this

I don't think we need to automatize it, should be added in playbook and become a best practice to update it as sone as we make a new protocol release.

fadeev:
Replaced JSON with markdown. Since it's a regular table I think it's easier for us to just write down git tags (versions) without commits, timestamps, etc. Knowing a tag is enough to get all other info programmatically.

fadeev:
We still don't have Sui version, though. We need to make a release.

lumtis (commented):
Can we add a link to this doc in this section? https://github.com/zeta-chain/node/blob/develop/readme.md#protocol-contracts

lumtis:
> We still don't have Sui version, though. We need to make a release.

Created v1.0.0 https://github.com/zeta-chain/protocol-contracts-sui/releases/tag/v1.0.0

versions.md
```diff
@@ -0,0 +1,3 @@
+| Node    | EVM     | Solana | Sui | TON        |
+| ------- | ------- | ------ | --- | ---------- |
+| v30.0.0 | v13.0.0 | v3.0.1 |     | v1.0.0-rc3 |
```
lumtis, line 3:
```suggestion
| v30.0.0 | v13.0.0 | v3.0.1 |  v1.0.0 | v1.0.0-rc3 |
| v31.0.0 | v13.0.0 | v5.0.0 |  v1.0.0 | v1.0.0-rc3 |
```

lumtis (approved):

swift1337 (approved):

Issue #3822, opened 2025-04-23
Parse events from inner instructions for solana inbounds


skosito:
Currently inbound event parsing for solana just consider direct gateway calls. But if contract calls gateway, that call will be in inner instructions (found in txResult.Meta).

Purpose of this is to research if inner instructions can be parsed and to modify filter event method to support this, or to document if it's not possible.

CharlieMc0:
@skosito can these tx be rejected at gateway contract level to prevent people from sending unsupported txs?

PR #3839, opened 2025-04-30
feat: parse solana inbounds from inner instructions

skosito:
# Description

- extend FilterInboundEvents to consider inner instructions for parsing inbound events
- use connected example program to call gateway deposit (https://github.com/zeta-chain/protocol-contracts-solana/pull/105)
- add e2e test
- add and update inbound event tests

closes: https://github.com/zeta-chain/node/issues/3822

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Added support for parsing Solana inbound transactions from inner instructions, improving detection of inbound events.
  - Introduced a new end-to-end test for depositing SOL through a connected Solana program.
  - Added a new instruction to the Solana gateway contract for resetting the PDA nonce.

- **Bug Fixes**
  - Enhanced inbound event filtering to include inner instructions for more comprehensive event detection.

- **Tests**
  - Expanded test coverage for parsing inbound events from inner instructions.
  - Added new test data for Solana inbound transaction scenarios.

- **Documentation**
  - Updated the changelog to reflect new Solana inbound transaction parsing capabilities.

- **Refactor**
  - Improved internal structure and error handling in Solana inbound event parsing and related test code.

### DIFF
--- a/zetaclient/chains/solana/observer/inbound_parser.go
+++ b/zetaclient/chains/solana/observer/inbound_parser.go
@@ -26,6 +26,7 @@ type InboundEventParser struct {
 	seenDepositSPL bool
 	seenCall       bool
 	events         []*clienttypes.InboundEvent
+	eventIndex     uint32
 }
 
 // NewInboundEventParser creates a new InboundEventParser
@@ -47,15 +48,12 @@ func NewInboundEventParser(
 		tx:            tx,
 		txResult:      txResult,
 		events:        make([]*clienttypes.InboundEvent, 0),
+		eventIndex:    0,
 	}, nil
 }
 
 // parseInstruction parses a single instruction and adds any detected events
-func (p *InboundEventParser) parseInstruction(
-	instruction solana.CompiledInstruction,
-	location string,
-	instructionIndex uint32,
-) error {
+func (p *InboundEventParser) parseInstruction(instruction solana.CompiledInstruction, location string) error {
 	// get the program ID
 	programPk, err := p.tx.Message.Program(instruction.ProgramIDIndex)
 	if err != nil {
@@ -90,17 +88,18 @@ func (p *InboundEventParser) parseInstruction(
 				Memo:             deposit.Memo,
 				BlockNumber:      deposit.Slot,
 				TxHash:           p.tx.Signatures[0].String(),
-				Index:            instructionIndex,
+				Index:            p.eventIndex,
 				CoinType:         coin.CoinType_Gas,
 				Asset:            deposit.Asset,
 				IsCrossChainCall: deposit.IsCrossChainCall,
 				RevertOptions:    deposit.RevertOptions,
 			})
+			p.eventIndex++
 			p.logger.Info().
 				Str(logs.FieldMethod, "parseInstruction").
 				Str("signature", p.tx.Signatures[0].String()).
 				Str("location", location).
-				Uint32("instructionIndex", instructionIndex).
+				Uint32("eventIndex", p.eventIndex-1).
 				Msg("deposit detected")
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


zetaclient/chains/solana/observer/inbound.go
```diff
@@ -167,30 +168,32 @@ func FilterInboundEvents(
 	seenCall := false
 	events := make([]*clienttypes.InboundEvent, 0)
 
-	// loop through instruction list to filter the 1st valid event
-	for i, instruction := range tx.Message.Instructions {
+	// helper to parse any instruction
+	parseInstruction := func(instruction solana.CompiledInstruction, location string) error {
```
kingpinXD, line 172: Do you think it makes sense to return a list of events here and move this out to a spearate function ? It would improve the testability

zetaclient/chains/solana/observer/inbound.go
```diff
@@ -281,16 +288,39 @@ func FilterInboundEvents(
...
+		return nil
+	}
+
+	// parse top-level instructions
+	for i, instruction := range tx.Message.Instructions {
+		err := parseInstruction(instruction, fmt.Sprintf("top-level instruction %d", i))
+		if err != nil {
+			return nil, err
+		}
+	}
```
kingpinXD, line 312: if an outer instruction is found , do we need to scan for inner after that ? 

I guess only one instruction can be processed per TX , cause the second one would fail at `IsFinalizedInbound` check. Not 100% sure , maybe we can test this ?

zetaclient/chains/solana/observer/inbound.go
ws4charlie: Looks like a good point, the function now spans 170+ lines. If we could create a separate func
`func parseInstructions(instructions []solana.CompiledInstruction) ([]*clienttypes.InboundEvent, error)`.

The whole input will be: `tx.Message.Instructions + inner.Instructions` and the flags `seenDeposit, seenDepositSPL and seenCall` can live inside `parseInstructions`. Not sure how this will impact your unit tests?

ws4charlie (approved):
Looks good. minor comment.

zetaclient/chains/solana/observer/inbound.go
ws4charlie: Yeah. The original idea was to keep the logic consistent with EVM chains, one gas/erc20 deposit per tx, so here is one gas/spl deposit tx. The cap on num of events was to prevent spamming. I think keep one for each event type is okay.

e2e/runner/solana.go
```diff
@@ -75,6 +75,36 @@ func (r *E2ERunner) CreateDepositInstruction(
...
 }
 
+// CreateDepositInstructionThroughProgram creates a 'trigger_deposit' instruction for connected example program
+func (r *E2ERunner) CreateDepositInstructionThroughProgram(
+	signer solana.PublicKey,
+	receiver ethcommon.Address,
+	amount uint64,
+	revertOptions *solanacontract.RevertOptions,
+) solana.Instruction {
+	triggerDepositDiscriminator := [8]byte{154, 34, 24, 72, 18, 230, 27, 82}
```
swift1337, line 85: Can we use a string here that is wrapped to some SOL util call that converts it to the same `[8]byte{}`?

zetaclient/chains/solana/observer/inbound.go
kingpinXD: The event index is set to 0 irrespective of the type, so even if we vote on multiple events, only the first one would be processed, right?

The key for IsFinalized being `chainID, inboundHash, eventIndex` 
So we would have multiple msg.Digest , but same `IsFinalizedKey`

ws4charlie: That's a good point. We probably have to use actual instruction index inside the transaction as the event index.

lumtis (commented):
Asking as parsing is different from EVM, do we make sure we skip the observation doesn't happen for a deposit through program that fails?

zetaclient/chains/solana/observer/inbound_parser.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+
+	// parse inner instructions
+	if p.txResult.Meta != nil && p.txResult.Meta.InnerInstructions != nil {
+		for _, inner := range p.txResult.Meta.InnerInstructions {
+			for j, instruction := range inner.Instructions {
+				// For inner instructions, we use the outer instruction index as the base
+				// and add the inner instruction index to ensure uniqueness
+				instructionIndex := uint32(inner.Index) + uint32(j)
```
ws4charlie, line 216: Nothing but a question: will this `instructionIndex` be conflicting with above `uint32(i))` in the top-level instruction loop?

ws4charlie (approved):
looks good. Left a question to confirm my understanding

lumtis (approved):

PR #3882, opened 2025-05-16
feat: support setting default config.toml values

skosito:
# Description

This should enable us to reduce block time in code, without coordinating with all nodes, it will override config.toml timeout_commit field.
Can be skipped in e2e tests or localnet with a flag.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Added automatic override of the consensus `TimeoutCommit` setting to 5 seconds in the configuration file when starting the node, improving default behavior.
  - Introduced a new `--skip-config-override` flag to allow users to opt out of automatic configuration overrides during node startup.

- **Chores**
  - Updated documentation to reflect the new configuration override feature.


lumtis (approved):
LGTM

swift1337 (approved):

lumtis:
@skosito need to run `make generate`?

kingpinXD (approved):

kingpinXD:
> @lumtis @kingpinXD i fixed upgrade tests like this [97a0178](https://github.com/zeta-chain/node/commit/97a017839db2280b58766da7cc506d5641ec7ecc)
> 
> old binary doesn't have this old flag, i think this should be ok we can remove the check after new release, please let me know if we did things like this differently

lgtm

PR #3904, opened 2025-05-22
fix: improve observer emissions distribution

kingpinXD:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


x/observer/types/ballot.go
```diff
@@ -93,37 +93,26 @@ func CreateVotes(listSize int) []VoteType {
 
 // BuildRewardsDistribution builds the rewards distribution map for the ballot
 // It returns the total rewards units which account for the observer block rewards
```
lumtis, line 95: Outdated, but I would consider still intializing the map in this function and returning it.
Right now we pass an empty map as an argument which is odd

x/emissions/abci.go
```diff
@@ -185,31 +185,48 @@ func distributeRewardsForMaturedBallots(
...
+
+		// After calculating the rewardsMap, the net positive observers are rewarded.
+		ballot.BuildRewardsDistribution(rewardsDistributeMap)
 	}
+	sortedKeys := make([]string, 0, len(rewardsDistributeMap))
+
+	for address, rewardUnits := range rewardsDistributeMap {
+		sortedKeys = append(sortedKeys, address)
+		// Rewards are only distributed for correct votes,calculate the total rewards units from the final map to allocate maximum rewards possible to observers
+		if rewardUnits > 0 {
```
lumtis, line 207: Not needed

kingpinXD: You mean `rewardUnits > 0`?

This is needed, as we only allocate rewards for correct votes. This makes sure the entire rewards are utilizied

x/emissions/abci.go
```diff
@@ -185,31 +185,48 @@ func distributeRewardsForMaturedBallots(
...
 		if err != nil {
 			ctx.Logger().Error("Error while parsing observer address ", "error", err, "address", key)
 			continue
 		}
-		// observerRewardUnits can be negative if the observer has been slashed
-		// an observers earn 1 unit for a correct vote, and -1 unit for an incorrect vote
 		observerRewardUnits := rewardsDistributeMap[key]
-
 		if observerRewardUnits == 0 {
 			finalDistributionList = append(finalDistributionList, &types.ObserverEmission{
```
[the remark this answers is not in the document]
kingpinXD: This finalDistribution list is used for emitting events only . I maintained the logic to ensure consistency with existing events. 
We could at some point remove this event entirely if no one is using it .

x/emissions/abci.go
kingpinXD: I moved the totalRewards caluculation out , so this should not change the allocation .
Just the portion of rewards which was allocated for incorrect observers would not now be distributed among others instreat of being retained by protocol

x/observer/types/ballot.go
kingpinXD: The map is initially empty, and values are added since this function is called for each ballot.

ws4charlie (approved):
the new logic looks correct

x/observer/types/ballot.go
lumtis: Yeah but it's a cleaner API then to have the function populating the map to initialize it and return it.

x/observer/types/ballot.go
```diff
@@ -92,38 +92,31 @@ func CreateVotes(listSize int) []VoteType {
...
-func (m Ballot) BuildRewardsDistribution(rewardsMap map[string]int64) int64 {
-	// If the ballot is in progress, return 0, we do not want to distribute rewards for in progress ballots
-	if m.BallotStatus == BallotStatus_BallotInProgress {
-		return 0
-	}
-
-	// Initial value for total reward units
-	totalRewardUnits := int64(0)
+func BuildRewardsDistribution(ballots []Ballot) map[string]int64 {
+	rewardsMap := map[string]int64{}
```
lumtis, line 96: Can we use make to initialize the map?

PR #3915, opened 2025-05-26
fix: add support for non-base64 formatted payload in Sui DepositAndCall event parsing

lumtis:
Some context: [Inconsistent payload format when querying depositAndCall event in Sui node](https://github.com/zeta-chain/node/issues/3919)

The depositAndCall parsing will not work on testnet if the node doesn't return an error with base64

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of payload data from different Sui RPC environments, ensuring compatibility with both base64-encoded and raw byte formats.

- **Tests**
	- Enhanced test coverage to validate correct parsing of both base64-encoded and raw byte payloads.
	- Added a new live test to verify successful transaction parsing on the Sui testnet.

### DIFF
--- a/pkg/contracts/sui/gateway.go
+++ b/pkg/contracts/sui/gateway.go
@@ -379,10 +379,8 @@ func convertPayload(data []any) ([]byte, error) {
 	// TODO: fix this discrepancy
 	// https://github.com/zeta-chain/node/issues/3919
 	base64DecodedPayload, err := base64.StdEncoding.DecodeString(string(payload))
-	if err == nil { 
-	   return base64DecodedPayload, nil
-	}
-		return payload, nil
+	if err == nil {
+		return base64DecodedPayload, nil
 	}
 	return payload, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/contracts/sui/gateway.go
```diff
@@ -373,13 +373,16 @@ func convertPayload(data []any) ([]byte, error) {
...
 
-	// Sui encode bytes in base64
-	decodedPayload, err := base64.StdEncoding.DecodeString(string(payload))
+	// Try decoding the payload from base64
+	// If the payload is not base64 encoded, directly return the payload bytes as is
+	// Currently the localnet Sui RPC will return bytes in Base64 for the payload data while live network return the actual bytes of the payload directly
+	// TODO: fix this discrepancy
+	// https://github.com/zeta-chain/node/issues/3919
+	base64DecodedPayload, err := base64.StdEncoding.DecodeString(string(payload))
 	if err != nil {
```
swift1337, line 382:
```suggestion
	if err == nil { 
	   return base64DecodedPayload, nil
	}
```

pkg/contracts/sui/gateway.go
```diff
@@ -373,13 +373,16 @@ func convertPayload(data []any) ([]byte, error) {
...
-		return nil, errors.Wrap(err, "failed to decode payload from base64")
+		return payload, nil
 	}
-
-	return decodedPayload, nil
+	return base64DecodedPayload, nil
```
swift1337, line 385:
```suggestion
	return payload, nil
```

PR #3930, opened 2025-05-28
fix: post zero priority fee to avoid gas price bump failure in zetacore

ws4charlie:
# Description

The Base mainnet CCTX outbound keeps failing. Most likely the gas price in the [CCTX](https://zetachain.blockpi.network/lcd/v1/public/zeta-chain/crosschain/gasPrice/8453) is too low. The outbound tx is dropped by the endpoint right after broadcasting.

We rely on the gas stability pool to increase CCTX's gas price but the gas bumping logic fails due to below error. The fix is to post `zero` priority fee to avoid this failure and allow gas stability pool to kick in.

image

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/evm/observer/observer_gas.go
+++ b/zetaclient/chains/evm/observer/observer_gas.go
@@ -15,11 +15,17 @@ func (ob *Observer) PostGasPrice(ctx context.Context) error {
 		return errors.Wrap(err, "unable to suggest gas price")
 	}
 
+	// The zetaclients now only build legacy tx rather than EIP-1559 tx.
+	// Hardcode priority fee to zero to avoid gas price bump failure in the zetacore:
+	// https://github.com/zeta-chain/node/blob/release%2Fv30/x/crosschain/keeper/abci.go#L182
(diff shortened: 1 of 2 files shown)


zetaclient/chains/evm/observer/observer_gas.go
```diff
@@ -15,11 +15,17 @@ func (ob *Observer) PostGasPrice(ctx context.Context) error {
 		return errors.Wrap(err, "unable to suggest gas price")
 	}
 
+	// The zetaclients now only build legacy tx rather than EIP-1559 tx.
+	// Hardcode priority fee to zero to avoid gas price bump failure in the zetacore:
+	// https://github.com/zeta-chain/node/blob/release%2Fv30/x/crosschain/keeper/abci.go#L182
+	// https://github.com/zeta-chain/node/issues/3221
+	priorityFee := big.NewInt(0)
```
lumtis, line 22: Just use 0 directly as a uint64

kingpinXD (approved):

Issue #3947, opened 2025-06-03
ZETA Gateway integration E2E tests


lumtis:
Add E2E test coverage for ZETA token support with the gateway (exclusively for EVM chain for now)

Minimum coverage:
```
zeta_deposit
zeta_deposit_and_call
zeta_deposit_and_call_revert
zeta_deposit_and_call_revert_with_call
zeta_withdraw
zeta_withdraw_and_call
zeta_withdraw_and_call_revert
zeta_withdraw_and_call_revert_with_call
```

Should contain a new runner in the local E2E tests

PR #3962, opened 2025-06-06
feat: add a new message to migrate connector funds

kingpinXD:
# Description
Closes : https://github.com/zeta-chain/node/issues/3947

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to migrate connector funds from V1 to V2 connectors via a new CLI command and message type.
  - Added new end-to-end tests for V2 connector migration and V2 Zeta deposit flows.
  - Updated OpenAPI and CLI documentation to include the new migration command and response schema.
  - Deployed and integrated the new ZetaConnectorNative contract in protocol setup.

- **Bug Fixes**
  - Improved error reporting for revert transaction failures in cross-chain processing.

- **Documentation**
  - Expanded CLI and protocol documentation to cover connector fund migration commands and messages.

- **Tests**
  - Added comprehensive unit and integration tests for migration logic, CLI commands, and message validation.

- **Chores**
  - Updated dependencies and internal workflow to support new test categories and contract versions.


x/crosschain/keeper/msg_server_migrate_connector_funds.go
```diff
@@ -0,0 +1,99 @@
...
+	"context"
+
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	authoritytypes "github.com/zeta-chain/node/x/authority/types"
+	"github.com/zeta-chain/node/x/crosschain/types"
+)
+
+func (k msgServer) MigrateConnectorFunds(
```
lumtis, line 13: The PR should not be marked test if we introduce such message

lumtis (commented):
Before introducing a new Cosmos message for migrating the funds of the connector, I would rather prefer to explore if there is a way to do it entirely on the smart contract side.

For example an admin can update the TSS, so you could think about a batch tx transaction containing:
- update TSS address to the sender
- use new provilege to move the ZETA
- revert back old TSS address
- eventually pause the connector

For now as we told I would prefer to let this PR just keep the structure of how the migration test looks like, even if we don't migrate connector funds yet.

PR #3960, opened 2025-06-05
feat: migrate to golangci-lint v2

swift1337:
- [x] Migrate linter to v2
- [x] Apply linter and fixes
- [x] Embed golines formatter into the linter
- [x] Update CI
- [x] Optimize exclusion list

## Summary by CodeRabbit

- **Chores**
  - Updated linting and formatting tools to newer versions and modernized configuration for improved code quality checks.
  - Streamlined build and formatting processes by replacing custom scripts with integrated tooling.
- **Refactor**
  - Improved code readability and maintainability by simplifying conditional logic, removing redundant code, and clarifying boolean expressions throughout the codebase.
  - Enhanced resource management in periodic tasks to prevent potential leaks.
- **Style**
  - Applied consistent formatting and removed unnecessary comments and directives for cleaner code.
- **Bug Fixes**
  - Tightened authorization checks for tracker management, ensuring stricter validation of permissions.

PR #3958, opened 2025-06-05
feat(zetaclient): integrate TON HTTP-RPC

swift1337:
This PR drops Lite-Client support in favor of HTTP-RPC based on https://toncenter.com/api/v2/#/

- [x] Integrate toncenter http-api v2 into zetaclient. Fully compatible with our needs and existing binding to Tongo!
- [x] **Breaking**: `voteInbound` for TON now reports block_height=0. Should not be an issue
- [x] Fix e2e tests
- [x] Update ton-docker image
- [x] Simplify ensureLastScannedTx
- [x] Update & simplify configs
- [x] Add http mock for ton rpc
- [x] Add prom metrics to rpc
- [x] Use the new **ton-docker** image

Close https://github.com/zeta-chain/node/issues/3928

## Summary by CodeRabbit

- **New Features**
  - Integrated TON http-rpc, replacing previous lite-client-based connectivity for improved blockchain interaction and configuration.

- **Refactor**
  - Migrated all TON blockchain operations from lite-client to http-rpc, including observer, signer, orchestrator, and test utilities.
  - Updated configuration files and environment defaults to use TON Center v2 RPC endpoints.
  - Simplified client initialization and improved error handling for TON network operations.

- **Bug Fixes**
  - Enhanced transaction polling and status checks for more reliable TON network synchronization.

- **Tests**
  - Updated and expanded test suites to support the new TON http-rpc interface, including new test utilities for RPC server mocking and account parsing.

- **Chores**
  - Removed deprecated lite-client code, configuration support, and related test files to streamline the codebase.

PR #3929, opened 2025-05-28
feat(ton): Implement HTTP-RPC based on toncenter v2

swift1337:
- [x] `GetMasterchainInfo`
- [x] `GetBlockHeader`
- [x] `HealthCheck`
- [x] `GetAccountState`
- [x] `GetConfigParam`
- [x] `GetTransactions`
- [x] `GetTransaction`
- [x] `GetTransactionsSince`
- [x] `GetFirstTransaction`
- [x] `SendMessage`
- [x] Add prometheus metrics

Closes https://github.com/zeta-chain/node/issues/3927

PR #3973, opened 2025-06-11
feat(v31): cherry-pick TON RPC

swift1337:
- https://github.com/zeta-chain/node/pull/3929
- https://github.com/zeta-chain/node/pull/3960
- https://github.com/zeta-chain/node/pull/3958

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## UNRELEASED
+## v31.1
 
 ### Features
 


changelog.md
```diff
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## v31.1
+
```
ws4charlie, line 4: Not sure if we need this in release patch (there is a `no-changelog` label). Otherwise, the PR# would be 3973?

PR #4016, opened 2025-07-16
fix: sui stability pool

kingpinXD:
# Description

- Adds a new migration script to burn all the zrc20 tokens minted to the Stabilty Pool.
- Updates the `ObservedOutboundEffectiveGasLimit` to the same number as `ObservedOutboundGasUsed` . Although the numbers are not accurate this disables any funcding to the stabilty pool. We can also consider adding logic to zetacore directly to ignore sui chain when funding the stabilty pool , but doing it this way has the most minimal changes . It also provides us to enable stabilty pool for sui through a zetaclient only fix . 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/x/fungible/migrations/v4/migrate.go
+++ b/x/fungible/migrations/v4/migrate.go
@@ -1,0 +1,66 @@
+package v4
+
+import (
+	"fmt"
+	"math/big"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	ethermint "github.com/zeta-chain/ethermint/types"
+	"github.com/zeta-chain/node/pkg/chains"
+
+	"github.com/zeta-chain/node/x/fungible/types"
+)
+
+type fungibleKeeper interface {
+	GetAllForeignCoins(ctx sdk.Context) (list []types.ForeignCoins)
+	QuerySystemContractGasCoinZRC20(ctx sdk.Context, chainid *big.Int) (ethcommon.Address, error)
+	ZRC20BalanceOf(ctx sdk.Context, zrc20Address, owner ethcommon.Address) (*big.Int, error)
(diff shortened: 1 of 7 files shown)


lumtis:
> The PR should be targeted to `release/v32`, not `release-v32.0.`

Actually no, we need to create `release/v33` from `release/v32` and target this PR to v33

x/fungible/migrations/v4/migrate.go
```diff
@@ -0,0 +1,72 @@
...
+	stabilityPoolAddress := types.GasStabilityPoolAddressEVM()
+
+	suiBalance, err := fungibleKeeper.ZRC20BalanceOf(ctx, suiGasZRC20, stabilityPoolAddress)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get SUI balance for stability pool %s", stabilityPoolAddress)
+	}
+
+	err = fungibleKeeper.CallZRC20Burn(ctx, stabilityPoolAddress, suiGasZRC20, suiBalance, true)
+	if err != nil {
+		return err
```
lumtis, line 56: Why returning error here unlike the above?

zetaclient/chains/sui/observer/outbound.go
```diff
@@ -134,6 +133,9 @@ func (ob *Observer) VoteOutbound(ctx context.Context, cctx *cctypes.CrossChainTx
 		return errors.Wrap(err, "unable to parse gas used")
 	}
 
+	// TODO : Update the effective gas limit and gas used.
+	// Setting the effective gas limit to the gas used disables the gas stability pool funding
+	// https://github.com/zeta-chain/node/issues/4015
 	// Create message
 	msg := cctypes.NewMsgVoteOutbound(
```
lumtis, line 140: Even with this fix, can we disable the gas stability pool funding for Sui as it will not be necessary?

This is a extra safety issue, we can test the fix on `develop` in the future

lumtis:
Only ZetaClient upgrade might be needed in the end, let's wait from feedbacks from the rest of the team.

PR #4020, opened 2025-07-16
fix: burn sui stabilty pool balance using a migration script

kingpinXD:
# Description
- Adds a migration script to delete ZRC20 SUI gas tokens from stability pool 
- Sets v32 as the base release for upgrade tests 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

* **New Features**
  * Upgraded the consensus version for the fungible module to version 4, introducing a new migration that burns SUI gas tokens from the stability pool during the upgrade.
  * Updated end-to-end tests to require the latest supported version.

* **Bug Fixes**
  * Improved error handling and logging during the migration process to ensure smoother upgrades.

* **Tests**
  * Added comprehensive tests for the new migration logic and helper functions.
  * Removed outdated migration tests for previous versions.

* **Chores**
  * Updated version references and migration paths to align with the latest release.


x/fungible/migrations/v4/migrate.go
```diff
@@ -0,0 +1,79 @@
...
+	) error
+}
+
+// MigrateStore migrates the store from consensus version 3 to 4.
+// It burns the SUI gas ZRC20 from the stability pool address.
+func MigrateStore(ctx sdk.Context, fungibleKeeper fungibleKeeper) error {
+	chainID, err := ethermint.ParseChainID(ctx.ChainID())
+	if err != nil {
+		// It's fine to return nil here and not try to execute the migration at all if the parsing fails
+		ctx.Logger().Error("failed to parse chain ID", "chain_id", ctx.ChainID(), "error", err)
```
kingpinXD, line 34: @swift1337 
Answering https://github.com/zeta-chain/node/pull/4016#discussion_r2209793782

Keeping the error log here is fine, this would be executed only once at the upgrade height , we are not returning an error so it is not redundant in this case

kingpinXD:
> Looks good to me
> 
> As an extra check: the gas stability pool will be filled with Sui token in the previous version for upgrade tests Maybe we could add Sui tests in `start-upgrade-test` then check the balance is 0 after the upgrade

6b7ef3e891190623370c59b06f55d7b9dcf77536

e2e/runner/upgrade.go
```diff
@@ -31,7 +34,7 @@ func (r *E2ERunner) RunGatewayUpgradeTestsExternalChains(conf config.Config, opt
 		r.SolanaVerifyGatewayContractsUpgrade(conf.AdditionalAccounts.UserSolana.SolanaPrivateKey.String())
 	}
 
-	if opts.TestSui {
+	if opts.TestSui && !r.IsRunningUpgrade() {
```
kingpinXD, line 37: Not entirely sure why the gateway upgrade for sui is failing , still looking into it

PR #4045, opened 2025-07-23
fix: just in case we need to delay gas price vote

ws4charlie:
# Description

In case we want to avoid too many gas price vote due to repeated `zetaclient` restart.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


ws4charlie:
The CCTX processing was restored. Close this.

PR #4049, opened 2025-07-24
test(e2e): increase upgrade tests upgrade height to 240

lumtis:
This value mitigate the bug attached, seems we added too many tests over time that don't complete before the upgrade hieght.

## Summary by CodeRabbit

* **Chores**
  * Updated the upgrade test environment to trigger at a new block height.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -380,7 +380,7 @@ start-upgrade-test: zetanode-upgrade solana
 	@echo "--> Starting upgrade test"
 	export LOCALNET_MODE=upgrade && \
 	export UPGRADE_HEIGHT=225 && \
-	export E2E_ARGS="--test-solana --test-sui" && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -379,7 +379,7 @@ endif
 start-upgrade-test: zetanode-upgrade solana
 	@echo "--> Starting upgrade test"
 	export LOCALNET_MODE=upgrade && \
-	export UPGRADE_HEIGHT=225 && \
+	export UPGRADE_HEIGHT=240 && \
```
kingpinXD, line 382: NIT : `start-upgrade-import-mainnet-test` also uses this height , maybe consider using a constant for both values.

ws4charlie (approved):
cool

PR #4061, opened 2025-07-29
feat(`crosschain`): add event containing error when inbound processing fail

lumtis:
# Description

In the future, this will allow to check in the Cosmos tx associated with an inobund, the reason for the failure at the processing level


proto/zetachain/zetacore/crosschain/events.proto
```diff
@@ -81,3 +81,8 @@ message EventERC20CustodyPausing {
   bool pause = 2;
   string cctx_index = 3;
 }
+
+message EventInboundProcessingFailure {
+  string inbound_hash = 1;
```
swift1337, line 86: Should it include `chain_id`, `cctx_index`?

lumtis: No `cctx_index` since the cctx is not created at this point.
For chain ID it could be parsed from the logs or doing the event emitting deeper in the stack but I would keep it simpler as these data can be retrieved from the inbound hash

swift1337 (approved):

PR #4067, opened 2025-07-31
fix: disable sui authenticated call temporarily until gateway upgrade

ws4charlie:
# Description

What's changed
1.  Still use old `gateway.mv` and pre-compiled `connected.mv` in the E2E tests setup.
2. Still use old upgrade package (in `protocol-contracts-sui-grade-legacy`) for gateway upgrade test.
3. Add legacy non-authenticated call E2E tests; disable authenticated call E2E tests.
4. Disable `MessageContext` object/ID usage in all related functions in the `zetaclient` Sui signer.
5. Still use old `withdrawAndCall` error handling; disable new error handling logic.

The PR to re-enable Sui authenticated call feature can be worked on once we cut `release/v33`. @lumtis 

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporarily disabled the Sui authenticated call feature and related tests until a gateway upgrade is performed.
  * Switched to legacy Sui gateway contracts and binaries for end-to-end testing and deployment.
  * Updated error handling and retry logic to use legacy methods for Sui transactions.
  * Disabled or replaced message context handling in Sui gateway flows and related tests.
  * Added and enabled legacy Sui withdraw-and-call tests to maintain coverage during the transition.


skosito (approved):

lumtis (commented):
Instead of commenting out, I was wondering if it was possible to introduce a boolean "feature flags" instead to switch the logic? 
In theory main difference is that you don't need to fetch a messageContext and pass it to onCall?
Also wondering if we could do the same with the withdrawAndCall tests? In theory nothing change for the user perspective outside of the contract you want to call (since the authenticated call should verify you call from the right caller)

ws4charlie:
> Instead of commenting out, I was wondering if it was possible to introduce a boolean "feature flags" instead to switch the logic? In theory main difference is that you don't need to fetch a messageContext and pass it to onCall? Also wondering if we could do the same with the withdrawAndCall tests? In theory nothing change for the user perspective outside of the contract you want to call (since the authenticated call should verify you call from the right caller)

Did you mean bring [this version of code](https://github.com/zeta-chain/node/pull/4007#discussion_r2210838018) back, with an additional bool switch? like keeping both examples packages in the config, e2e, and also support both styles of `on_call`? It is not a quick change if you want to cut the code today. wdyt @lumtis

lumtis (approved):

PR #4096, opened 2025-08-13
fix: use evm chain id from eth config in rpc backend

skosito:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/cmd/zetacored/root.go
+++ b/cmd/zetacored/root.go
@@ -204,8 +204,8 @@ func InitAppConfig(denom string, evmChainID uint64) (string, interface{}) {
 	// - if you set srvCfg.MinGasPrices non-empty, validators CAN tweak their
 	//   own app.toml to override, or use this default value.
 	//
-	// In this example application, we set the min gas prices to 0.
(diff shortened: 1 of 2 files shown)


cmd/zetacored/root.go
```diff
@@ -204,8 +204,8 @@ func InitAppConfig(denom string, evmChainID uint64) (string, interface{}) {
 	// - if you set srvCfg.MinGasPrices non-empty, validators CAN tweak their
 	//   own app.toml to override, or use this default value.
 	//
-	// In this example application, we set the min gas prices to 0.
-	srvCfg.MinGasPrices = "0" + denom
+	// In this application, we set the min gas prices to 1.
+	srvCfg.MinGasPrices = "1" + denom
```
skosito, line 208: i think if this is in app.toml already it is fine, this is just to be consistent with live networks min gas prices

ws4charlie (approved):

kingpinXD (approved):

Issue #4102, opened 2025-08-14
Error deploying contracts on testnet


fadeev:
> Anyone stumbled upon testnet deployment issue after recent upgrade? Same scripts with the same code were working 2 days ago.

RPC zetachain-athens.g.allthatnode
```
[⠊] Compiling...
[⠒] Compiling 2 files with Solc 0.8.19
[⠢] Solc 0.8.19 finished in 45.36ms
Compiler run successful!
Error: deserialization error: expected value at line 1 column 1
```

RPC zetachain-athens-evm.blockpi.network
```
Error: Failure on receiving a receipt for 0xb7bc18da13ff9efb945771a57de08c79f5e42466620de23021f5077a9e1c9025:
server returned an error response: error code -32000: GetTxByEthHash 0xb7bc18da13ff9efb945771a57de08c79f5e42466620de23021f5077a9e1c9025: ethereum tx not found

RPC Alchemy
```
[⠊] Compiling...
[⠒] Compiling 2 files with Solc 0.8.19
[⠢] Solc 0.8.19 finished in 44.91ms
Compiler run successful!
Error: Failed to send transaction

Context:
- server returned an error response: error code -32000: incorrect chain-id; expected 262144, got 7001
```

deployment script:

```
forge create src/Test.sol:Test \
  --rpc-url $ZETACHAIN_RPC_URL \
  --private-key $PRIVATE_KEY \
  --chain-id 7001 \
  --broadcast \
  --constructor-args \
  $ENTROPY_CONTRACT_ADDRESS \
```
  $VAULT_ADDR

foundry.toml:
[profile.default]
src = "src"
out = "out"
libs = ["lib","node_modules"]
solc_version = "0.8.19"  # Pre-Shanghai version
evm_version = "london"   # Disables PUSH0

PR #4112, opened 2025-08-18
fix: error when deploying contracts on testnet

skosito:
# Description

closes: #4102 

in `eth_getTransactionReceipt` nil should be returned if tx is not found, same as geth, so it is recognized by tooling like forge (source: https://github.com/cosmos/evm/pull/379)

tested locally (`make start-localnet`), before the fix:

```
forge create --rpc-url http://localhost:9545 --private-key $PRIVATE_KEY e2e/contracts/example/Example.sol:Example --broadcast

...

Error: contract was not deployed
```

after the fix:

```
forge create --rpc-url http://localhost:9545 --private-key $PRIVATE_KEY e2e/contracts/example/Example.sol:Example --broadcast

...

Deployer: 0xE5C5367B8224807Ac2207d350E60e1b6F27a7ecC
Deployed to: 0xA8D5060feb6B456e886F023709A2795373691E63
Transaction hash: 0x2cea603bc12431c377276553f1c5d30738a7c1a0c47c8abab9ce38b86ff33208
```

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- New Behavior
  - Requesting a transaction receipt for an unknown transaction no longer returns an error; clients now receive a nil/empty receipt, improving API consistency.

- Tests
  - Updated test cases to reflect the non-error response when a transaction receipt is not found.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@
 ### Fixes
 
 * [4090](https://github.com/zeta-chain/node/pull/4090) - print error message in detail if unable to decode Bitcoin memo
+* [4112](https://github.com/zeta-chain/node/pull/4112) - fix error when deploying contracts on testnet
 
 ### Tests
 


rpc/backend/tx_info.go
```diff
@@ -174,7 +174,7 @@ func (b *Backend) GetTransactionReceipt(hash common.Hash) (map[string]interface{
 	res, additional, err := b.GetTxByEthHash(hash)
 	if err != nil {
 		b.Logger.Debug("tx not found", "hash", hexTx, "error", err.Error())
-		return nil, err
```
lumtis, line 177: Is it the only type of error you can have?
Wondering if there could be some filtering we could do here

fadeev:
Brilliant, thanks for fixing this so quickly!

PR #4115, opened 2025-08-19
chore: set v34 as upgrade test initial version and set back precompiles in localnet setup

lumtis:
# Description

Add the precompiles back now that the upgrade should work since the initial version has the precompiles in the params

Check that the staking precompiles work by just providing staking and slashing as part of the enabled precompiles .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -355,7 +355,7 @@ ifdef UPGRADE_TEST_FROM_SOURCE
 zetanode-upgrade: e2e-images
 	@echo "Building zetanode-upgrade from source"
 	$(DOCKER) build -t zetanode:old -f Dockerfile-localnet --target old-runtime-source \
-		--build-arg OLD_VERSION='release/v32' \
+		--build-arg OLD_VERSION='release/v34' \
 		--build-arg NODE_VERSION=$(NODE_VERSION) \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -375,23 +375,20 @@ start-upgrade-test: zetanode-upgrade solana
...
 	export UPGRADE_HEIGHT=60 && \
-	export USE_ZETAE2E_ANTE=true && \
 	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile upgrade -f docker-compose-upgrade.yml up -d
 
 start-upgrade-test-admin: zetanode-upgrade
 	@echo "--> Starting admin upgrade test"
 	export LOCALNET_MODE=upgrade && \
 	export UPGRADE_HEIGHT=90 && \
 	export E2E_ARGS="${E2E_ARGS} --skip-regular --test-admin" && \
-	export USE_ZETAE2E_ANTE=true && \
```
skosito, line 394: should we also clean up upgrade handlers added in zetaclient? `upgrade_v33.go` is the name of the file

lumtis: Yeah, we'll need to add comment next time to precise that this is a section to clean up after upgrade

kingpinXD (approved):

Issue #4106, opened 2025-08-18
Remove ConfirmationMode from inbound digest calculation


lumtis:
Since the ConfirmationMode doesn't pertain to the inbound data and its uniqueness itself but to how it is observed, the ConfirmationMode value should be ignored when calculating the digest of an inbound to get the ballot index.

PR #4116, opened 2025-08-19
fix(`crosschain`): remove confirmation mode from outbound and inbound digest

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/4106

## Summary by CodeRabbit

- Bug Fixes
  - Standardized digest computation for inbound and outbound vote messages to be independent of confirmation mode, ensuring consistent hashes and improving reliability of cross-chain message verification.

- Tests
  - Updated test cases to confirm that changing the confirmation mode does not affect digest results.

- Documentation
  - Added changelog entry under Unreleased > Fixes noting the removal of confirmation mode influence from digest calculations.

### DIFF
--- a/x/crosschain/types/message_vote_inbound.go
+++ b/x/crosschain/types/message_vote_inbound.go
@@ -162,6 +162,7 @@ func (msg *MsgVoteInbound) Digest() string {
 	m := *msg
 	m.Creator = ""
 	m.InboundBlockHeight = 0
+	m.ConfirmationMode = ConfirmationMode_SAFE
 	hash := crypto.Keccak256Hash([]byte(m.String()))
 	return hash.Hex()
 }
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@
 ### Fixes
 
 * [4090](https://github.com/zeta-chain/node/pull/4090) - print error message in detail if unable to decode Bitcoin memo
+* [4116](https://github.com/zeta-chain/node/pull/4116) - remove confirmation mode from outbound and inbound digest
 
 ### Tests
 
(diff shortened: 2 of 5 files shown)


x/crosschain/types/message_vote_inbound.go
```diff
@@ -162,6 +162,7 @@ func (msg *MsgVoteInbound) Digest() string {
 	m := *msg
 	m.Creator = ""
 	m.InboundBlockHeight = 0
+	m.ConfirmationMode = ConfirmationMode_SAFE
```
kingpinXD, line 165: Replaying old transactions with ConfirmationMode_FAST would now generate a new index , but we should still be protected since the finalization index would exist . 
Please conisder adding a unit test to verify that .

kingpinXD: I don't think the ballot index would be the same for an inbound which was previously created with `ConfirmationMode_FAST`. 
In cases where we end up replaying a block , or creating a tracker .

But I belive the finalization index would still protect us , feel free to resolve the comment if you still  think the test is unnecessary

lumtis: >I don't think the ballot/cctx index would be the same for an inbound which was previously created with ConfirmationMode_FAST.
In cases where we end up replaying a block , or creating a tracker .

Before this upgrade we will have the patch that enforce all inbound to use SAFE so we will not have the index issue during the transition if this is the question

>Using only one single digest definitely fixes the issue. Looks like a fix in msg_server_vote_inbound_tx.go is also a small change (maybe I'm wrong?).
I was able to replicate it in a branch replicate-double-spent and start-e2e-test will trigger the edge case in modified E2E TestETHDeposit.
Feel free to resolve my comment if you think allowing two digests is not a easy fix because I might be wrong.

We should not add it in the digest even if it's fixed outside. The digest should be based on inbound data itself, not how it gets observed

kingpinXD (approved):

ws4charlie (approved):

PR #4119, opened 2025-08-20
fix: dbg trace block by number gas limit legacy

skosito:
# Description

Started noticing issues with v34, but checking blocks on testnet, it was failing for pre-v31 blocks, and worked after v31. 
With tracer specified in request error was:
```
"rpc error: code = Internal desc = runtime error: index out of range [0] with length 0"
```

Without tracer specified:
```
"error": "rpc error: code = Internal desc = apply message: intrinsic gas too low"
```

 Only change in v31 changelog that could impact this is indexing gas limit for synthetic txs and using correct gas limit: https://github.com/zeta-chain/node/pull/3920/files. However, for pre-v31 txs `GasUsed` field was user instead, so tracing those blocks was using wrong value (0 since it cant be read from cosmos events), leading to issues when applying msgs in trace block.
 
 This PR falls back to previously used value in case gas limit is not present in cosmos events for synthetic txs.

# How Has This Been Tested?

Tested locally by reverting this fix on branch branched out from v32, and running upgrade tests v32 -> v34, error was there for blocks before upgrade, and gone after fixing this in v34 code.

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/rpc/backend/blocks.go
+++ b/rpc/backend/blocks.go
@@ -323,10 +323,14 @@ func (b *Backend) parseSyntethicTxFromAdditionalFields(
 	additional *rpctypes.TxResultAdditionalFields,
 ) *evmtypes.MsgEthereumTx {
 	recipient := additional.Recipient
+	gas := additional.GasUsed
+	if additional.GasLimit != nil {
+		gas = *additional.GasLimit
+	}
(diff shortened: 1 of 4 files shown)


ws4charlie (approved):
change looks reasonable

kingpinXD (approved):

PR #4125, opened 2025-08-21
feat: add conditional logic for adding erc20 module based on chain id

kingpinXD:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/app/setup_handlers.go
+++ b/app/setup_handlers.go
@@ -7,7 +7,6 @@ import (
 	storetypes "cosmossdk.io/store/types"
 	"cosmossdk.io/x/upgrade/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
-	erc20types "github.com/cosmos/evm/x/erc20/types"
 	"golang.org/x/mod/semver"
 
 	"github.com/zeta-chain/node/pkg/constant"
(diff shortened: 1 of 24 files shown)


app/setup_handlers.go
```diff
@@ -30,28 +33,27 @@ func GetDefaultUpgradeHandlerVersion() string {
...
+		storeUpgrade: &storetypes.StoreUpgrades{
+			Added: []string{erc20types.ModuleName},
 		},
+	}
+
+	var upgrades []upgradeTrackerItem
+	if app.ChainID() != "" {
+		evmChaindID, err := chains.CosmosToEthChainID(app.ChainID())
+		if err != nil {
+			fmt.Println("Skip here")
```
lumtis, line 48: I think we should rather panic, better to have the chain halting than moving into an inconsistent state

app/setup_handlers.go
```diff
@@ -30,28 +33,27 @@ func GetDefaultUpgradeHandlerVersion() string {
...
+	addErc20ModuleUpgradeTracker := upgradeTrackerItem{
+		index: 1752528615,
+		storeUpgrade: &storetypes.StoreUpgrades{
+			Added: []string{erc20types.ModuleName},
 		},
+	}
+
+	var upgrades []upgradeTrackerItem
+	if app.ChainID() != "" {
+		evmChaindID, err := chains.CosmosToEthChainID(app.ChainID())
```
lumtis, line 46: I think this specific logic can be put into a function unit tested

app/setup_handlers.go
```diff
@@ -29,29 +31,30 @@ func GetDefaultUpgradeHandlerVersion() string {
...
+		}
+	}
+	return upgrades
+}
+
 func SetupHandlers(app *App) {
 	allUpgrades := upgradeTracker{
-		upgrades: []upgradeTrackerItem{
-			// TODO: enable back IBC
-			// these commented lines allow for the IBC modules to be added to the upgrade tracker
```
lumtis, line 36: Why removing? It nice to keep these TODO if we want to work on IBC later

lumtis (approved):

Issue #4132, opened 2025-08-25
Allow the observer to observe Sui deposit events from several packages


lumtis:
Part of the process to add support or Sui authenticated withdrawAndCall will be to upgrade the current Sui gateway package.

When you upgrade a Sui package, a new package containing the updated code will be published. The old package still exist, and can still be called but contains the old code.

**Problem**

Currently the observer listen for deposit events from a specific gateway package specified in the chain params.

Part of the gateway upgrade process is to publish this new upgraded version and update the chain params with the new package (so the withdraw can be called with the right interface)

BUT after updating the package in the chain params, the observer will stop observing events from the old package, while user can still deposits in the old package (there is no way to pause this package, only the gateway object can be paused which will pause all gateway packages at the same time). These deposits will be loss of funds to the users, the tokens will be deposited to the gateway but no inbound observed.

**Solutions**

1: simplest: give a very long notice to devs and users that the gateway will be updated and old one will be ignored

2: better: add support to observe several gateway packages at the same time. At least 2 to allow a transition period during a gateway upgrade.

Example:

The Sui chain params gateway is currently in the following format:
```
<package-id>,<gateway-id>
```
A new optional field can be added that contains the former package id for the observer to read old deposits.
```
<package-id>,<gateway-id>,<former-package-id>
```

Important:
The deposit on the former package should be tested in the Sui upgrade test, to ensure that it is observed a single time, there is no double observation.

PR #4138, opened 2025-08-27
feat: observe inbound from previous gateway package after upgrade

ws4charlie:
# Description

### The current Sui gateway address is in the following format
```text
<package-id>,<gateway-id>
````

After gateway upgrading, the observer still needs the `previous gateway package ID`. The most obvious reason is that filtering inbound events requires the old package ID as parameter, otherwise the Sui API will return empty result if `new package ID` is passed. The new format of Sui gateway address will be

```text
// <withdrawcap-id>, <previous-package-id> and <original-package-id> are mandatory after gateway upgrade!!
<package-id>,<gateway-id>,<withdrawcap-id>,<previous-package-id>,<original-package-id>
```

### In conclusion, the `zetaclient` will

1. Continue using the `originalPackageID` to query inbound events (`DepositEvent`, `DepositAndCallEvent`). They were defined in previous package.
2. Use the `<withdrawcap-id>` in the chain params for withdrawals if present, otherwise query the withdrawCap object under `originalPackageID`.
3. Use `new package ID` to call `increase_nonce` function. It is defined in new package and the signature has changed.

### The contents of this PR

- [x] Inbound cursor migration script. `Zetaclient` has to handle two packages, so it good to have separate scanning cursors in the DB under each package ID.
- [x] Support an additional `<withdrawcap-id>` in the Sui chain params' gateway address.
- [x] Support an additional `<previous-package-id>` and `<original-package-id>` in the Sui chain params' gateway address, allowing `zetaclientd` to work for both pre-upgrade and post-upgrade package IDs.
- [x] Add Sui **post-upgrade** E2E tests to ensure that deposits from both new and old gateway packages are working fine. 
- [x] Add Sui **post-upgrade** E2E tests to ensure that deposit from deprecated previous package won't be observed.

Closes https://github.com/zeta-chain/node/issues/4132

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- New Features
  - Sui gateway now supports observing deposits from both current and previous packages, ensuring continuity across upgrades.
  - Per-package inbound cursors with automatic migration from the previous single-cursor setup.

- Improvements
  - More informative observer logs (Sui, Solana) and added nonce logging on Sui outbound.
  - Minor wording and status log enhancements.

- Tests
  - Expanded coverage for Sui gateway multi-package handling, cursor migration, and generic string storage.

- Chores
  - Database auto-migration now includes storage for auxiliary string values.

- Documentation
  - Changelog updated.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -21,6 +21,7 @@
 * [4112](https://github.com/zeta-chain/node/pull/4112) - fix error when deploying contracts on testnet
 * [4121](https://github.com/zeta-chain/node/pull/4121) - dbg trace block by number gas limit legacy
 * [4183](https://github.com/zeta-chain/node/pull/4183) - force rescan if inbound vote monitoring fails
+* [4194](https://github.com/zeta-chain/node/pull/4194) - remove duplicate solana post-gas-price goroutine
 
 ### Tests
 
--- a/zetaclient/chains/solana/solana.go
+++ b/zetaclient/chains/solana/solana.go
@@ -100,7 +100,6 @@ func (s *Solana) Start(ctx context.Context) error {
 	register(s.observer.ProcessInboundTrackers, "process_inbound_trackers", optInboundInterval, optInboundSkipper)
 	register(s.observer.PostGasPrice, "post_gas_price", optGasInterval, optGenericSkipper)
 	register(s.observer.CheckRPCStatus, "check_rpc_status")
-	register(s.observer.PostGasPrice, "post_gas_price", optGasInterval, optGenericSkipper)
 	register(s.observer.ProcessOutboundTrackers, "process_outbound_trackers", optOutboundInterval, optOutboundSkipper)
 
 	// CCTX scheduler (every zetachain block)


zetaclient/chains/base/observer.go
```diff
@@ -381,6 +392,64 @@ func (ob *Observer) ReadLastTxScannedFromDB() (string, error) {
 	return lastTx.Hash, nil
 }
 
+// GetAnyString get any string data by key
+func (ob *Observer) GetAnyString(key string) string {
+	return ob.anyStringMap[key]
+}
+
```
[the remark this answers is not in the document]
ws4charlie: https://github.com/zeta-chain/node/pull/4138/commits/f5e172de7632f65fdde0dd49cf7e176c837f4a86

pkg/contracts/sui/gateway.go
```diff
@@ -57,22 +67,61 @@ func ActiveMessageContextDynamicFieldName() (json.RawMessage, error) {
 	return dynamicFieldNameToJSONArray("active_message_context")
 }
 
-// NewGatewayFromPairID creates a new Sui Gateway
-// from pair of `$packageID,$gatewayObjectID`
-func NewGatewayFromPairID(pair string) (*Gateway, error) {
-	packageID, gatewayObjectID, err := parsePair(pair)
+// NewGatewayFromAddress creates a new Sui Gateway struct from the gateway address in chain params.
+// The gateway address is in the form of `$packageID,$gatewayObjectID[,$withdrawCapID,originalPackageID]`
```
lumtis, line 71: gateway address is not a good name, imply a Sui address
pair is good
could be gatewayInfo as well

zetaclient/orchestrator/bootstrap.go
```diff
@@ -216,6 +218,11 @@ func (oc *Orchestrator) bootstrapSui(ctx context.Context, chain zctx.Chain) (*su
 
 	observer := suiobserver.New(baseObserver, suiClient, gateway)
 
+	// migrate inbound cursor
+	if err = observer.MigrateInboundCursorToV35(); err != nil {
```
lumtis, line 222: orchestrator bootstrap is supposed to be unopiniated

Logic related to migrate from upgrade should be clearly explained in the comment and explained, why it is needed.
It is not v35 but v36, we don't upgrade this right at v36 so the name should be different, mentioning the authenticated call change
We should also add a to track removing this from the codebase after the upgrade

pkg/contracts/sui/gateway.go
```diff
@@ -57,22 +67,61 @@ func ActiveMessageContextDynamicFieldName() (json.RawMessage, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	return NewGateway(packageID, gatewayObjectID), nil
+	return &Gateway{
+		packageID:         packageID,
+		objectID:          gatewayObjectID,
+		withdrawCapID:     withdrawCapID,
+		originalPackageID: originalPackageID,
```
lumtis, line 82: Should not be named `originalPackageID` but `previousPackageID`

We keep track of n-1, not the originall

lumtis (changes requested):
A few things need some rectifications

pkg/contracts/sui/gateway.go
ws4charlie: Not sure how correct will be the prefix `previous`. For say, we upgrade the package in future again at some point (e.g. V3 added more methods), the `previousPackageID` should change and points to V2?

e2e/utils/zetacore.go
```diff
@@ -43,17 +43,16 @@ func GetCCTXByInboundHash(
 	ctx context.Context,
 	client crosschaintypes.QueryClient,
 	inboundHash string,
-) *crosschaintypes.CrossChainTx {
+) []crosschaintypes.CrossChainTx {
```
kingpinXD, line 46: Why is this needed ? The caller is veryfying the length to be 1 anyways

zetaclient/orchestrator/bootstrap.go
ws4charlie: renamed in commit https://github.com/zeta-chain/node/pull/4138/commits/a49f613fce8b5775b0279ff52b79cedfed533980

pkg/contracts/sui/gateway.go
ws4charlie: renamed in same commit https://github.com/zeta-chain/node/pull/4138/commits/a49f613fce8b5775b0279ff52b79cedfed533980

e2e/runner/setup_sui.go
```diff
@@ -472,7 +472,7 @@ func (r *E2ERunner) whitelistSuiFakeUSDC(signer *zetasui.SignerSecp256k1, fakeUS
 }
 
 // set the chain params for Sui
-func (r *E2ERunner) setSuiChainParams() error {
+func (r *E2ERunner) setSuiChainParams(resetNonces bool) error {
```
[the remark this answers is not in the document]
ws4charlie: At the end of gateway upgrade,  the chain params (gateway address) needs to be updated with new `packageID`, but the nonces should not reset. Otherwise, the post-upgrade e2e tests (the withdrawals) will not work because the gateway nonce won't match pending nonces (reset to `0`).

zetaclient/chains/base/observer.go
```diff
@@ -390,6 +401,71 @@ func (ob *Observer) ReadLastTxScannedFromDB() (string, error) {
 	return lastTx.Hash, nil
 }
 
+// GetAuxString get any auxiliary string data by key
```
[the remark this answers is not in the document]
ws4charlie: It is used by [setCursor](https://github.com/zeta-chain/node/blob/feat-sui-observe-inbound-from-old-gateway/zetaclient/chains/sui/observer/observer.go#L178) function to save a cursor under a packageID.

e2e/runner/sui_gateway_upgrade.go
```diff
@@ -25,24 +29,35 @@ func (r *E2ERunner) SuiVerifyGatewayPackageUpgrade() {
...
+	// wait for 2 blocks to ensure the new gateway package ID is effective
+	utils.WaitForZetaBlocks(r.Ctx, r, r.ZEVMClient, 2, 10*time.Second)
+
+	// deposit from new gateway package should be observed
+	r.Logger.Print("🏃 Verifying Sui deposit from new package")
+	r.suiVerifyDepositFromPackage(r.SuiGateway.PackageID(), big.NewInt(10000000000))
+
+	// deposit from original gateway package should be observed
+	r.Logger.Print("🏃 Verifying Sui deposit from original package")
+	r.suiVerifyDepositFromPackage(r.SuiGateway.Original().PackageID(), big.NewInt(2000000))
```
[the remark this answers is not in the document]
ws4charlie: changed to previous: https://github.com/zeta-chain/node/pull/4138/commits/f68209a33c0f8e769d5b69c6b04b054e70875f05

> Original is only used to determine the type of the withdrawCap ID?

used to determine withdrawCap type and filter events.

PR #4176, opened 2025-09-10
test: adding logs for investigating missed inbound 

kingpinXD:
# Description

DO NOT MERGE 
Opening pr for any suggestions for things to check 

Things to verify
1. When we retry with a higher gas limit, I want to be sure that the `out of gas` error is becuase the votes is the `executor` [Which needs higher gas ]. In this case, querying the ballot right before the vote gets submitted should show that we are just missing 1 vote to reach 66% [ Added some logs for it ]

2 . 
```
		go func() {
		ctxForWorker := zctx.Copy(ctx, context.Background())

		errMonitor := c.MonitorVoteInboundResult(ctxForWorker, zetaTxHash, retryGasLimit, msg)
		if errMonitor != nil {
			c.logger.Error().Err(err).Msg("PostVoteInbound: failed to monitor vote inbound result")
		}
	}()

	return zetaTxHash, ballotIndex, nil
```

This logic here , although logs the error  , does not propagate it ,  it might be a good idea to propagate the error back to the called , Not sure though if it has some other impact on the zetaclient functioning 
The log line also does not contain any information about which Vote did it fail to monitor , the zeta tx should should be logged

### DIFF
--- a/zetaclient/zetacore/client_vote.go
+++ b/zetaclient/zetacore/client_vote.go
@@ -190,7 +190,11 @@ func (c *Client) PostVoteInbound(
 
 		errMonitor := c.MonitorVoteInboundResult(ctxForWorker, zetaTxHash, retryGasLimit, msg)
 		if errMonitor != nil {
-			c.logger.Error().Err(err).Msg("PostVoteInbound: failed to monitor vote inbound result")
+			logFields := map[string]any{
+				"inbound.hash":       msg.InboundHash,
+				"monitoring.zeta.tx": zetaTxHash,
+			}
+			c.logger.Error().Fields(logFields).Err(errMonitor).Msg("PostVoteInbound: failed to monitor vote inbound result")
 		}
 	}()
 


zetaclient/zetacore/client_monitor.go
```diff
@@ -165,6 +167,38 @@ func (c *Client) monitorVoteInboundResult(
...
+		}
+	}
+
+	votes := map[string]any{
+		"total":   totalVotes,
+		"success": successVotes,
+		"fail":    failVotes,
+		"no_vote": noVotes,
+	}
+	c.logger.Debug().Fields(votes).Msgf("QueryBallot Before Executing with higher limit: ballot status %s", ballot.BallotStatus.String())
```
ws4charlie, line 199: you would need `Debug() -> Info()` to be able to print out the log message in the Datadog. Additionally, adding the `inboundHash` (and even together with `zetaTxHash`) to log fields would be very helpful.

PR #4127, opened 2025-08-22
feat: Sui message context ID as dynamic field

ws4charlie:
# Description

Backport [PR 4087: Sui message context ID as dynamic field](https://github.com/zeta-chain/node/pull/4087)

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [ ] Tested in development environment
- [x] Go unit tests
- [ ] Go integration tests
- [ ] Tested via GitHub Actions

## Summary by CodeRabbit

* **New Features**
  * Added support for Sui message context ID as a gateway dynamic field, enabling improved state management.

* **Bug Fixes**
  * Enhanced payload validation with clearer error messages for invalid payloads.

* **Chores**
  * Removed legacy Sui gateway upgrade testing paths and outdated contract binaries.
  * Simplified end-to-end test infrastructure by consolidating to current implementation standards.

PR #4182, opened 2025-09-10
feat(backport/v36): observe inbound from previous gateway package after upgrade

ws4charlie:
# Description

This PR backport [4138](https://github.com/zeta-chain/node/pull/4138) and depends on [4127](https://github.com/zeta-chain/node/pull/4127)

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/cmd/zetae2e/local/performance.go
+++ b/cmd/zetae2e/local/performance.go
@@ -358,7 +358,7 @@ func suiWithdrawPerformanceRoutine(
 		r.RequestSuiFromFaucet(conf.RPCs.SuiFaucet, suiSigner.Address())
 
 		// deposit initial SUI tokens to ZEVM, 100 SUI in MIST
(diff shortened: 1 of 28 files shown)


lumtis (approved):
Looks similar to previous PR already approved

ws4charlie:
> i think it would be great to add a bit more context on why is this needed - if we upgrade gateway, then why we should observe inbounds from previous gateway version? or i maybe missunderstand something

Correct. I attached the full background of this PR to the description.

skosito:
> > i think it would be great to add a bit more context on why is this needed - if we upgrade gateway, then why we should observe inbounds from previous gateway version? or i maybe missunderstand something
> 
> Correct. I attached the full background of this PR to the description.

thanks, i didnt know this was the case, it is a bit unfortunate and it seems sui-specific, but if there is no way to pause or disable old package i guess this is the only way

Issue #4108, opened 2025-08-18
Remove `isNew` condition in FinalizedInbound check


lumtis:
In this line: https://github.com/zeta-chain/node/blob/6e8dfeed63b338c2b2565bc1816f265b8f5b51fa/x/crosschain/keeper/msg_server_vote_inbound_tx.go#L84

This condition removes the check as a condition for double spend if similar inbounds producing difference digests are observed in a short timeframe. The first inbound will not be finalized before the second one get observed.

All votes for new inbounds should be blocked as soon as a inbound with the same data is finalized.

The condition has originally been left to ensure all observers can cast their vote for an inbounds, even if they do it after the inbound get the 2/3 votes need to finalize.
We should investigate an alternative to make the check stricter:
Example: still allowing vote if the associated ballot was already finalized before vote, this means the inbound has been observed, casting more votes has no implication.

`VoteOnInboundBallot` currently return if the vote is finalized at all. This doesn't tell if it just got finalized in the current vote.
```
finalized, isNew, err := k.zetaObserverKeeper.VoteOnInboundBallot
```

The function could return the information if the inbound just got finalized or was already finalized:
```
justFinalized, wasFinalized, isNew, err := k.zetaObserverKeeper.VoteOnInboundBallot
```
If the inbound was already finalized, we can skip the check to ensure last vote can be cast
```
if !wasFinalized {
		if k.IsFinalizedInbound(tmpCtx, msg.InboundHash, msg.SenderChainId, msg.EventIndex) {
			return nil, sdkerrors.Wrapf(
				types.ErrObservedTxAlreadyFinalized,
				"%s, InboundHash:%s, SenderChainID:%d, EventIndex:%d",
				voteInboundID,
				msg.InboundHash,
				msg.SenderChainId,
				msg.EventIndex,
			)
		}
	}
```
**Alternative**

Since `VoteOnInboundBallot` start returning more and more value, we could also return a vote state in the returned values instead
```
VOTE_STATE = NEW | PENDING | FINALIZING | FINALIZED
```

kingpinXD:
We could also refactor to check `IsFinalizedInbound` to be after `finalized` 
Even if two inbounds are created in the same block they would still be finalized one after the other. We can just remove `isNew` in that case 

```
	// If the ballot is not finalized return nil here to add vote to commit state
	if !finalized {
		return &types.MsgVoteInboundResponse{}, nil
	}

[...]

PR #4197, opened 2025-09-13
fix: re-check for finalized ballot when executing inbound vote to create cctx

kingpinXD:
# Description
Closes https://github.com/zeta-chain/node/issues/4108

The pr fixes the issue mentioned but uses a different /simpler approach 
It adds the check for a finalized ballot when executing the vote in additon to the previus check when creating the ballot.
I chose to keep both becuase 
- Check during ballot creation : This prevents creation of unncessary ballots which might increase the state size 
- Check with executing the vote : This prevents the actuall double spending 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents double finalization of inbound cross-chain transactions, avoiding duplicate processing.
  * Improves validation and error handling when inbound events are re-submitted.

* **Refactor**
  * Streamlined the inbound finalization flow to use a dedicated handler for clearer, more reliable processing.

* **Tests**
  * Added scenarios to ensure duplicate inbound submissions are rejected, with coverage for edge cases.


lumtis:
Please don't forget to attach related issue

kingpinXD:
> Please don't forget to attach related issue

Attached issue and opened for review

lumtis (commented):
Seems more complex than it should be to me.
Why not just replacing the condition line 84 with:

```
	if isNew || finalized {
		if k.IsFinalizedInbound(tmpCtx, msg.InboundHash, msg.SenderChainId, msg.EventIndex) {
			return nil, sdkerrors.Wrapf(
				types.ErrObservedTxAlreadyFinalized,
				"%s, InboundHash:%s, SenderChainID:%d, EventIndex:%d",
				voteInboundID,
				msg.InboundHash,
				msg.SenderChainId,
				msg.EventIndex,
			)
		}
	}
```

lumtis:
> > ew || finalized {
> 
> The commit happens between the two checks We can't move the commit before IsNew, as it would save the ballot ( the ones that are not needed).

Why does the commit matter? When we return error, no state is committed anyway.

kingpinXD:
> > > > > ew || finalized {
> > > > 
> > > > 
> > > > The commit happens between the two checks We can't move the commit before IsNew, as it would save the ballot ( the ones that are not needed).
> > > 
> > > 
> > > Why does the commit matter? When we return error, no state is committed anyway.
> > 
> > 
> > When using the cached context, calling commit writes the changes directly into the state db , irrespective of what we return. The discarding of state from failed transactions is handled at a higher level by the sdk
> 
> I don't see why it changes anything since IsFinalizedInbound use the tmpCtx

Okay, you mean something like this? 

```
	if isNew || finalized {
		if k.IsFinalizedInbound(tmpCtx, msg.InboundHash, msg.SenderChainId, msg.EventIndex) {
			return nil, sdkerrors.Wrapf(
				types.ErrObservedTxAlreadyFinalized,
				"%s, InboundHash:%s, SenderChainID:%d, EventIndex:%d",
				voteInboundID,
				msg.InboundHash,
				msg.SenderChainId,
				msg.EventIndex,
			)
		}
	}
	commit()
        
        Still keep this check 
        
	//// If the ballot is not finalized return nil here to add vote to commit state
	if ! finalized {
		return &types.MsgVoteInboundResponse{}, nil
	}
```	
I assumed you meant remove the 	if `! finalized {` check later . Okay I think this can be done . I agree this looks cleaner than callingit twice

skosito (approved):

PR #4202, opened 2025-09-15
fix(backport):  force rescan if inbound vote monitoring fails using a context that can timeout

kingpinXD:
# Description

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/zetaclient/chains/base/observer.go
+++ b/zetaclient/chains/base/observer.go
@@ -526,9 +526,9 @@ func (ob *Observer) handleMonitoringError(
 			}
 		}
 	case <-ctx.Done():
-		logger.Error().
+		logger.Debug().
 			Str(logs.FieldZetaTx, zetaHash).
-			Msg("unable to monitor vote transaction: timeout")
+			Msg("tx succeeded no err from monitor channel before timeout")
 	}
 }
 
--- a/zetaclient/chains/base/observer.go
+++ b/zetaclient/chains/base/observer.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/pkg/errors"
@@ -34,6 +35,9 @@ const (
 	// DefaultBlockCacheSize is the default number of blocks that the observer will keep in cache for performance (without RPC calls)
 	// Cached blocks can be used to get block information and verify transactions
 	DefaultBlockCacheSize = 1000
+
+	// MonitoringErrHandlerRoutineTimeout is the timeout for the handleMonitoring routine that waits for an error from the monitorVote channel
+	MonitoringErrHandlerRoutineTimeout = 5 * time.Minute
 )
 
 // Observer is the base structure for chain observers, grouping the common logic for each chain observer client.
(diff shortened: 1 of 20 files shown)


zetaclient/chains/base/observer.go
```diff
@@ -426,8 +465,14 @@ func (ob *Observer) PostVoteInbound(
 		return "", nil
 	}
 
+	monitorErrCh := make(chan zetaerrors.ErrTxMonitor, 1)
+
+	// ctxWithTimeout is a context with timeout used for monitoring the vote transaction
+	ctxWithTimeout, _ := zctx.CopyWithTimeout(ctx, context.Background(), MonitoringErrHandlerRoutineTimeout)
```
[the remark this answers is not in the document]
kingpinXD: That is the expected behavior, calling the context here would terminated the goroutines using the context which is not desirable

zetaclient/chains/base/observer.go
```diff
@@ -215,15 +224,31 @@ func (ob *Observer) WithLastBlock(lastBlock uint64) *Observer {
 
 // LastBlockScanned get last block scanned (not necessarily caught up with the chain; could be slow/paused).
 func (ob *Observer) LastBlockScanned() uint64 {
+	ob.mu.Lock()
+	defer ob.mu.Unlock()
 	height := atomic.LoadUint64(&ob.lastBlockScanned)
```
renan061, line 229: No need for `atomic.Load` if the method is already locking the mutex.
Same for the method that writes to `lastBlockScanned`.

zetaclient/chains/base/observer.go
```diff
@@ -215,15 +224,31 @@ func (ob *Observer) WithLastBlock(lastBlock uint64) *Observer {
...
+	wasForceReset := ob.forceResetLastScanned
+	ob.forceResetLastScanned = forceResetLastScanned
+
+	// forceResetLastScanned was set to true before; it means the monitoring thread would have updated it
+	// In this case we should not update the last scanned block and just return
+	if wasForceReset && !forceResetLastScanned {
+		return ob, wasForceReset
+	}
+
 	atomic.StoreUint64(&ob.lastBlockScanned, blockNumber)
```
renan061, line 249: Doesn't need to be atomic.

zetaclient/chains/base/observer.go
```diff
@@ -215,15 +224,31 @@ func (ob *Observer) WithLastBlock(lastBlock uint64) *Observer {
...
+	defer ob.mu.Unlock()
 	height := atomic.LoadUint64(&ob.lastBlockScanned)
 	return height
 }
 
 // WithLastBlockScanned set last block scanned (not necessarily caught up with the chain; could be slow/paused).
-func (ob *Observer) WithLastBlockScanned(blockNumber uint64) *Observer {
+// it also set the value of forceResetLastScanned and returns the previous value.
+// If forceResetLastScanned was true before, it means the monitoring thread would have updated it and so it skips updating the last scanned block.
+func (ob *Observer) WithLastBlockScanned(blockNumber uint64, forceResetLastScanned bool) (*Observer, bool) {
```
renan061, line 236: Can rename to `SetLastBlockScanned`.

zetaclient/chains/base/observer.go
kingpinXD: yeah, I tried to keep changes as minimal as possible. If there is no obvious concern, we can keep it as it is.

lumtis:
I will merge since remaining issues are not a concern, we can apply these on the develop branch

PR #4206, opened 2025-09-16
fix: skip writing config file to the filesystem

kingpinXD:
# Description

- Remove restrictions for mainnet, and instead use the value of ` --skip-config-overwrite` flag directly
- Remove logic to write the config back to the filesystem

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

### DIFF
--- a/server/timeout.go
+++ b/server/timeout.go
@@ -59,6 +59,7 @@ func overWriteConfig(cmd *cobra.Command) error {
 			return fmt.Errorf("failed to set server context: %w", err)
 		}
 
+		// Skip writing to config file to prevent error for read-only file systems
 		//err = updateConfigFile(cmd, serverCtx.Config)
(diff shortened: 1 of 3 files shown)


server/timeout.go
```diff
@@ -59,10 +59,10 @@ func overWriteConfig(cmd *cobra.Command) error {
 			return fmt.Errorf("failed to set server context: %w", err)
 		}
 
-		err = updateConfigFile(cmd, serverCtx.Config)
-		if err != nil {
-			return err
-		}
+		//err = updateConfigFile(cmd, serverCtx.Config)
```
lumtis, line 62: Do we know nobody use it?

skosito (approved):

Issue #4074, opened 2025-08-05
Provide more information when Bitcoin deposit fail with invalid memo


lumtis:
**Describe the Issue**

When a Bitcoin deposit failed because of invalid memo, the cctx doesn't contain information allowing to debug the error efficiently.

Example: https://zetachain-athens.blockpi.network/lcd/v1/public/zeta-chain/crosschain/cctx/0xe907b8712607c099a37db56d1fb1ad48cbf89f8bba1c66116384b0dd054b34c7

Only contains the string `invalid memo`.

This stems from this line of code: https://github.com/zeta-chain/node/blob/262abd8263f7c4fcf1d90629586d21acee1cafd9/x/crosschain/keeper/cctx_gateway_zevm.go#L41

We should provide more useful information:

- Allow in the inbound vote object to specify cause of invalid observation with a string
- Attach this string to this error

**Alternative**

Since the change is consensus breaking, a temporary solution to help would be to add more information in the log emitted with invalid observation: https://github.com/zeta-chain/node/blob/f26ceaa2c4ae7001136d5985c4bc1b4e9f089232/zetaclient/chains/bitcoin/observer/inbound.go#L200

Right now the log doesn't contain the error.

PR #4211, opened 2025-09-17
feat: provide error information in CCTX struct when Bitcoin deposit fail

ws4charlie:
# Description

When a Bitcoin deposit failed because of invalid memo or insufficient depositor fee, the `CCTX` doesn't contain informative debug error message.

This PR adds error message to CCTX struct when Bitcoin deposit failed
- with invalid memo
- with insufficient depositor fee

The example error messages will look like
image

image

Closes https://github.com/zeta-chain/node/issues/4074

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

- New Features
  - Surface detailed error messages in cross-chain transaction status for failed Bitcoin deposits (e.g., insufficient depositor fee, invalid memo, invalid revert address).
  - Enhanced inbound error propagation so outbound status includes specific failure reasons.
  - More informative validation for legacy memos (e.g., minimum length) and revert address checks.

- Documentation
  - OpenAPI updated with an errorMessage field for inbound status.
  - Specs updated to include error_message in inbound vote messages.
  - Changelog entry added for the new feature.


proto/zetachain/zetacore/crosschain/tx.proto
```diff
@@ -223,6 +223,10 @@ message MsgVoteInbound {
 
   // confirmation mode used for the inbound
   ConfirmationMode confirmation_mode = 21;
+
+  // error_message carries information about the error that caused non-SUCCESS
+  // status of inbound observation
+  string error_message = 22;
```
kingpinXD, line 229: Do we need this error message field to be part of the msg.Digest() ?
Imo it's safer to just remove it if not needed , we have the status as part of the digest already.

ws4charlie: I think this is a good point. The observation status in the `MsgVoteInbound` already indicates the category of the failure. The only side effect is that the final `errorMessage` is determined by the finalizing vote ( the 6th vote).

kingpinXD: Yeah, the message itself is just for debugging, so it might not be a problem if we use it for the final message 
I can't also see any problems with keeping it in the digest, since we directly abort the cctx anyway. Up to you either seems fine

lumtis: Even when the cctx is aborted/reverted the risk of double observation remain.

lumtis (changes requested):
Need to keep it excluded from digest

proto/zetachain/zetacore/crosschain/tx.proto
ws4charlie: Excluded it from digest https://github.com/zeta-chain/node/pull/4211/commits/dbb0bbaf895ac991ed6b560eea97564ddea0a978

lumtis (approved):

kingpinXD (approved):

PR #4227, opened 2025-09-23
feat: add support for ALT

skosito:
# Description

Solana PR: https://github.com/zeta-chain/protocol-contracts-solana/pull/128

Regarding changes in solana-go, it is basically adding instructions to create and extend ALT, which is just used in e2e tests, so change is safe to use until upstream PR is merged

This PR adds additional encoding for withdraw and call payload:

```
type ExecuteMsgALT struct {
	AltAddress       [32]byte
	WriteableIndexes []uint8
	Data             []byte
}
```

old one still is supported as legacy, but we can remove it if it's practically not used, but just for illustration:

```
type ExecuteMsg struct {
	Accounts []AccountMeta
	Data     []byte
}
```

Information we need for ALT to work are ALT address and indexes in ALT of mutable/writable accounts. If there is some better way from UX perspective to encode this information while still maintain ok size in terms of solidity encoding, please leave suggestions in PR.

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests


skosito:
> https://github.com/zeta-chain/toolkit/blob/26abfdaa7dd246cb986ad5bfff7b83c32cab7aae/packages/client/src/solanaEncode.ts#L87-L90

yes, should be way easier in a sense there is no nested tuple

a bit more complex ux because caller will have setup ALT and pass mutable indexes, but i dont think there is much simpler way to achieve this

e2e/e2etests/test_solana_withdraw_and_call.go
```diff
@@ -21,10 +21,16 @@ import (
 // in connected program pda and account provided in remaining accounts to demonstrate that lamports
 // can be moved to accounts in connected program as well as gateway program
 func TestSolanaWithdrawAndCall(r *runner.E2ERunner, args []string) {
-	require.Len(r, args, 1)
+	require.LessOrEqual(r, len(args), 3)
```
ws4charlie, line 24: do you want to make it more strict, e.g. `require.True(r, len(args) == 1 || len(args) == 3)`?

e2e/runner/solana.go
```diff
@@ -813,3 +815,121 @@ func (r *E2ERunner) WaitAndVerifySPLBalanceChange(
...
+// used to test ALTs with large amount of accounts
+func (r *E2ERunner) SetupTestALTWithRandomWallets(accounts []solana.PublicKey) (solana.PublicKey, []solana.PublicKey) {
+	privkey := r.GetSolanaPrivKey()
+
+	recentSlot, err := r.SolanaClient.GetSlot(r.Ctx, rpc.CommitmentFinalized)
+	require.NoError(r, err)
+
+	// prefunded random wallets used for testing big number of accounts in payload
+	randomWallets := []solana.PublicKey{
+		solana.MustPublicKeyFromBase58("4C2kkMnqXMfPJ8PPK5v6TCg42k5z16f2kzqVocXoSDcq"),
```
ws4charlie, line 829: The maximum addresses accepted by a non-ALT transaction is `32`, and this is because Solana enforces a `1232 bytes` cap on its transaction size. To verify that ALT works in e2e test and make live network test easier, could we use fewer addresses for test, e.g. `33` or `34` or `35`? And, if we we devOps team to send funds to these static addresses, we could declare address list in a `var` block so they see it clearly.

zetaclient/chains/solana/signer/signer.go
```diff
@@ -247,37 +247,45 @@ func (signer *Signer) TryProcessOutbound(
...
 }
 
-// signTx creates and signs solana tx containing provided instruction with relayer key.
+// signTx creates and signs a Solana tx containing the provided instruction with the relayer key.
+// If `alt` is non-nil and `addrs` is non-empty, the transaction will include an address lookup table.
 func (signer *Signer) signTx(
 	ctx context.Context,
 	inst *solana.GenericInstruction,
 	limit uint64,
+	alt *solana.PublicKey,
```
ws4charlie, line 256: Do we really need this `alt` argument as we can infer the availability of ALT from `len(addrs) > 0`?

ws4charlie (commented):
the changes look good, I left a few comments if you could take a look

e2e/runner/solana.go
skosito: my idea was that we will create ALT out of e2e tests for live networks and just provide ALT for test in argument so we dont have to create new ALT for every run

that would require script to create and extend ALT, wdyt is better approach?

zetaclient/chains/solana/signer/signer.go
skosito: yea it is needed here:

```
opts = append(opts, sol.TransactionAddressTables(map[sol.PublicKey]sol.PublicKeySlice{
			*alt: addrs,
		}))
```

zetaclient/chains/solana/signer/execute.go
```diff
@@ -43,12 +46,91 @@ func (signer *Signer) prepareExecuteTx(
...
+	}
+
+	alt, err := addresslookuptable.GetAddressLookupTableStateWithOpts(
+		ctx,
+		signer.client.(*rpc.Client),
+		*msg.ALTAddress(),
+		&rpc.GetAccountInfoOpts{Commitment: rpc.CommitmentProcessed},
+	)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "cannot get alt")
```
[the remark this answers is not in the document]
skosito: need to double check but since this is part of `prepareExecuteTx` it will return and wont reach broadcast and fail, so should be retried

zetaclient/chains/solana/signer/execute.go
```diff
@@ -43,12 +46,91 @@ func (signer *Signer) prepareExecuteTx(
...
+
+	writableSet := make(map[int]struct{}, len(msg.Alt.WriteableIndexes))
+	for _, j := range msg.Alt.WriteableIndexes {
+		writableSet[int(j)] = struct{}{}
+	}
+
+	for i, a := range alt.Addresses {
+		_, isWritable := writableSet[i]
+		remainingAccounts = append(remainingAccounts, &solana.AccountMeta{
+			PublicKey:  solana.PublicKey(a),
```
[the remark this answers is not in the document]
skosito: not sure i understand this part, could you please explain why it is a problem if ALT contains PDA? i think user can put whatever is needed inside ALT for specific connected program

zetaclient/chains/solana/signer/execute.go
```diff
@@ -43,12 +46,91 @@ func (signer *Signer) prepareExecuteTx(
...
+		*msg.ALTAddress(),
+		&rpc.GetAccountInfoOpts{Commitment: rpc.CommitmentProcessed},
+	)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "cannot get alt")
+	}
+
+	writableSet := make(map[int]struct{}, len(msg.Alt.WriteableIndexes))
+	for _, j := range msg.Alt.WriteableIndexes {
+		writableSet[int(j)] = struct{}{}
```
ws4charlie, line 117: Do we need to validate the value of `j` is in the range of `[0, len(alt.Addresses))`?  Otherwise it's a invalid tx?

skosito:
recreated here: https://github.com/zeta-chain/node/pull/4266

zetaclient/chains/solana/signer/execute.go
skosito: it is up to user, it will fail and be replaced with fallback increment nonce

otherwise if we stop it here this will get retried since it is in prepare part, not sure we have cancellation in this part for solana yet, just increment nonce fallback need to check

PR #4303, opened 2025-10-07
chore: increase upgrade test height

lumtis:
# Description

Upgrade tests are failing a lot in the CI because the first tests pass complete after the upgrade height.

This PR increase the height from 240 to 260 to give the time for tests to complete. This is a change we already did on the main branch to fix this problem in the past.


lumtis:
The upgrade tests complete in 19min which I think is an reasonable time for now, so I think to leave it to 280.

We should be able to disable the upgrade test once we have the zetaclient specific upgrade tests implemented.

kingpinXD (approved):

Issue #4260, opened 2025-09-26
Create a CI actions to perform ZetaClient and ZetaCore releases separately


lumtis:
In the current process we perform release of zetacore and zetaclient simultaneously from a release/ branch.

As per the docs for the new git workflow, we should create new actions allowing to perform separate release.

zetaclient from `release/zetaclient/`
zetacore from `release/zetacore/`

The CI to run will mostly be the same with the exception that we should add more check for consensus breaking change for ZetaClient

Issue #4327, opened 2025-10-13
Remove legacy publish release workflow


lumtis:
Followup of https://github.com/zeta-chain/node/issues/4258

We will keep the legacy publish release for now in case a patch release need to be made with the legacy workflow.

Once the first ZetaClient-only and ZetaCore-only releases are performed, we should be able to only rely on the new workflow and remove the following files:

```
.goreleaser.yaml
.github/workflows/publish-release.yml
```

PR #4329, opened 2025-10-13
ci: create a separate release workflow for zetaclient and zetacore

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/4260

zetae2e release is currently associated with zetaclientd for simplification since we should always do zetaclientd release associated with new zetacored release, even in the new workflow.

Note: not removing the older publish release for now in case we still need it for a patch. This will be addresses in [Remove legacy publish release workflow](https://github.com/zeta-chain/node/issues/4327)

## Summary by CodeRabbit

* **New Features**
  * Official release artifacts for ZetaCore and ZetaClient across Linux, macOS, and Windows (amd64/arm64).
  * Automatic changelogs, checksums, and provenance attestations attached to releases.
  * Support for pre-releases and robust tag handling.

* **Chores**
  * Added automated release workflows for publishing and follow-up TypeScript publishing.
  * Introduced Makefile targets for snapshot and full releases to streamline builds.
  * Adjusted version propagation in the release pipeline for consistency.

### DIFF
--- a/.github/workflows/release-template.yml
+++ b/.github/workflows/release-template.yml
@@ -62,7 +62,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build release snapshot
         if: inputs.skip_checks != true
-        run: make ${{ inputs.make_snapshot_target }}
+        env:
+          MAKE_SNAPSHOT_TARGET: ${{ inputs.make_snapshot_target }}
+        run: make "$MAKE_SNAPSHOT_TARGET"
 
   publish-release:
     permissions:


.github/workflows/publish-release-zetacore.yml
```diff
@@ -0,0 +1,128 @@
+name: Publish ZetaCore Release
```
skosito: if there is a duplicate between these 2 workflows, would it make sense to extract some common parts, like having more generic publish-release workflow and something like:

```
name: Publish ZetaClient Release
on:
  workflow_dispatch:
    inputs:
      version:
        description: 'Version for ZetaClient Release'
        required: true
      skip_checks:
        type: boolean
        default: false
      skip_release:
        type: boolean
        default: false

jobs:
  publish:
    uses: ./.github/workflows/publish-release.yml
    with:
      component: zetaclient
      version: ${{ inputs.version }}
      skip_checks: ${{ inputs.skip_checks }}
      skip_release: ${{ inputs.skip_release }}
    secrets: inherit

```

and similar to zetacore

.goreleaser.yaml
```diff
@@ -12,7 +12,7 @@ env:
   - CC_darwin_arm64=oa64-clang
   - CXX_darwin_arm64=oa64-clang++
   - CC_darwin_amd64=o64-clang
-  - CXX_darwin_amd64=o64-clang+
+  - CXX_darwin_amd64=o64-clang++
```
skosito: was this not working before? wondering how we didnt spot something failing, nice catch

do we still need this file? i see there is an issue to remove legacy makefile cmds, i guess this one will be removed then as well

kingpinXD (approved):
lgtm

.github/workflows/publish-release-zetaclient.yml
```diff
@@ -0,0 +1,134 @@
...
+        run: |
+          gh release upload "${GITHUB_TAG_VERSION}" "${{ steps.attestation.outputs.bundle-path }}"
+
+      - name: Clean Up Workspace
+        if: always()
+        shell: bash
+        run: sudo rm -rf * || echo "failed to cleanup workspace please investigate"
+
+  publish-typescript:
+    needs: publish-release
```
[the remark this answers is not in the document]
morde08: +1 I don't think we need to publish-typescript for a Zetaclient release

PR #4357, opened 2025-10-20
test(e2e): json rpc checks during e2e and upgrade tests

skosito:
# Description

Purpose is to check if some rpc is broken after upgrade, since we had couple of issues recently that could've been spotted with tests like this, especially when rpc method returns error or empty results. There is no need to test fields in details, for that we have unit tests for rpcs - this gives flexibility to also run it on live networks. Some of requirements:

- run on localnet or live network
- if run on localnet, should be deterministic, so added a new user for this that submits deposit and withdraw, so we can get inbounds and outbounds submitted by this user - most importantly, this will enable checking all txs (before and after upgrade is hit in upgrade tests)
- on live network, list of tx hashes can be provided, so it can sanity check bunch of predefined hashes
- i think it is fine that this is part of e2e tests, everything is set up already 

# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

* **Tests**
  * Added JSON-RPC end-to-end test suite validating transaction receipts, block data retrieval, and debug transaction traces for enhanced test coverage.


e2e/e2etests/e2etests.go
```diff
@@ -1478,12 +1479,19 @@ var AllE2ETests = []runner.E2ETest{
...
 		TestOpcodesName,
 		"test opcodes support in ZEVM",
 		[]runner.ArgDefinition{},
 		TestOpcodes,
 	),
+	runner.NewE2ETest(
+		TestRPCName,
+		"test json rpc support in ZEVM",
+		[]runner.ArgDefinition{
+			{Description: "tx hashes", DefaultValue: ""},
```
lumtis: Could we give more details about this argument? And eventually the test description as well, as it is unclear the association between `test json rpc support in ZEVM` and `tx hashes`

cmd/zetae2e/local/local.go
```diff
@@ -557,6 +557,13 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 		))
 	}
 
+	eg.Go(rpcTestRoutine(
```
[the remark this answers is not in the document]
skosito: no it is just one deposit and withdraw to seed transactions, it is in separate go routine so it should be fine to always run it, it finishes much faster than whole thing

e2e/e2etests/e2etests.go
skosito: i wanted to have a way to just provide some info from live networks or even locally with zetae2e run to check transactions directly

eg, we could have on nightly run one tx from before upgrade, one tx after upgrade on testnet, and this test would work

e2e/e2etests/test_rpc.go
```diff
@@ -0,0 +1,138 @@
...
+// If no transaction hashes are provided via args, it seeds test transactions and discovers existing
+// transactions from this user via CCTX queries (both ZEVM and EVM chains). This enables testing
+// RPC methods against transactions submitted before upgrades to detect any regressions.
+func TestRPC(r *runner.E2ERunner, args []string) {
+	r.Logger.Info("starting JSON-RPC tests")
+
+	rpcWrapper := NewZEVMRPC(r.ZEVMClient)
+
+	txHashes := []string{}
+	if len(args) == 1 && args[0] != "" {
```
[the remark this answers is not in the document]
skosito: we can remove this, i thought it might be useful to have bunch of testnet or mainnet transactions we want to specify to run rpc tests, eg: we release something on testnet, we can specify bunch of transactions before upgrade and after upgrade, we won't be able to seed those pre-upgrade transactions anymore when upgrade hits

cmd/zetae2e/local/rpc.go
```diff
@@ -0,0 +1,59 @@
...
+			runner.WithZetaTxServer(deployerRunner.ZetaTxServer),
+		)
+		if err != nil {
+			return err
+		}
+
+		rpcTestRunner.Logger.Print("🏃 starting RPC tests")
+		startTime := time.Now()
+
+		if len(testNames) == 0 {
```
ws4charlie: nit: the `rpcTestRunner.RunE2ETests` already handles empty list internally. Do we really need this special handling?

e2e/rpcwrapper/rpc_wrapper.go
```diff
@@ -0,0 +1,233 @@
...
+	"encoding/json"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// ZEVMRPC provides thin wrappers over eth JSON-RPC using CallContext, plus helpers to parse common responses.
+type ZEVMRPC struct {
```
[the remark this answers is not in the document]
ws4charlie: just curious, do you know which API in `ethclient` isn't working for zevm?

e2e/e2etests/test_rpc.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// perform the withdraw
+	tx = r.ETHWithdraw(r.EVMAddress(), amount, gatewayzevm.RevertOptions{OnRevertGasLimit: big.NewInt(0)})
+
+	// wait for the cctx to be mined
+	cctx2 := utils.WaitCctxMinedByInboundHash(r.Ctx, tx.Hash().Hex(), r.CctxClient, r.Logger, r.CctxTimeout)
+	r.Logger.CCTX(*cctx2, "withdraw")
+	require.Equal(r, crosschaintypes.CctxStatus_OutboundMined, cctx2.CctxStatus.Status)
+
+	cctxsRes, err := r.CctxClient.CctxAll(r.Ctx, &crosschaintypes.QueryAllCctxRequest{})
```
ws4charlie: In the live networks, does it mean fetching the 1st page of 100 CCTXs and filtering out those triggered by one of `ZEVMAuth.From.Hex()` and  `EVMAuth.From.Hex()`? I'm not sure how many relevant CCTXs we can get in this case, especially when the CCTX traffic is high in ZetaChain.

Do we really have to filter by address?

e2e/e2etests/test_rpc.go
```diff
@@ -0,0 +1,156 @@
...
+		blockByHash, err := rpcWrapper.EthGetBlockByHash(r.Ctx, common.HexToHash(receipt.BlockHash), false)
+		require.NoError(r, err)
+		require.NotNil(r, blockByHash)
+		require.Equal(r, receipt.BlockHash, blockByHash.Hash)
+		require.NotEmpty(r, blockByHash.Number)
+		require.NotEmpty(r, blockByHash.ParentHash)
+		require.NotEmpty(r, blockByHash.Timestamp)
+		require.GreaterOrEqual(r, len(blockByHash.Transactions), 1)
+
+		traceTx, err := rpcWrapper.DebugTraceTransaction(r.Ctx, common.HexToHash(txHash))
```
ws4charlie: Does `DebugTraceTransaction` API relies on the state history? If yes, it has to be ran either on a archive node endpoint or within a given limit of state history length (by blocks).

e2e/e2etests/test_rpc.go
```diff
@@ -0,0 +1,156 @@
...
+
+	zetaChainID, err := r.ZEVMClient.ChainID(r.Ctx)
+	require.NoError(r, err)
+
+	evmChainID, err := r.EVMClient.ChainID(r.Ctx)
+	require.NoError(r, err)
+
+	txHashes := []string{}
+	// this will also fetch txs submitted by this user before upgrade
+	for _, cctx := range cctxsRes.CrossChainTx {
```
kingpinXD: why cant we do this right after we wait for the cctx to be mined ?

skosito: this is kinda main point i wanted to cover with this PR, but i see it is confusing, not sure if there is better way to achieve this

e2e/e2etests/test_rpc.go
ws4charlie: if not intended for live network at this moment, it's should be fine.

kingpinXD (approved):

Issue #4247, opened 2025-09-24
Document all gas related parameters and mechanisms


CharlieMc0:
**Describe the Issue**
We need to have all the gas parameters documented so they are easy to view in one place. 

**Expected Outcome**
On the docs page we should have a page that outlines all the various gas parameters and where to adjustment them. This should include dynamic on-chain parameters as well as any hard coded values. 

Ideally the on-chain parameters should be queried automatically from an RPC node. 

It should describe all the gas mechanisms in play, for EVM, Cosmos, cross-chain, Solana, SUI, etc.

PR #4394, opened 2025-10-29
docs: add documentation for gas and fees with ZetaChain

lumtis:
# Description

Closes https://github.com/zeta-chain/node/issues/4247

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guide on gas and fees mechanisms in ZetaChain, covering fee calculations across multiple blockchain contexts and supported chains (EVM, Bitcoin, Solana, Sui, TON), with details on deposits, withdrawals, and transaction fee handling.

### DIFF
--- a/docs/zetacore/gas_and_fees.md
+++ b/docs/zetacore/gas_and_fees.md
@@ -280,7 +280,7 @@ The logic to calculate this depositor fee can be found in the [following page](h
 
 [Solana Fee Documentation](https://solana.com/docs/core/fees).
 
-No additional fees are charged for deposits from the Solana network to ZetaChain.
+An additional deposit fee of 2,000,000 Lamports (`0.002` SOL) is charged for deposits from the Solana network to ZetaChain, on top of the standard Solana network transaction fees.
 
 ### Sui
 


docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+* **Connected Chains Withdraw Fees**: The calculation and payment of fees when a user initiates an outgoing cross-chain transaction from ZetaChain to a connected chain.
+
+* **Connected Chains Deposit Fees**: The calculation and payment of fees when a user initiates an incoming cross-chain transaction from a connected chain to ZetaChain.
+
+* **Cosmos Transaction Fees**: As a Layer 1 blockchain built with Cosmos SDK, sending direct Cosmos transactions on ZetaChain involves using the Cosmos gas mechanism for fee calculation and payment.
+
+## ZetaChain EVM Transaction Fees
+
+When users interact with the ZetaChain EVM, for any operations, they will pay fees using the standard EVM fee mechanism.
+More information can be found at the [EVM Fees Documentation](https://ethereum.org/developers/docs/gas/).
```
skosito: i think we should double check between cosmos evm and eth to be sure it is 1:1 mapping in behavior, if not already checked

docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+In addition to the Bitcoin network fees, a depositor fee is deducted from the deposited amount.
+The depositor fee is charged to cover the cost of spending the deposited UTXO in the future.
+
+The logic to calculate this depositor fee can be found in the [following page](https://www.zetachain.com/docs/developers/chains/bitcoin#fees).
+
+### Solana
+
+[Solana Fee Documentation](https://solana.com/docs/core/fees).
+
+No additional fees are charged for deposits from the Solana network to ZetaChain.
```
skosito: there is additional deposit fee hardcoded for solana, i think other chains might have something similar, there was discussion about ton recently

docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+
+For these system messages, the ZetaClient process uses fixed gas limit values defined in [this file](https://github.com/zeta-chain/node/blob/develop/zetaclient/zetacore/constant.go).
+
+### Gas Limit for Inbound Voting
+
+The gas limit used for inbound voting has some specificities.
+There is currently no distinctive message between voting and executing inbound—the vote message reaching sufficient quorum will execute the inbound automatically.
+However, the execution of the inbound might trigger a smart contract call that requires an increased amount of gas.
+Since higher gas limits result in higher fees in Cosmos transactions, the following mechanism is used to avoid requiring high gas limits for each vote message:
+
```
renan061: This paragraph left me a bit confused. A suggestion:

> The gas limit used for inbound voting has some specificities, given that the execution of the inbound might trigger a smart contract call that requires an increased amount of gas.
> 
> Because the vote message that reaches sufficient quorum automatically executes the inbound, there is currently there is no way to determine if a specific vote will be the one that causes the inbound execution.
> 
> Thus, there is no way to predict if a vote message requires the increased amount of gas associated with the execution of a smart contract call.
> 
> Since higher gas limits result in higher fees in Cosmos transactions, the following mechanism is used to avoid requiring high gas limits for each vote message:

What do you think?

docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+The meaning of gas price and gas limit depends on the connected chain.
+
+These differences are summarized below.
+
+### EVM
+
+For EVM-compatible chains, ZetaChain relies on the standard Ethereum gas mechanism.
+More information about Ethereum gas can be found in the [Ethereum Gas Documentation](https://ethereum.org/developers/docs/gas/).
+
+The withdraw fee for EVM chains is calculated using the gas price and gas limit parameters:
```
kingpinXD: Should we specify that this gas limit is set by the user, and different from the gas limit for the, zevm transaction creating the withdraw

docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+```
+Withdraw Fee = Gas Price × Gas Limit
+```
+
+The gas price represents the cost per unit of gas on the connected EVM chain,
+while the gas limit represents the maximum amount of gas allocated for executing the transaction on that chain.
+
+**Important Consideration**: The gas limit is fully consumed regardless of whether the cross-chain call uses all the allocated gas.
+In the current architecture, any remaining unused gas is sent to a gas stability pool.
```
kingpinXD: a portion is refunded to the user 
https://github.com/zeta-chain/node/blob/3daaba6b8725ba25a57e648f9fb38e549a2a57ff/docs/zetacore/managment_for_unused_gas_fee.md

docs/zetacore/gas_and_fees.md
```diff
@@ -0,0 +1,334 @@
...
+* **Connected Chains Deposit Fees**: The calculation and payment of fees when a user initiates an incoming cross-chain transaction from a connected chain to ZetaChain.
+
+* **Cosmos Transaction Fees**: As a Layer 1 blockchain built with Cosmos SDK, sending direct Cosmos transactions on ZetaChain involves using the Cosmos gas mechanism for fee calculation and payment.
+
+## ZetaChain EVM Transaction Fees
+
+When users interact with the ZetaChain EVM, for any operations, they will pay fees using the standard EVM fee mechanism.
+More information can be found at the [EVM Fees Documentation](https://ethereum.org/developers/docs/gas/).
+
+ZetaChain currently uses Go-Ethereum `v1.15.11` and supports both legacy and EIP-1559 transaction types.
```
CharlieMc0: Is it possible to pull this dynamically? It's one more thing we have to update each time we update cosmos-evm. 

If it'll take more than a few minutes to pull automatically then lets just skip it.

docs/zetacore/gas_and_fees.md
lumtis: I changed the paragraph slightly but prefer to keep it a bit simpler than the siggestion

docs/zetacore/gas_and_fees.md
lumtis: Not implemented yet on live network, we'll have to revisit later

skosito (approved):

docs/zetacore/gas_and_fees.md
lumtis: calloptions.Gaslimit is optional

kingpinXD (approved):
Lgtm

docs/zetacore/gas_and_fees.md
lumtis: I think it will be ok for now, might revisit in the future

PR #4514, opened 2026-01-02
fix: use zeta height to calculate the artificial tss keysign height

ws4charlie:
# Description

The `block height` is the only variable used to pick TSS keysign leader, see method: [PickLeader](https://github.com/zeta-chain/go-tss/blob/892e71030254532848c0207dee48de20467005a2/p2p/protocol.go#L56).

If we pass a fixed artificial keysign height to function `SignBatch`, we potentially face constant keysign failure if that leader happens to be offline. This PR uses `zetaHeight` as one of the variables to derive artificial keysign height, instead of only using a fixed batch number.

# How Has This Been Tested?

- [x] Tested CCTX in localnet
- [x] Go unit tests
- [ ] Go integration tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced EVM chain artificial height calculation for TSS keysign operations by incorporating Zeta height as a primary calculation factor. This improvement ensures more accurate height determination during cross-chain key signing operations for EVM-compatible blockchain networks.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/zetaclient/chains/base/signer_batch_info.go
+++ b/zetaclient/chains/base/signer_batch_info.go
@@ -2,7 +2,9 @@ package base
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/zeta-chain/node/pkg/constant"
 	mathpkg "github.com/zeta-chain/node/pkg/math"
 )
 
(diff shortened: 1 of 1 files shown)


zetaclient/chains/base/signer_batch_info.go
```diff
@@ -13,6 +11,10 @@ const (
 	// signing a 10-digest batch takes about 3~4 seconds on average, and
 	// increasing batch size won't result in significant performance boost
 	batchSize = 10
+
+	// zetaHeightDivisor is the divisor to map zeta height to keysign height bucket.
+	// 10 blocks (apprx. 20 seconds) is used to avoid keysign height drifting too fast.
+	zetaHeightDivisor = 10
```
kingpinXD: Would it make sense to make this a functiom of blocktime ?

if the block time is changed to 1s this would mean a new signer is picked every 10s. I dont think its a prblem , but just wanted to point it out

kingpinXD (approved):

zetaclient/chains/base/signer_batch_info.go
ws4charlie: good point. It's changed to time https://github.com/zeta-chain/node/pull/4514/commits/95c37fde1ac910f32d4d7d2d0f86e696cc4f81ad

PR #4515, opened 2026-01-03
feat: disable v2 trasanctions for coin-type zeta

kingpinXD:
# Description

- Adds a flag to dynamically switch ZETA V2 Flows on and off .
-  Updates E2E tests to use the flag to verify behaviour.

 | Flow               | Behavior                                 | User Funds                                 | Track Transaction                                                  |
  |--------------------|------------------------------------------|--------------------------------------------|--------------------------------------------------------------------|
  | V2 ZETA Deposit    | SetAbort() - CCTX aborted, no processing | ZETA locked in connector on external chain | Query CCTX by inbound hash to see error message and aborted status |
  | V2 ZETA Withdrawal | Transaction reverts                      | User keeps ZETA on ZetaChain               | No on-chain record exists; zetacore logs contain error details     |

- Note: Aborted transactions can be manually refunded, if needed. I chose not to do it automatically, as an automatic refund means a deposit to a zEVM address. Just uses the abort address instead of the receiver 
# How Has This Been Tested?

- [ ] Tested CCTX in localnet
- [ ] Go unit tests
- [ ] Go integration tests

---

> [!NOTE]
> Adds a policy-controlled switch for V2 ZETA gateway flows and wires it through node, APIs, and tests.
> 
> - Introduces `isV2ZetaEnabled` in `CrosschainFlags` and new `MsgUpdateV2ZetaFlows` (proto, gRPC, OpenAPI, TS bindings) to enable/disable V2 ZETA flows
> - Authority migration and defaults updated; authorization added for `MsgUpdateV2ZetaFlows`; observer keeper exposes `IsV2ZetaEnabled`
> - Crosschain enforcement:
>   - ZEVM inbound: abort V2 ZETA deposits with `ErrZetaThroughGateway` when disabled
>   - V2 ZETA withdrawals rejected/ignored when disabled, with log/context
> - E2E: new `--v2-zeta-flows` flag and `start-e2e-v2ZETA-test`; tests conditionally assert success vs. aborted/reverted paths; runner helpers to enable/check flag
> - Docs/changelog updated; new error `ErrZetaThroughGateway`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b1e477328d0b2128d0319a34e583e5518ad01cd. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a toggleable V2 ZETA flows feature with command-line flag support
  * Introduced configuration capability to enable/disable V2 ZETA flows

* **Behavior Changes**
  * ZETA deposits and withdrawals now conditionally execute based on V2 flows enablement status
  * Operations abort gracefully with descriptive error messages when V2 flows are disabled

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -35,6 +35,7 @@ The `zetacored` binary must be upgraded to trigger chain parameters data migrati
 * [4490](https://github.com/zeta-chain/node/pull/4490) - allow object for `tracerConfig` in the zevm debug APIs
 * [4511](https://github.com/zeta-chain/node/pull/4511) - false mempool congested warning
 * [4509](https://github.com/zeta-chain/node/pull/4509) - use outbound schedule interval in sui cctx scheduling
+* [4514](https://github.com/zeta-chain/node/pull/4514) - use Zeta height as a factor to calculate the EVM chain artificial height for TSS keysign
 
 ### Tests
 
--- a/zetaclient/chains/base/signer_batch_info.go
+++ b/zetaclient/chains/base/signer_batch_info.go
@@ -2,9 +2,9 @@ package base
 
 import (
 	"fmt"
+	"time"
 
-	"github.com/pkg/errors"
-
+	"github.com/zeta-chain/node/pkg/constant"
 	mathpkg "github.com/zeta-chain/node/pkg/math"
 )
 
@@ -13,6 +13,10 @@ const (
 	// signing a 10-digest batch takes about 3~4 seconds on average, and
 	// increasing batch size won't result in significant performance boost
 	batchSize = 10
+
+	// keysignHeightBucket is the bucket size to map zeta height to keysign height bucket.
+	// 20 seconds is used to avoid keysign height drifting too fast.
+	keysignHeightBucket = int64(20 * time.Second / constant.ZetaBlockTime)
 )
 
 // TSSKeysignInfo represents a record of TSS keysign information.
@@ -118,23 +122,24 @@ func (b *TSSKeysignBatch) ContainsNonce(nonce uint64) bool {
 
 // KeysignHeight calculates an artificial keysign height tweaked by chainID.
 // This is used to uniquely identify TSS keysign request across all chains.
-func (b *TSSKeysignBatch) KeysignHeight(chainID int64) (uint64, error) {
-	batchNumber := b.BatchNumber()
-	if batchNumber >= mathpkg.MaxPairValue {
-		return 0, errors.New("batch number is too large")
+func KeysignHeight(chainID int64, zetaHeight int64) (uint64, error) {
+	heightBucket := zetaHeight / keysignHeightBucket
+	if zetaHeight <= 0 || heightBucket >= mathpkg.MaxPairValue {
+		return 0, fmt.Errorf("invalid zeta height: %d", zetaHeight)
 	}
 
 	if chainID <= 0 || chainID > mathpkg.MaxPairValue {
 		return 0, fmt.Errorf("invalid chain ID: %d", chainID)
 	}
 
-	// use batch number as unique identifier, added 1 to avoid 0
+	// use height bucket as unique identifier, added 1 to avoid 0
 	// #nosec G115 - checked in range
-	identifier := uint32(batchNumber + 1)
+	identifier := uint32(heightBucket + 1)
 
 	// #nosec G115 - checked in range
 	chainID32 := uint32(chainID)
 
+	// create artificial height using pairing function
 	return mathpkg.CantorPair(identifier, chainID32), nil
 }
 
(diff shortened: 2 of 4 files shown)


x/crosschain/keeper/cctx_gateway_zevm.go
```diff
@@ -44,6 +45,14 @@ func (c CCTXGatewayZEVM) InitiateOutbound(
 		return newCCTXStatus, nil
 	case types.InboundStatus_SUCCESS:
 		// process the deposit normally
+		if config.CCTX.InboundParams.CoinType == coin.CoinType_Zeta &&
```
kingpinXD: This is sligthy different from the withdraw , 
For a v2 withdraw the tx is directly reverted with no balance change for the user , but in this case we abort the tx 

We could also ignore the deposit and not create the cctx at all but ignring it on the zetaclient 

```
	if msg.CoinType == coin.CoinType_Zeta {
			app, err := zctx.FromContext(ctx)
			if err != nil {
				return lastScanned - 1, errors.Wrap(err, "unable to get app context")
			}
			if !app.IsV2ZetaEnabled() {
				ob.Logger().Inbound.Info().
					Stringer(logs.FieldTx, event.Raw.TxHash).
					Str("coin_type", coin.CoinType_Zeta.String()).
					Msg("V2 ZETA deposit ignored: V2 ZETA flows are disabled")
				continue
			}
		}
```


However that would mean that funds get stuck and the user looses whatever amount they deposited 

I think aborting makes more sense as it gives us an option to refund if we want to 






Second consideration is tracebility 
Creating an abort provides complete tracebility 

The withdraw is not trackeble right 
Failing a withdrawal does emit an event
https://github.com/zeta-chain/zeta-node/blob/88f4df439231873b70352b6b4ae9fbf9cfe4c468/x/crosschain/keeper/evm_hooks.go#L98-L98
I think this was added to track such failed withdrawals, however, the logic does not work, since after emitting the event, we are returning an error.

This essentially rolls back the logs
Code from ApplyTransaction
```
		if err = k.PostTxProcessing(tmpCtx, signerAddr, *msg, receipt); err != nil {
			// If hooks returns an error, revert the whole tx.
			res.VmError = errorsmod.Wrap(err, "failed to execute post transaction processing").Error()
			k.Logger(ctx).Error("tx post processing failed", "error", err)
			// If the tx failed in post processing hooks, we should clear all log-related data
			// to match EVM behavior where transaction reverts clear all effects including logs
			res.Logs = nil
			receipt.Logs = nil
			receipt.Bloom = ethtypes.Bloom{} // Clear bloom filter
		} else if commitFn != nil {
```

We emit a log when rolling back also which seems like the easiest way to trace at the moment unless we can do a quick refactor for the V2 contracts to add a flag whcih can reject creating V2 withdraws

cmd/zetae2e/local/local.go
```diff
@@ -278,6 +281,10 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 		}))
 	}
 
+	if v2ZETAFlows {
```
skosito: is this going to be removed in future? wondering if this is becoming too granular with all these flags, but i guess in current setup of e2e tests there is no better way to do this

e2e/e2etests/test_zeta_deposit.go
```diff
@@ -17,19 +17,26 @@ func TestZetaDeposit(r *runner.E2ERunner, args []string) {
...
-
 	// wait for the cctx to be mined
 	cctx := utils.WaitCctxMinedByInboundHash(r.Ctx, tx.Hash().Hex(), r.CctxClient, r.Logger, r.CctxTimeout)
 	r.Logger.CCTX(*cctx, "zeta_deposit")
-	utils.RequireCCTXStatus(r, cctx, crosschaintypes.CctxStatus_OutboundMined)
 
-	newBalance, err := r.ZEVMClient.BalanceAt(r.Ctx, receiverAddress, nil)
-	require.NoError(r, err)
-	require.Equal(r, new(big.Int).Add(oldBalance, amount), newBalance)
+	if r.IsV2ZETAEnabled() {
```
skosito: would this work without side effects or hidden issues if these tests are run on live network?

ws4charlie (approved):
LGTM.  Does this PR need a light security check?

x/observer/keeper/msg_server_update_v2_zeta_flows.go
```diff
@@ -0,0 +1,40 @@
...
+	// check permission
+	err := k.GetAuthorityKeeper().CheckAuthorization(ctx, msg)
+	if err != nil {
+		return nil, errors.Wrap(authoritytypes.ErrUnauthorized, err.Error())
+	}
+
+	// check if the value exists,
+	// if not, set the default value for the Inbound and Outbound flags only
+	flags, isFound := k.GetCrosschainFlags(ctx)
+	if !isFound {
```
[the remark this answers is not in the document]
ws4charlie: okay, it's up to you and not a blocker

PR #4518, opened 2026-01-06
chore: release reforge 

kingpinXD:
# Description

This PR updates main with all changes for the Reforge release 

Important zetacored changes for review 

- Stability Pool Gas Refund Mechanism

  Setting UserGasFeePaid: When a withdrawal CCTX is created, UserGasFeePaid = GasLimit × GasPrice is calculated from median gas values and stored in outbound params. This occurs in PayGasNativeAndUpdateCctx, PayGasInERC20AndUpdateCctx, or PayGasInZetaAndUpdateCctx depending on the coin type.

  Refund Calculation: On outbound completion, the keeper calculates:

  ActualFee = GasUsed × EffectiveGasPrice
  RemainingFees = UserGasFeePaid - ActualFee
  UsableRemaining = RemainingFees × 0.95
  StabilityPoolAmount = UsableRemaining × (StabilityPoolPercentage / 100)
  UserRefund = UsableRemaining - StabilityPoolAmount

  For zEVM-originated withdrawals with valid hex sender, StabilityPoolPercentage is read from chain params (default 100%). Non-withdrawal transactions send 100% to the stability pool. The function returns early without action if EffectiveGasPrice is nil, ActualFee ≥ UserGasFeePaid, or UsableRemaining rounds to zero.                                                                      

- Remove an observer and test TSS migration with a reduced observer set
The MsgRemoveObserver is used to remove an observer from the set . It removes the observer from the observer set and node account list
There are two tests in the makefile `start-tss-migration-add-observer`and `start-tss-migration-remove-observer` which test both the flows.
Another key thing to review is the TSS funds migration, this is. triggered between a tss generation and tss update to migrate funds . This should be done for all EVM and BITCOIN chains , however it has not been possible for us to test this against all evm chains on localnet , we just have ETH
Also added a new tool under zetatool , to fetch tss balances across all chains , this is specially important for BTC chains as we would neeed to use that to provide migration amount subtracting the expected fee

- V2 ZETA deposits and withdraws , (and add a flag to toggle it on and off)
This enables zeta deposits and withdraws using connector v2 , but is disabled by the flag .
The new connectors would not be enabled initially , so v37 should still fully support legacy flows
After the upgrade , we can deploy the v2 contracts but will not migrate fiunds from old connector to new , nor update the connector address on zetacore , the the EnableV2ZETAFlows will be false by default. 
This makes sure all zeta deposits and withdraw work as before using V1 flows until we decide to migrate to v2 and enable the V2ZETAFlows ` make start-connector-migration-test` tries to replicate this 
  | Aspect           | V1 (Legacy)                      | V2 (Gateway)                        |
  |------------------|----------------------------------|-------------------------------------|
  | Contract         | Connector (onReceive)            | GatewayZEVM                         |
  | Protocol Version | ProtocolContractVersion_V1       | ProtocolContractVersion_V2          |
  | Flag Required    | No                               | Yes (IsV2ZetaEnabled)               |
  | Message Encoding | Base64                           | Hex                                 |
  | Deposit Function | LegacyZETADepositAndCallContract | ProcessDeposit → processZetaDeposit |
  | Revert Function  | ZETARevertAndCallContract        | ProcessRevert → processZetaRevert   |
  | Abort Support    | Legacy refund only               | Full onAbort callback               |

### DIFF
--- a/x/crosschain/keeper/msg_server_vote_outbound.go
+++ b/x/crosschain/keeper/msg_server_vote_outbound.go
@@ -176,6 +176,11 @@ func (k Keeper) refundUnusedGas(
 	cctx *types.CrossChainTx,
 ) error {
 	outboundParams := cctx.GetCurrentOutboundParam()
+
+	if outboundParams.EffectiveGasPrice.IsNil() {
+		return nil
+	}
+
 	outboundTxFeePaid := math.NewUint(outboundParams.GasUsed).
 		Mul(math.NewUintFromBigInt(outboundParams.EffectiveGasPrice.BigInt()))
 	userGasFeePaid := outboundParams.UserGasFeePaid
@@ -244,6 +249,10 @@ func (k Keeper) FundGasStabilityPoolFromRemainingFees(
 	OutboundParams types.OutboundParams,
 	chainID int64,
 ) error {
+	if OutboundParams.EffectiveGasPrice.IsNil() {
+		return nil
+	}
+
 	gasUsed := OutboundParams.GasUsed
 	gasLimit := OutboundParams.EffectiveGasLimit
 	gasPrice := math.NewUintFromBigInt(OutboundParams.EffectiveGasPrice.BigInt())
--- a/x/crosschain/keeper/msg_server_vote_outbound_test.go
+++ b/x/crosschain/keeper/msg_server_vote_outbound_test.go
@@ -86,6 +86,7 @@ func TestKeeper_ManageUnusedGasFee(t *testing.T) {
 	tt := []struct {
 		name                               string
 		userGasFeePaidIsNil                bool
+		effectiveGasPriceIsNil             bool
 		gasUsed                            uint64
 		effectiveGasPrice                  math.Int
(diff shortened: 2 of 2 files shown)


kingpinXD:
> We do seem to be deleting an additional file: `zetaclient/chains/testutils/mocks/evm_client.go` when comparing `develop` branch to `release-reforge` branch [develop...release-reforge#diff-1e5fce4478dd1fad8c7b8c88099684ffad6288f0c58075d5035c78da6e95a486](https://github.com/zeta-chain/node/compare/develop...release-reforge#diff-1e5fce4478dd1fad8c7b8c88099684ffad6288f0c58075d5035c78da6e95a486) but if that's expected LGTM

That's fine , the testutils has been moved here 
https://github.com/zeta-chain/zeta-node/blob/e7691cf683701e84c21a0863c702bd8f650f2bfc/zetaclient/testutils/mocks/evm_client.go#L1-L537

the folder on develop is just the older code which was not removed

ws4charlie (approved):
the changes are unexpected. LGTM

PR #4596, opened 2026-05-13
fix(sui): process multiple gateway events per tx

### LINKED ISSUE #4584: Sui inbound observer rejects valid Gateway events with EventIndex != 0 and advances cursor past them
## Summary

The Sui inbound observer rejects any Gateway event with `EventIndex != 0` and the scanner cursor advances past the rejected event. On Sui, a single Programmable Transaction Block (PTB) can call the Gateway deposit / depositAndCall functions multiple times, generating Gateway events at `EventSeq = 0, 1, …`. Today only the first event is processed; later valid events in the same transaction are permanently skipped, leaving the user's funds in the Sui Gateway vault with no corresponding CCTX on ZetaChain.

## Code

`zetaclient/chains/sui/observer/inbound.go:172-176`

```go
case event.EventIndex != 0:
    // Is it possible to have multiple events per tx?
    // e.g. contract "A" calls Gateway multiple times in a single tx (deposit to multiple accounts)
    // most likely not, so let's explicitly fail to prevent undefined behavior.
    return errors.Errorf("unexpected event index %d for tx %s", event.EventIndex, event.TxHash)
```

`zetaclient/chains/sui/observer/inbound.go:73-102` — the cursor advancement loop only halts on `errTxNotFound` and `errVoteInbound`. Any other error (including the `EventIndex != 0` rejection above) falls through to `setCursor`, so the rejected event's slot is recorded and never revisited. Tracker recovery calls the same `processInboundEvent` path, so manual recovery hits the same gate.

`pkg/contracts/sui/gateway.go:318` — `EventIndex` is parsed from `event.Id.EventSeq`, so the parser already supports non-zero indices.

## Reproduction

A user submits a Sui PTB calling both `gateway::deposit<SUI>` and `gateway::deposit_and_call<SUI>` in a single transaction. The Sui transaction succeeds and emits two Gateway events at `eventSeq=0` and `eventSeq=1`. ZetaChain's observer processes the first event normally, rejects the second with "unexpected event index 1", and advances the cursor to `<tx>,1`. A later scanner pass does not replay the skipped event, and `MsgAddInboundTracker` cannot recover it because the tracker path also routes through `processInboundEvent`.

ZCNode-252 includes a working local Sui validator reproduction confirming the multi-event PTB shape.

## Fix direction

Either of the following resolves it; the first is preferred:

- Remove the `EventIndex != 0` rejection and process each Gateway event independently. The vote message already carries `EventIndex`, so each event maps to its own ballot.
- Alternatively, mirror the EVM observer's "first event per transaction" filter rather than a hard failure, and do not advance the cursor past valid events that were not voted on.

In either case, the cursor must not advance past a Gateway event that did not result in an inbound vote.

## HackenProof reports

- https://dashboard.hackenproof.com/manager/companies/zetachain/zetachain-blockchain-software/reports/ZCNode-243
- https://dashboard.hackenproof.com/manager/companies/zetachain/zetachain-blockchain-software/reports/ZCNode-252 (duplicate of ZCNode-243, adds local Sui validator PTB reproduction)

alan747271363-art:
## Summary
- allow Sui inbound observer to process Gateway events with non-zero event indexes
- keep each Gateway event mapped to its own inbound vote using the existing EventIndex field
- cover both block-scan and inbound-tracker recovery paths for multiple events in one Sui transaction

## Tests
- go test -tags pebbledb,ledger,test ./zetaclient/chains/sui/observer

Closes #4584

<h3>Greptile Summary</h3>

This PR fixes the Sui inbound observer to correctly handle transactions that emit more than one gateway event (e.g. a contract depositing to multiple accounts in a single PTB). Previously a hard guard rejected any event whose `EventIndex` was non-zero, which silently skipped all but the first event in such transactions.

- Removes the `event.EventIndex != 0` early-return in `processInboundEvent`, allowing every gateway event in a transaction to be parsed, validated, and voted on independently using its own `EventIndex`.
- Adds test coverage for the block-scan (`ObserveInbound`) and recovery (`ProcessInboundTrackers`) paths, verifying that two distinct inbound votes are produced with the correct receiver, amount, and event index for a two-event transaction.
- Introduces `SampleEventWithSeq` test helper to construct mock events with an arbitrary sequence number.

<h3>Confidence Score: 4/5</h3>

Safe to merge; the fix is focused, cursor advancement and vote idempotency are handled correctly, and both recovery paths are covered by tests.

The production change is a five-line deletion of a guard that blocked multi-event transactions. Cursor management, compliance skipping, and errTxNotFound/errVoteInbound retry semantics all continue to work correctly for the multi-event case. The one non-critical concern is that the block-scan path issues a redundant SuiGetTransactionBlock call per event for the same digest; the test even hard-codes two identical OnGetTx registrations for the same hash, confirming the behaviour is present but benign at current scale.

No files require special attention; inbound.go is the only production change and it is small and well-tested.

| Filename | Overview |
|----------|----------|
| zetaclient/chains/sui/observer/inbound.go | Removes the guard that rejected events with EventIndex != 0, enabling correct processing of multiple gateway events per transaction in both block-scan and tracker paths. |
| zetaclient/chains/sui/observer/observer_test.go | Adds two new integration-style test cases for multi-event transactions covering both ObserveInbound and ProcessInboundTrackers, and introduces SampleEventWithSeq helper to construct events with a configurable sequence number. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Observer
    participant SuiRPC
    participant ZetaCore

    Note over Observer: Block-scan path (ObserveInbound)
    Observer->>SuiRPC: QueryModuleEvents(cursor)
    SuiRPC-->>Observer: "[Event{TxA, seq=0}, Event{TxA, seq=1}, ...]"

    loop for each event
        Observer->>SuiRPC: SuiGetTransactionBlock(TxA)
        SuiRPC-->>Observer: tx (checkpoint, effects)
        Observer->>ZetaCore: "VoteInbound(TxA, EventIndex=N)"
        Observer->>Observer: setCursor(packageID, event.Id)
    end

    Note over Observer: Tracker path (ProcessInboundTrackers)
    Observer->>SuiRPC: "SuiGetTransactionBlock(TxA, ShowEvents=true)"
    SuiRPC-->>Observer: "tx with Events[seq=0, seq=1]"

    loop for each tx.Event
        Observer->>ZetaCore: "VoteInbound(TxA, EventIndex=N)"
    end
```

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `zetaclient/chains/sui/observer/inbound.go`, line 73-102 ([link](https://github.com/zeta-chain/node/blob/914c6f6c9e00503ea2a8c707d60a4d958a15c237/zetaclient/chains/sui/observer/inbound.go#L73-L102)) 

   <a href="#">P2</a> **Duplicate RPC call per event for same tx in block-scan path**

   When a single transaction emits N gateway events, `observeGatewayInbound` will call `SuiGetTransactionBlock` once per event (N times in total), even though all events share the same `TxDigest`. The test explicitly registers `OnGetTx` twice for the same hash (`ts.OnGetTx(txHash, "10000", true, false, nil)` × 2) confirming the current behavior.

   Under normal operation this is just wasteful, but it becomes a concern at scale when the same tx produces many events and the RPC node is rate-limited or has latency. Consider caching the fetched `*models.SuiTransactionBlockResponse` by `TxDigest` within the scan loop (a simple `map[string]*tx` is enough) and passing the cached value to `processInboundEvent` on subsequent events for the same digest.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: zetaclient/chains/sui/observer/inbound.go
   Line: 73-102

   Comment:
   **Duplicate RPC call per event for same tx in block-scan path**

   When a single transaction emits N gateway events, `observeGatewayInbound` will call `SuiGetTransactionBlock` once per event (N times in total), even though all events share the same `TxDigest`. The test explicitly registers `OnGetTx` twice for the same hash (`ts.OnGetTx(txHash, "10000", true, false, nil)` × 2) confirming the current behavior.

   Under normal operation this is just wasteful, but it becomes a concern at scale when the same tx produces many events and the RPC node is rate-limited or has latency. Consider caching the fetched `*models.SuiTransactionBlockResponse` by `TxDigest` within the scan loop (a simple `map[string]*tx` is enough) and passing the cached value to `processInboundEvent` on subsequent events for the same digest.

   How can I resolve this? If you propose a fix, please make it concise.
   `````
   </details>

</details>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
zetaclient/chains/sui/observer/inbound.go:73-102
**Duplicate RPC call per event for same tx in block-scan path**

When a single transaction emits N gateway events, `observeGatewayInbound` will call `SuiGetTransactionBlock` once per event (N times in total), even though all events share the same `TxDigest`. The test explicitly registers `OnGetTx` twice for the same hash (`ts.OnGetTx(txHash, "10000", true, false, nil)` × 2) confirming the current behavior.

Under normal operation this is just wasteful, but it becomes a concern at scale when the same tx produces many events and the RPC node is rate-limited or has latency. Consider caching the fetched `*models.SuiTransactionBlockResponse` by `TxDigest` within the scan loop (a simple `map[string]*tx` is enough) and passing the cached value to `processInboundEvent` on subsequent events for the same digest.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(sui): process multiple gateway event..."](https://github.com/zeta-chain/node/commit/914c6f6c9e00503ea2a8c707d60a4d958a15c237) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32000866)</sub>

## Summary by CodeRabbit

* **Performance**
  * Optimized inbound event processing for the Sui chain by implementing per-scan transaction caching, reducing redundant RPC calls when multiple events reference the same transaction.

* **Tests**
  * Expanded test coverage to verify proper handling of multiple gateway events within a single transaction on the Sui chain.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zeta-chain/node/pull/4596?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/zetaclient/chains/sui/observer/inbound.go
+++ b/zetaclient/chains/sui/observer/inbound.go
@@ -169,11 +169,6 @@ func (ob *Observer) processInboundEvent(
 		return nil
 	case err != nil:
 		return errors.Wrap(err, "unable to parse event")
-	case event.EventIndex != 0:
-		// Is it possible to have multiple events per tx?
-		// e.g. contract "A" calls Gateway multiple times in a single tx (deposit to multiple accounts)
-		// most likely not, so let's explicitly fail to prevent undefined behavior.
-		return errors.Errorf("unexpected event index %d for tx %s", event.EventIndex, event.TxHash)
 	}
 
 	if tx == nil {
--- a/zetaclient/chains/sui/observer/inbound.go
+++ b/zetaclient/chains/sui/observer/inbound.go
@@ -70,10 +70,12 @@ func (ob *Observer) observeGatewayInbound(ctx context.Context, packageID string)
 		Int("events", len(events)).
 		Msg("processing inbound events")
 
+	txCache := make(map[string]*models.SuiTransactionBlockResponse)
+
 	for _, event := range events {
 		// Note: we can make this concurrent if needed.
 		// Let's revisit later
-		err := ob.processInboundEvent(ctx, event, nil, false, false)
+		err := ob.processInboundEvent(ctx, event, nil, txCache, false, false)
 
 		switch {
 		case errors.Is(err, errTxNotFound),
@@ -159,6 +161,7 @@ func (ob *Observer) processInboundEvent(
 	ctx context.Context,
 	raw models.SuiEventResponse,
 	tx *models.SuiTransactionBlockResponse,
+	txCache map[string]*models.SuiTransactionBlockResponse,
 	fromTracker bool,
 	isInternalTracker bool,
 ) error {
@@ -169,6 +172,12 @@ func (ob *Observer) processInboundEvent(
 		return nil
 	case err != nil:
 		return errors.Wrap(err, "unable to parse event")
+	}
+
+	if tx == nil && txCache != nil {
+		if cached, ok := txCache[event.TxHash]; ok {
+			tx = cached
+		}
 	}
 
 	if tx == nil {
@@ -182,6 +191,9 @@ func (ob *Observer) processInboundEvent(
 		}
 
 		tx = &txFresh
+		if txCache != nil {
+			txCache[event.TxHash] = tx
+		}
 	}
 
 	msg, err := ob.constructInboundVote(event, *tx)
@@ -221,7 +233,7 @@ func (ob *Observer) processInboundTracker(ctx context.Context, tracker cctypes.I
 	}
 
 	for _, event := range tx.Events {
-		if err := ob.processInboundEvent(ctx, event, &tx, true, isInternal); err != nil {
+		if err := ob.processInboundEvent(ctx, event, &tx, nil, true, isInternal); err != nil {
 			return errors.Wrapf(err, "unable to process inbound event %s", event.Id.EventSeq)
 		}
 	}
--- a/zetaclient/chains/sui/observer/observer_test.go
+++ b/zetaclient/chains/sui/observer/observer_test.go
@@ -276,6 +276,64 @@ func TestObserver(t *testing.T) {
 		require.Empty(t, ts.inboundVotesBag)
 	})
 
+	t.Run("ObserveInbound handles multiple gateway events in the same tx", func(t *testing.T) {
+		// ARRANGE
+		ts := newTestSuite(t)
+		packageID := ts.gateway.PackageID()
+		txHash := "TX_MULTI_EVENT"
+
+		evmBob := sample.EthAddress()
+		evmAlice := sample.EthAddress()
+
+		expectedQuery := client.EventQuery{
+			PackageID: packageID,
+			Module:    sui.GatewayModule,
+			Cursor:    "",
+			Limit:     client.DefaultEventsLimit,
+		}
+
+		events := []models.SuiEventResponse{
+			ts.SampleEventWithSeq(packageID, txHash, "0", string(sui.DepositEvent), map[string]any{
+				"coin_type": string(sui.SUI),
+				"amount":    "200",
+				"sender":    "SUI_BOB",
+				"receiver":  evmBob.String(),
+			}),
+			ts.SampleEventWithSeq(packageID, txHash, "1", string(sui.DepositAndCallEvent), map[string]any{
+				"coin_type": string(sui.SUI),
+				"amount":    "300",
+				"sender":    "SUI_ALICE",
+				"receiver":  evmAlice.String(),
+				"payload":   preparePayload([]byte{1, 2, 3}),
+			}),
+		}
+
+		ts.suiMock.On("QueryModuleEvents", mock.Anything, expectedQuery).Return(events, "", nil)
+		ts.OnGetTx(txHash, "10000", true, false, nil)
+		ts.OnGetTx(txHash, "10000", true, false, nil)
+
+		getCctxByHashErr := grpcstatus.Error(grpccodes.InvalidArgument, "anything")
+		ts.zetaMock.MockGetCctxByHash("", getCctxByHashErr)
+		ts.CatchInboundVotes()
+
+		// ACT
+		err := ts.ObserveInbound(ts.ctx)
+
+		// ASSERT
+		require.NoError(t, err)
+		require.Equal(t, "TX_MULTI_EVENT,1", ts.GetAuxString(packageID))
+		require.Len(t, ts.inboundVotesBag, 2)
+
+		require.Equal(t, uint64(0), ts.inboundVotesBag[0].EventIndex)
+		require.Equal(t, evmBob.String(), ts.inboundVotesBag[0].Receiver)
+		require.Equal(t, math.NewUint(200), ts.inboundVotesBag[0].Amount)
+
+		require.Equal(t, uint64(1), ts.inboundVotesBag[1].EventIndex)
+		require.Equal(t, evmAlice.String(), ts.inboundVotesBag[1].Receiver)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by alan747271363-art** [comment]
Follow-up pushed in `be326be2` to address the Greptile note about duplicate `SuiGetTransactionBlock` calls in the block-scan path.

What changed:
- Added a per-scan `TxDigest` cache in `observeGatewayInbound`.
- Reused the cached transaction block when multiple gateway events come from the same Sui transaction.
- Updated the multi-event regression so the shared transaction block is mocked once while still asserting two inbound votes with event indexes 0 and 1.

Validation:
- `git diff --check HEAD~1..HEAD` passes.
- I attempted the focused Go test, but this local Windows environment does not have Go installed, and Docker Desktop is not running, so I could not execute `go test` locally.

**Comment by roopd3v** [comment]
I am the researcher who found and reported this bug.

And after ZetaChain used my report to fix the vulnerability, the program manager downgraded it to Low ( = 0$ ) with this ridiculous explanation:
```
@Tanmay12312
manager
May 26, 2026 05:19am
Thanks for bringing this up
The issue has been fixed, and the fix was applied after your investigation; therefore, this report has proved useful.

However that attack itself has no benift for the attacker , I would classify this as low
```

So let me make this very clear for every researcher reading this.

ZetaChain accepted that the issue was real.
ZetaChain opened a public GitHub security issue for it.
ZetaChain referenced the HackenProof reports, including ZCNode-243.
ZetaChain fixed the exact bug.
ZetaChain added tests for the exact scenario.
And only when it came time to pay, the impact suddenly became “Low” just because low severity means 0$ payment.

That is shameful.

The public GitHub issue says the Sui inbound observer rejected valid Gateway events with EventIndex != 0, advanced the scanner cursor past the rejected event and could leave user funds in the Sui Gateway vault with no corresponding CCTX on ZetaChain.

The public fix PR is literally titled:

“fix(sui): process multiple gateway events per tx”

The PR says it allows the Sui inbound observer to process Gateway events with non-zero event indexes and keeps each Gateway event mapped to its own inbound vote.

So do not insult researchers by pretending this was some worthless Low issue.

If the bug was useful enough to create a public security issue, useful enough to reference my report, useful enough to fix in the core node and useful enough to add regression tests for, then it was useful enough to pay for!

“The attacker has no benefit” is a shameful excuse.

Security impact is not only attacker profit. A bug can harm users, break protocol execution, cause stuck funds, skip valid deposits, break cross-chain flows and still be a valid bounty issue.

This bug caused exactly that:

- a valid Sui Gateway event could be skipped;
- the observer could reject the later event;
- the cursor could advance past it;
- no inbound vote would be created;
- no CCTX would be created;
- the user’s cross-chain action could remain stuck.

Calling that Low after using the report and applying the fix is not honest triage. It is free security labor exploitation.

This is the kind of behavior that makes researchers stop reporting bugs.

If ZetaChain wants to run a bug bounty program, then act like it. Do not use researchers’ work to fix your software and then hide behind a nonsense downgrade to avoid paying.

Have some shame!!!

Other researchers should see this before spending time helping this program. If a valid report can be accepted, used, publicly fixed, referenced in GitHub and then downgraded below the payment threshold with this kind of excuse, then people deserve to know exactly how this was handled.

**White hats/Researchers: don't waste your time on this project!**

PR #4612, opened 2026-07-21
feat(zetaclient): emergency TSS native-fund drain (EVM + BTC)

kingpinXD:
## Summary
- Adds a **zetaclient-only** way to drain all native TSS funds — every EVM chain's gas token plus BTC — to a safe wallet for a crosschain shutdown. No zetacore changes.
- An off-chain, signed JSON payload (served by `zetatool drain-payload --serve`, draft → freeze → final) pins the fully-resolved transaction for each chain. Every zetaclient verifies it against a baked-in key and, at the trigger height, TSS-signs and broadcasts. The receiver is **hardcoded per network** as an anti-redirect anchor.
- BTC sweeps spend only the payload-pinned UTXOs (≤20 inputs each), signed through the existing `SignTx` path.
- Gated behind a `drain` build tag, off by default; **fails closed** if the pubkey or receiver isn't configured.
- Design, coordination model, and security notes: `docs/rfc/002_tss_emergency_drain.md`.
- Verified end to end: a real 2-node localnet drains both EVM and BTC (`drain_tss`).

**Before merge / any real drain (operator):** fill `OperatorPubKeyHex` and the testnet/mainnet receiver addresses — both stay fail-closed sentinels until set.

---

> [!NOTE]
> **High Risk**
> Moves all native TSS funds via TSS signing with security relying on compile-time receivers, operator signatures, and optional drain builds—misconfiguration or a bad release could redirect or prematurely sweep funds.
> 
> **Overview**
> Adds an **emergency shutdown path** to sweep native TSS balances on external EVM chains and Bitcoin to **network-hardcoded safe wallets**, without zetacore changes.
> 
> Operators use **`zetatool drain-payload`** to build operator-signed JSON from live balances, gas, nonces, and pinned UTXOs, optionally served via a **draft → freeze → final** HTTP cron. **`zetaclientd`** with the `drain` build tag arms only when `ZETACLIENT_DRAIN_URL` is set: it polls, verifies the baked-in operator key, enforces receiver anchors, and at the trigger Zeta height TSS-signs and broadcasts pinned EVM transfers and BTC sweeps (≤20 inputs per tx).
> 
> Shared fee math lives in **`pkg/migration`** (EVM migration CCTX now calls `ComputeEVMMigration`). Localnet CI adds **`start-drain-test`**, `drain`/`drain_localnet` image tags, and **`drain_tss`** e2e. Testnet/mainnet pubkey and receivers remain **UNSET** until configured before any real drain.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49e2cd2619109164b84f3c53a63c416a70aa87e8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

Adds a build-tag-gated emergency mechanism for moving native TSS funds to network-specific receiver anchors.
- Introduces signed, canonical drain payloads and a draft-to-final payload server.
- Generates pinned EVM transfers and batched Bitcoin sweeps from live chain state.
- Adds a zetaclient poller that verifies payload signatures, trigger windows, receivers, fees, and signer availability before TSS signing and broadcasting.
- Adds localnet orchestration, end-to-end coverage, build configuration, and operational documentation.

The PR appears safe to merge based on the reviewed drain generation, verification, signing, and broadcast paths.

The drain remains disabled outside tagged and explicitly armed builds, validates signed payloads against fixed receiver anchors, and constrains transaction construction and execution without an identified actionable failure.

| Filename | Overview |
|----------|----------|
| zetaclient/drain/poller.go | Implements signed-payload polling, trigger-window coordination, receiver and fee validation, retry tracking, and concurrent EVM/BTC execution. |
| cmd/zetaclientd/drain_on.go | Arms the drain only for tagged builds with validated network anchors, operator key, Bitcoin address, and live signer resolvers. |
| cmd/zetatool/cli/drain.go | Adds payload generation and serving from live TSS addresses, EVM state, Bitcoin UTXOs, filters, and operator-supplied timing. |
| pkg/drain/generate.go | Deterministically computes EVM drain transactions and partitions economical Bitcoin sweeps into bounded input groups. |
| pkg/draintx/payload.go | Defines the drain transaction schema and deterministic signed encoding covering all execution-relevant fields. |
| zetaclient/chains/bitcoin/signer/sign.go | Exposes the existing TSS Bitcoin transaction-signing path for the newly constructed pinned sweeps. |
| zetaclient/chains/evm/signer/sign_drain.go | Adds deterministic signing and broadcasting for payload-pinned native EVM transfers. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant O as Operator / zetatool
    participant S as Payload server
    participant Z as Zetaclient poller
    participant C as Zetacore height provider
    participant T as TSS signers
    participant E as External chains
    O->>O: Resolve balances, nonces, gas, and UTXOs
    O->>O: Build and sign final payload
    O->>S: Publish final payload
    loop Poll interval
        Z->>S: Fetch payload
        Z->>Z: Verify signature and receiver anchors
        Z->>C: Read current Zeta height
    end
    alt Inside trigger window and signers ready
        Z->>T: Sign pinned EVM and BTC transactions
        T-->>Z: Signed transactions
        Z->>E: Broadcast transactions
    else Not ready or outside window
        Z->>Z: Wait, retry, or reject payload
    end
```

<sub>Reviews (1): Last reviewed commit: ["chore: fix changelog, formatting, and li..."](https://github.com/zeta-chain/node/commit/49e2cd2619109164b84f3c53a63c416a70aa87e8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46758287)</sub>


cmd/zetaclientd/drain_on.go
```diff
@@ -0,0 +1,226 @@
...
+			var evmChains, btcChains []int64
+			for _, cs := range orch.ObserverSigners() {
+				switch c := cs.(type) {
+				case *evm.EVM:
+					evmChains = append(evmChains, c.Chain().ChainId)
+				case *bitcoin.Bitcoin:
+					btcChains = append(btcChains, c.Chain().ChainId)
+				}
+			}
+			if len(evmChains) > 0 && len(btcChains) > 0 {
```
ws4charlie: waitForSigners bails the whole drain if *either* the EVM or BTC signer family isn't up within 2min (this line needs both), so if BTC signers are down or late in an emergency the healthy EVM drain never even starts. Since the poller already resolves signers live at fire time (Config.ResolveEVMSigner/ResolveBTCSigner, used in attemptPending), could we start once at least one family is ready and let per-chain resolution handle the rest?

kingpinXD: Good catch. I dropped the arm-time both-families gate. It was too blunt, blocking a healthy EVM drain just because BTC signers were slow to bootstrap.

I went a little differently than "start once one family is ready". The payload is now the source of truth for what drains, and the poller gates per-payload at fire time, fail-closed: it fires only if it can resolve a signer for every chain in that payload, otherwise it skips the block and waits for the payload to be reset. So it stays strict (never half-fires a payload or signs for a chain whose signer is down) without blocking on a family I'm not draining.

That also gives the offline-family and retry behaviour straight from the payload:
- BTC signers down but EVM needs to move? Publish an EVM-only payload, it fires and BTC just isn't in it.
- Some chains fail to broadcast? Republish at a higher trigger height with just the remaining chains. The poller only fires a payload whose trigger height is newer than the last one it fired, so nothing old re-triggers in between.

Per-chain live resolution (`ResolveEVMSigner`/`ResolveBTCSigner`) at fire time stays as you pointed out.

pkg/drain/receivers.go
```diff
@@ -0,0 +1,123 @@
...
+// test can inject its own material; testnet/mainnet always use the compiled anchors.
+func ResolveAnchors(network string) (pubKey []byte, receivers Receivers, err error) {
+	receivers, err = ReceiverForNetwork(network)
+	if err != nil {
+		return nil, Receivers{}, err
+	}
+
+	pubKeyHex := OperatorPubKeyHex
+	if network == NetworkLocalnet {
+		if v := os.Getenv(EnvLocalnetPubKey); v != "" {
```
hernan-clich: The localnet env overrides defeat the hardcoded-anchor security model on a production drain build. `ResolveAnchors` honors `ZETACLIENT_DRAIN_PUBKEY` / `_EVM_RECEIVER` / `_BTC_RECEIVER` whenever `network == "localnet"`, and `network` itself comes from an env var (`ZETACLIENT_DRAIN_NETWORK`). The drain binary operators run in the window is the same build that contains this path, so anyone with env access to the zetaclient process can set `ZETACLIENT_DRAIN_NETWORK=localnet` plus an attacker pubkey and receivers, and redirect the entire drain. That's exactly the "compromised host" the RFC says can change *when* but never *where* — env access is a lower bar than swapping a compiled binary, and this reopens the destination. Suggest gating the localnet path behind its own build tag so a production drain build physically can't honor those envs.

pkg/migration/migration.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// ComputeBTCSweep computes the single-output amount and fee (in satoshis) for one transaction
+// of the emergency drain's multi-tx BTC sweep.
+//
+// Unlike ComputeBTCMigration it charges the miner fee only (no RBF or nonce-mark reserve): the
+// drain partitions a large UTXO set into many sweep txs, and a per-tx reserve would both
+// multiply across N txs and, in a single-output sweep, be paid to miners rather than preserved.
+func ComputeBTCSweep(totalInputSats, feeRate int64) (outputSats, feeSats int64, err error) {
+	feeSats = btccommon.OutboundBytesMax * feeRate
```
hernan-clich: `ComputeBTCSweep` charges `OutboundBytesMax` (1543 vB, a 21-input/2-output tx) per sweep regardless of the actual input count. For 500 UTXOs in even groups of 20 that's a small per-tx overcharge, but a tail partition with a few inputs pays the full 21-input fee, and in a single-output sweep the excess is burned to miners (no change output). This is the residual of the ~0.25 BTC issue from the thread — dropping the RBF/nonce-mark reserve fixed the big one, this is what's left. Suggest sizing the fee to the real input count:
```go
feeSats = common.EstimateOutboundSize(int64(numInputs), 1, toAddrType) * feeRate
```

zetaclient/drain/poller.go
```diff
@@ -0,0 +1,496 @@
...
+		return errors.Errorf(
+			"btc amounts inconsistent: inputs %d != output %d + fee %d",
+			totalIn,
+			tx.OutputSats,
+			tx.FeeSats,
+		)
+	case tx.FeeSats <= 0:
+		return errors.Errorf("btc fee is non-positive: %d", tx.FeeSats)
+	case tx.FeeSats > totalIn/maxBTCFeeFraction:
+		return errors.Errorf("btc fee %d exceeds 1/%d of inputs %d", tx.FeeSats, maxBTCFeeFraction, totalIn)
```
hernan-clich: `validateBTCFee`'s 25%-of-inputs cap can silently strand a small tail partition. If a group's UTXOs are small enough that the (oversized, fixed) fee exceeds `totalIn/4`, the tx is rejected, that sweep never completes, the payload never reaches `allDone`, the window elapses, and the operator has to republish. Combined with the fixed-size fee above, a tail group of a few small UTXOs is the trigger. Right-sizing the fee (other comment) plus having the generator avoid tiny tail groups (e.g. rebalance the last two partitions) closes it.

cmd/zetaclientd/drain_on.go
```diff
@@ -0,0 +1,184 @@
...
+func drainWindowFromEnv() int64 {
+	if v := os.Getenv(envDrainWindow); v != "" {
+		if n, err := strconv.ParseInt(v, 10, 64); err == nil && n > 0 {
+			return n
+		}
+	}
+	return drainWindow
+}
+
+func btcNetParams(network string) (*chaincfg.Params, error) {
```
hernan-clich, line 189: `btcNetParams` defaults unknown networks to regtest params. Currently unreachable (a bad network fails earlier in `ReceiverForNetwork`), but for a money path a `default: return error` is more fail-closed than silently picking regtest.

hernan-clich (commented):
Strong PR, design is well-reasoned. One high priority finding (localnet env escape defeats the anchor model on a prod build) worth fixing before any real drain, plus BTC fee-sizing tightening.

pkg/drain/receivers.go
kingpinXD: Agreed, env access shouldn't be able to reopen the destination. I moved the whole localnet path (its baked-in receivers and the `ZETACLIENT_DRAIN_*` overrides) behind a `drain_localnet` build tag. A production drain build (`-tags drain`) no longer contains it at all, so `ZETACLIENT_DRAIN_NETWORK=localnet` there resolves to an unknown network and fails closed. Only the e2e image, built with the extra tag, honors those envs.

pkg/migration/migration.go
kingpinXD: Good point. I switched the sweep to size the fee by the real input count, `EstimateOutboundSize(numInputs, [payee]) * feeRate`, so a small tail group no longer pays the full 21-input fee. numInputs is bounded by the 20-input partition cap, so the size stays within `OutboundBytesMax`. The tss-balances display keeps the max-size estimate, since it's a conservative pre-selection number.

zetaclient/drain/poller.go
kingpinXD: Right-sizing the fee (other thread) makes this far less likely, and I also had the generator skip any group whose fee would exceed the same 1/4 cap the poller enforces, so it never emits a sweep the poller would reject.

On the UTXO limit: the normal path caps inputs at `MaxNoOfInputsPerTx` (20) inside `SelectUTXOs`. The drain pins its inputs instead, so it enforces the same cap itself by partitioning the UTXO set into groups of at most 20 (`common.MaxNoOfInputsPerTx`, now shared). `SignTx` has no cap of its own, so each group signs and broadcasts as one outbound within `OutboundBytesMax`, and a large set sweeps as several txs. Any partition left behind (dust or over-cap) is picked up by republishing the remaining chains at a higher trigger height.

cmd/zetaclientd/drain_on.go
kingpinXD: Done. `btcNetParams` now returns an error for an unknown network instead of defaulting to regtest.

s2imonovic:
The concept looks good overall to me.

ws4charlie:
bugbot run

ws4charlie:
Re-ran bugbot on the new head — heads up @kingpinXD, the 3 older flags (wrong BTC chain / empty payload / cron first-publish) are stale re-posts of already-fixed findings; they still cite the old line numbers and look resolved to me on `49e2cd2`. The 2 new ones look real though: `already-known broadcast never completes` (a node whose Broadcast hits already-known / nonce-too-low never marks the tx done and keeps re-signing until the window elapses — worth treating those as success) and `serve exits before clients can fire` (RunCron closes the server right after publishing the final at H-K, so over `--serve` clients cant fetch it at H — probably fine if prod distributes the final via an object store, but worth confirming). Could you take a look at those two?

julianrubino (commented):
# Review — emergency TSS drain 

Me read all code. Big design is **good**. Idea is smart: each node just sign one ready-made transaction, node decide nothing. Safe wallet address is baked inside the code, so a bad server can change **WHEN** money move, but never **WHERE**. This part is strong. Good work.

**Not merge yet is correct.** Almost all problems are the same type: **drain does not happen, but the log says "done".** For an emergency money-rescue tool, this is the worst kind of bug. No bug can send money to the wrong place (anchor is safe). But many bugs can leave money **stuck** or **not moved**, while the operator thinks everything is fine.

Also important: today the tool **cannot run a real drain**. `OperatorPubKeyHex` is all zeros (placeholder). Mainnet/testnet safe address is `UNSET`. The guard rejects both. So **no real money at risk today** — this is a skeleton waiting for the real addresses.

---

## 🔴 Must fix before any real drain

**1. Gas limit is fixed to 21000. This assumes receiver is a normal wallet. If safe wallet is a contract, EVERY EVM drain fails.**
`pkg/migration/migration.go:50`
21000 gas = only enough to send to a normal address (EOA). A "safe wallet" is often a Gnosis Safe / multisig = a **contract**. Sending to a contract runs its code, which needs more than 21000 gas → transaction reverts (out of gas) on every EVM chain → money does not move.
**Fix:** check the receiver is a normal wallet (EOA) before cutting the build. If it must be a contract, raise the gas limit and fix the fee.

**2. The payload server dies if one RPC fails → no final payload → drain never fires.**
`pkg/drain/server.go:107`, `cmd/zetatool/cli/drain.go:394`
The server builds the payload for all chains. If ONE chain's RPC fails (balance/gas/nonce), the whole server stops. Then no `final` payload is published. Nodes only have a draft, and they refuse to sign a draft. So at the trigger time → nothing happens. This tool must run a long time over many chains, so one flaky RPC is very likely.
**Fix:** the server should log the error and keep going (retry next tick), not stop. One bad chain = skip that chain, not kill everything.

**3. Node says "done" when the transaction is only in the mempool, not confirmed. Fee is fixed, cannot bump. A stuck transaction is a fake "done".**
`zetaclient/drain/poller.go:303`
The node marks "done" the moment the transaction is accepted to the mempool — not when it is mined. Gas price and BTC fee are fixed at build time (some blocks before the trigger). If the network is busy (very likely during a shutdown), the fee is too low → the transaction is stuck forever. Retry re-sends the **same** transaction (no higher fee). The BTC "RBF" only turns on a flag, it never re-signs higher. The poller prints "drain complete" but the money is still at the TSS.
**Fix:** "done" must mean **confirmed**, or the log must say "sent (not confirmed)". Write in the doc: to fix a stuck tx, the operator must republish with a new, higher trigger height and fresh fee.

---

## 🟠 Should fix

**4. Nonce is fixed in the payload. If a normal transaction is still pending, the nonce clashes → the drain is stuck on that chain forever.**
`cmd/zetaclientd/start.go`, `zetaclient/drain/poller.go`
The drain runs at the same time as the normal signer. The code does NOT stop inbound or drain the pending nonces (the RFC says to do it, but only as a manual step, not in code). If a normal outbound waits at the same nonce → one wins, the drain loses → the drain tx is rejected forever (it only ever uses the fixed nonce, never nonce+1).
**Fix:** at fire time, check that the **live nonce == payload nonce**; if not equal → stop. Better: only allow arming when inbound is already disabled.

**5. If some chains are skipped, the payload still signs and reads as "complete" for the missing chains.**
`cmd/zetatool/cli/drain.go:389`, `zetaclient/drain/poller.go:134`
Good news: a **fully-empty** payload IS blocked (`drain.go:304` refuses an empty final — so the Cursor "empty final" finding is mostly a false positive). Bad news: if only SOME chains are skipped (RPC missing, or balance shows 0 for a moment), the other chains still make a valid payload, and the poller says "complete" while the skipped chains never drain.
**Fix:** compare expected chains vs included chains. If a chain with money is missing → ask the operator to confirm, or fail. Also add a "not empty / count matches" check in the poller (extra safety).

**6. No production-only build target. The localnet escape has no guard.**
`Makefile:294`
The localnet redirect env-override is behind the `drain_localnet` build tag. On a correct production build it does NOT exist — so Hernan's finding is downgraded (not "defeats the anchor", just a hardening gap). BUT the only drain build example in the repo (`LOCALNET_BUILD_TAGS`) includes `drain_localnet`, and there is no production `-tags drain` example. An operator copying the only example → ships the escape. Then an attacker with env access sets `ZETACLIENT_DRAIN_NETWORK=localnet` + their own receiver → redirects all EVM money.
**Fix:** add a clear production build target (`-tags drain` only) + a runtime guard: if built with `drain_localnet`, refuse mainnet/testnet and print a big "NON-PRODUCTION BUILD" banner.

**7. BTC default fee 50 sat/vB is too low for a busy network. No auto bump.**
`pkg/migration/migration.go:28`
In 2023–2024 the fee spiked to 300–500 sat/vB. The emergency drain happens exactly when the network may be busy.
**Fix:** the operator must set `--fee-rate` from the live mempool at drain time. Document a floor / a live estimate source.

**8. The signed payload has no network name inside. Cross-network replay is possible if the same key + same address are reused.**
`pkg/draintx/payload.go:64`
The only thing that stops a testnet payload from working on mainnet = different receiver + different key. If ops reuse one safe address + one key for both → a testnet `final` is valid bytes on mainnet.
**Fix:** add a fixed tag (e.g. `"ZETADRAIN"`) + the zeta network / chain-id inside the signed bytes.

---

## 🔵 Small / nits
- `--fee-rate 0` (BTC) and `gasPrice 0` (EVM) create a payload that the poller/chain silently drops. Validate `> 0` at build time (fail fast).
- EVM has no fee cap; BTC caps the fee at 25% of inputs. If the operator key leaks, a huge `gasPrice` can burn EVM money. Add a cap for EVM. Also tighten the BTC 25% → ~10%.
- `lastFiredHeight` is only in memory. A crash + restart re-broadcasts (safe — no double spend, nonce/UTXO pinned) but re-reports "unfinished" → maybe a useless republish. Persist it, or treat "nonce too low" / "already known" as success.
- The "unfinished" summary shows only counts, not WHICH chains failed. Print the chain names so the operator knows what to republish. `poller.go:459`.

---

## ✅ Cleared / safe (please do not chase these)
- **Anti-redirect: strong.** Three anchors, AND the signer uses the **compiled** receiver, not the payload `to`. So even if the `to` check is bypassed, money still goes to the baked address. Cannot redirect. Fail-closed all verified (placeholder pubkey rejected in 2 layers, UNSET rejected, wrong-network BTC address aborts arming).
- **@Tanmay — the "0.25 BTC waste" is a false alarm, about 150x too big.** The fee is per-vByte to the real input count. 500 UTXOs → 25 sweeps → total ~**0.018 BTC** @50 sat/vB. The 20-input cap is a hard Bitcoin rule, not a knob. A single 500-input tx (not even allowed by standardness) would save only ~0.0016 BTC. **No BTC change needed here** — you can save your effort.
- **Cursor "wrong BTC chain" = false positive** (old revision). The code uses the hardcoded `GetBTCChainID` with an explicit "never scan SupportedChains" guard.
- EIP-155 replay = safe (chain-id bound live per chain). Underflow = guarded. BTC signing/partition = correct (all UTXOs covered, per-input sighash correct). All 3 `#nosec G115` = justified, hide nothing.

---

## Notes for the drain runbook (not code)
1. The real pubkey + the real safe addresses need their OWN review + secure setup before the build. This is as important as the code review.
2. Confirm the EVM safe wallet is an EOA (see #1).
3. Enforce quiescence (disable inbound + drain the pending nonces) and check the nonce at fire time (see #4).
4. Set live fee / gas at drain time (#3, #7).
5. After the drain: upgrade nodes off the drain build + take the endpoint down (the RFC says this).

---

Overall: good and careful work — the design is the right shape. The main thing is the **"log says done but money did not move"** family (#1–#5). Fix those and it is solid. I do not approve (you said do not merge, still testing) — just leaving thorough comments.

cmd/zetatool/cli/drain.go
```diff
@@ -0,0 +1,553 @@
...
+
+	isFinalTime := func(ctx context.Context) (bool, error) {
+		height, err := zetaHeight(ctx, gen.zetacore)
+		if err != nil {
+			return false, err
+		}
+		return height >= opts.triggerHigh-opts.freezeK, nil
+	}
+
+	return drain.RunCron(ctx, opts.interval, gen.generate, server.Publish, isFinalTime)
```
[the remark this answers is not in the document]
kingpinXD: Fixed in c22d003df. serveDrain no longer closes the server when the final is published at H-K; it keeps serving the final until the operator interrupts (SIGINT/SIGTERM), so clients can fetch it throughout the [H, H+window) firing window.

zetaclient/drain/poller.go
```diff
@@ -0,0 +1,494 @@
...
+	// #nosec G115 height is a positive zeta block height
+	if err := signer.SignTx(ctx, msgTx, inputAmounts, uint64(height), btcKeysignNonce); err != nil {
+		return errors.Wrap(err, "sign")
+	}
+	if err := signer.Broadcast(ctx, msgTx); err != nil {
+		return errors.Wrap(err, "broadcast")
+	}
+
+	p.Logger.Warn().Int64("chain", tx.ChainID).Stringer("tx", msgTx.TxHash()).Msg("broadcast btc drain sweep")
+	return nil
```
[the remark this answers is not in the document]
kingpinXD: Fixed in c22d003df. executeEVM/executeBTC now treat already-known / in-mempool / nonce-too-low broadcast errors as success. Drain txs are byte-identical across nodes, so once one node broadcasts the item is marked done instead of re-signed until the window elapses.

ws4charlie (approved):
LGTM — verified the two latest fixes on c22d003df (server now stays up through the [H, H+window) firing window, and already-known / nonce-too-low / in-mempool broadcasts are treated as done), and all earlier findings (localnet-env anchor bypass, BTC fee sizing, empty-final, wrong-BTC-chain, waitForSigners gate) are resolved. Please make sure CI is green before merging.

kingpinXD:
Addressed the review in `8c02edd05`:

**Must-fix**
- **#1 gas/multisig** — pin 100k gas for the drain EVM tx (on-chain migration stays at 21k). We drain to a known safe wallet, so a fixed higher limit that covers a Gnosis Safe `receive()` is simpler than probing `getCode`, and unused gas is refunded.
- **#2 one bad RPC** — the payload builder now skips a chain whose RPC fails and keeps going, so a `final` still publishes for the healthy chains (the fully-empty final is still refused).
- **#3 done ≠ confirmed** — the poller now logs `broadcast (not confirmed on-chain)`; the RFC documents that a stuck under-fee tx is fixed by republishing at a new higher trigger height with fresh fee/gas.

**Should-fix**
- **#4 nonce clash** — check live nonce == pinned nonce before signing; stop that chain otherwise.
- **#5 partial completeness** — log the chain IDs of unfinished and skipped chains.
- **#6 prod build** — added an `install-zetaclient-drain` target (`-tags drain`, no `drain_localnet`); a `drain_localnet` build refuses a non-localnet network and prints a NON-PRODUCTION banner.
- **#7 fees** — tightened the BTC fee cap to 1/10, capped EVM gas price at 10k gwei, and warn on a low `--fee-rate`; live-fee guidance is in the runbook.
- **#8 replay** — bound a `ZETADRAIN` domain tag + the network into the signed bytes; the poller rejects a payload built for another network.

**Nits** — fail fast on non-positive fee-rate / gas price / gas limit.

On the BTC "0.25 BTC waste": agreed, that was already resolved earlier (fee is per-input, not per-group). The real operator pubkey + safe addresses stay UNSET placeholders and will get their own review before any build.

skosito (approved):

s2imonovic (approved):

julianrubino (commented):
Re-reviewed the current head against the earlier findings + verified the anti-redirect anchor and both backports. Overall this is in strong shape — the design is right and it fails closed. Summary + a few non-blocking items to consider before the real safe addresses are compiled in.

## ✅ Verified good
- **Anti-redirect anchor intact.** Both signer paths use the *compiled* receiver, never `payload.To`: EVM `poller.go:438` → `SignDrainTx(ctx, p.EVMReceiver, …)`; BTC `poller.go:473` → `buildSweep(p.BTCReceiver, …)`. `payload.To` is only an equality pre-check. A tampered payload can change *when* funds move, never *where*.
- **Prod fails closed.** Compiled anchors are still `UNSET`/zero-pubkey placeholders (`pkg/drain/receivers.go`) → arm-time guard rejects, so no real drain is possible until the real addresses land. No live funds at risk today.
- **Prior findings resolved:** gas limit → `DrainEVMGasLimit=100_000` (fixed); fixed-nonce collision → live-nonce guard `poller.go:423-430` (fixed); replay → `Network` folded into the signed digest + `domainTag "ZETADRAIN"` + poller check `poller.go:286` (fixed); prod build target `install-zetaclient-drain` with no `drain_localnet` + runtime guard against a localnet-tagged prod build (fixed); done is now honestly labeled "broadcast, not confirmed" + duplicate-broadcast races handled.

## 🟠 Non-blocking findings (worth closing before real arming)
**1. [Medium] Cron aborts on any transient zetacore error → the final is never published.**
`pkg/drain/server.go:119` / `:131`: `if final, err := publishOnce(); err != nil || final { return err }`. A one-off failure of `isFinalTime`/`gen` (LastZetaHeight, currentTSS, GetTssAddressByFinalizedHeight, SupportedChains) on *any* draft tick returns from `RunCron` and tears the server down before the single final publishes → drain silently never fires. For an emergency tool this is the worst failure mode. Suggest: log-and-continue on draft-tick errors; only treat inability to publish the final *through the freeze window* as fatal.

**2. [Low/Med] EVM nonce guard is blind to an unmined pending tx.**
`PendingNonce` resolves to `NonceAt(latest confirmed)` (`sign_drain.go`). A normal outbound that is broadcast-but-unmined at the pinned nonce yields `live == pinned`, so the drain proceeds and either hits "replacement underpriced" or (if higher fee) evicts the legit outbound. Use `PendingNonceAt` so the guard sees the mempool nonce.

**3. [Low] BTC RBF flag buys nothing operationally.**
`buildSweep` sets `Sequence=1` but the poller never issues a fee-bumped replacement and the fee is pinned at build time — a sweep stuck below the going rate can only be rescued by republishing at a higher trigger height.

## By-design limits — please put in the runbook
- **done ≠ confirmed**, and the pinned tx **cannot be fee-bumped** (byte-identical re-sign). To rescue a stuck tx the operator must **republish at a higher trigger height with a fresh fee**. BTC default fee stays 50 sat/vB — set `--fee-rate` from the live mempool at drain time.
- **Partial-skip is operator-gated only:** a funded chain dropped because its RPC blipped at generation time is simply absent from the final, and the poller reports the payload complete. Empty-final is blocked, but a funded-but-unreachable chain is caught only by the stderr skipped-chains summary → the operator must verify it before arming.

Not blocking — the tool is a fail-closed skeleton and this is heavily reviewed. These are the things to land before the real anchors go in.

kingpinXD:
Fixed all three in `4c962fc6f`:

**F1 cron abort.** Draft-tick errors are logged and the loop continues, and the final keeps retrying on later ticks, so only ctx cancellation ends it. Errors surface through a required `onError` callback rather than a logger, since `zetatool` already reports via stderr WARN/ERROR at ~20 sites and a second channel would have split that stream. Two new tests, both red-checked against the old semantics.

**F2 pending nonce.** The signer's `EVMClient` now declares `PendingNonceAt` and `PendingNonce` uses it. Widening the interface was enough, `client.Client` already embeds `*ethclient.Client`. Worth flagging the consequence you get with it: the drain now declines to fire while any outbound is in flight at or after the pinned nonce, and retries next tick. Since each node has its own mempool view, nodes can briefly disagree on fire/no-fire, converging once the pending tx mines. The generator still pins from confirmed state so the payload stays deterministic.

**F3 RBF.** Documented that `rbfSequenceNum=1` buys no in-place fee bump and that recovery is republish at a higher trigger height with a fresh `--fee-rate`. Value and assertion unchanged.

**Runbook.** Folded both by-design limits into the RFC, extending the existing "Broadcast is not confirmed" callout with the byte-identical re-sign constraint and the 50 sat/vB default, plus a new callout for the partial-skip gate and the stderr skipped-chains check before arming.

All three fixes are re-backported to #4614 and #4615 as you asked. The shared files are byte-identical across the three branches.

julianrubino (approved):
Approve. Re-verified the current head (`4c962fc6f`) closes both actionable findings from my review:

- **Cron abort (Medium):** `RunCron`'s `tick()` now logs via `onError` and `return false` on a failing tick — the loop continues and only returns once a final has actually published. A transient zetacore error no longer tears down the server before the drain fires. ✅
- **Nonce guard (Low/Med):** `Signer.PendingNonce` now uses `PendingNonceAt` (mempool-aware), so a broadcast-but-unmined outbound at the pinned nonce is seen. ✅

Anti-redirect anchor intact (signers use the compiled receiver, never `payload.To`), prod anchors still fail closed. The remaining items (done ≠ confirmed, pinned tx can't be fee-bumped → republish higher, BTC RBF flag is a no-op operationally) are by-design — please keep them in the runbook. Thanks @kingpinXD for the quick turnaround. LGTM.

PR #4613, opened 2026-07-23
feat(snapshot): add ZETA snapshot tooling (state export + per-address balance compute)

kingpinXD:
## Summary
- **Stage 1, export** (`contrib/snapshot/export_snapshot.py` + `make snapshot-export`): produces a `zetacored export` genesis JSON for testnet or mainnet. Reuses `download_snapshot.py` for the fetch and skips the node sync a devnet fork needs, since `export` reads committed state straight from the data dir. Strips the logger preamble so the output is valid JSON.
- **Stage 2, compute** (`zetatool snapshot`): a pure, deterministic function over that export. Attributes `liquid + staked + unbonding` per address (slashing-aware via `TokensFromShares`), routes module accounts, the zero address and pinned contracts to a single remainder, and emits Postgres-ready CSV + `schema.sql`. `remainder = supply - attributed`, so nothing is dropped and the books balance by construction.
- **Contract pinning**: contracts hold native ZETA no keypair can claim, so they are pinned out of attribution. `--pin-file` reads them from a file, and `contrib/snapshot/pins_mainnet.txt` carries the 16 mainnet contracts found by an `eth_getCode` scan of the top-500 holders (9,351,261 ZETA). A pin that matches no holder now fails the run, which is what catches a typo or a wrong-network pin file.
- Verified against a real testnet export at height 22052574: sum of 2,411,443 balances equals recorded bank supply equals live supply, to the wei. Mainnet export at height 17619782 also produced and reconciled.
- Offline tooling only. No changes to consensus, modules, or runtime behavior.

`contrib/snapshot/README.md` has the exact commands for both stages, the flags, and a Verification status section listing what is and is not proven yet.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/zetatool/cli/snapshot.go
+++ b/cmd/zetatool/cli/snapshot.go
@@ -166,8 +166,9 @@ func readPinFile(path string) ([]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("read pin file %q: %w", path, err)
 	}
-	var pins []string
-	for _, line := range strings.Split(string(raw), "\n") {
+	lines := strings.Split(string(raw), "\n")
+	pins := make([]string, 0, len(lines))
+	for _, line := range lines {
 		if i := strings.Index(line, "#"); i >= 0 {
 			line = line[:i]
 		}
--- a/cmd/zetatool/cli/snapshot.go
+++ b/cmd/zetatool/cli/snapshot.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	sdkmath "cosmossdk.io/math"
 	"github.com/spf13/cobra"
@@ -19,6 +20,7 @@ const (
 	flagSnapshotOutput      = "output"
 	flagSnapshotChainID     = "chain-id"
 	flagSnapshotPin         = "pin"
+	flagSnapshotPinFile     = "pin-file"
 	flagSnapshotWZeta       = "wzeta"
 	flagSnapshotSupplyCheck = "supply-check"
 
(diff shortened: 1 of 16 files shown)


ws4charlie:
can you pls check the cursor + greptile comments? the stale-cache one especially — download_snapshot.py exits 0 when a cache already exists, so a rerun without --force silently exports the old height.

contrib/snapshot/export_snapshot.py
```diff
@@ -0,0 +1,278 @@
...
+    if args.skip_download:
+        print("--skip-download set; reusing existing snapshot cache.")
+    else:
+        run_command(
+            f"python3 {DOWNLOAD_SNAPSHOT_SCRIPT} --chain-id {chain_id}",
+            dry_run=args.dry_run,
+        )
+
+    data_dir = home / "data"
+    run_command(f'rm -rf "{data_dir}"', dry_run=args.dry_run)
```
ws4charlie, line 195: This does `rm -rf $home/data` on whatever `--home` points at, so an operator who reuses their real node home to avoid the 39 GB copy loses that node's data dir with no prompt and no way back. Could we bail out when `$home/config/priv_validator_key.json` exists (or when home resolves to `~/.zetacored`) unless they pass an explicit `--force`?

cmd/zetatool/cli/snapshot.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	pins = append(pins, wzeta...)
+	if pinFile != "" {
+		filePins, err := readPinFile(pinFile)
+		if err != nil {
+			return err
+		}
+		pins = append(pins, filePins...)
+	}
```
ws4charlie, line 110: If a mainnet export gets run through here without `--pin-file`, `pins` stays empty and all four checks still pass — the pin-match check at line 245 only catches pins that matched *nothing* — so WZETA's 6.6M ZETA lands as a claimable `eoa` row instead of the remainder. Could we fail closed when `--chain-id` is 7000 and no pins were supplied, unless the operator explicitly opts out?

PR #4618, opened 2026-08-12
fix(zetaclient): keep signing when the keygen record is reset

kingpinXD:
## Summary

- zetacore blanks the keygen record on any observer set change (`x/observer/abci.go` BeginBlocker). The struct literal it writes zeroes the grantee list and resets the status to pending at block `MaxInt64`.
- zetaclient treated that record as a gate in three places, so a routine validator unbonding took every mainnet signer down at once and none could start again.
- Prefer the finalized TSS instead: whitelist from `TSS.TssParticipantList`, skip the ceremony when a key already exists, and stop watching the keygen record for restarts.
- Dry mode (`start.go`) already derived everything from the TSS object this way.

## What happened on mainnet

An observer's validator was jailed and began unbonding. The staking hook dropped it from the observer set, and the next BeginBlocker blanked the keygen. `waitForNewKeygen` read that as a new keygen and shut every signer down; each then failed to restart because an empty grantee list meant an empty p2p whitelist.

No transaction was involved. Signing itself never reads the keygen record: `go-tss` takes both the threshold and the participant set from the local key share file, so the existing key stayed perfectly usable throughout.

## Behaviour change worth flagging: scheduled keygen rotation is disabled while a finalized key exists

An earlier version of this description said rotation was unaffected. That was wrong, and the reasoning was circular — thanks @julianrubino for catching it.

`KeygenCeremony` is only called from `Setup`, and `waitForNewKeygen` was the only thing that restarted zetaclient when a keygen was scheduled. With the ceremony skipped and the watcher gone, a `MsgUpdateKeygen` will not start a ceremony, so no new key reaches TSS history, so `waitForNewKeyGeneration` never fires.

That is deliberate — waiting on the record is exactly what strands the signer once it has been reset — but it must not be discovered by an operator mid-rotation. Rotating on purpose means removing the current TSS first, or reintroducing a deliberate trigger. Both `setup.go` step 5 and the `Listen` comment now say so.

Genesis and first keygen are unaffected: with no TSS on chain the keygen record is still authoritative.

## Tests

- `resolveTSSPeers` unit tests cover the blanked record, a scheduled keygen, a successful keygen, and the no-TSS-yet case. Mutation-checked: reverting the guard fails three of the four.
- New `tss_listener_test.go` (the listener had no tests). Covers no-shutdown on a blanked record, and that both TSS-address and TSS-history changes still shut down. Mutation-checked: re-adding a keygen watcher fails it.
- New isolated e2e, `make start-keygen-reset-test`. It resets the record, then asserts an outbound still signs under the same TSS key.

The e2e uses `MsgAddObserver` only as a **stand-in** — its handler writes byte-for-byte the same state as the BeginBlocker, in one transaction, without needing to drive a validator through jailing and unbonding. No add-observer transaction was involved in the incident. The test comment says so explicitly.

## Compatibility with the emergency drain (#4612)

Verified compatible. The drain has no keygen references, and its deterministic leader election comes from `localStateItem.ParticipantKeys` in the key share, not the peer whitelist.

Worth noting: `startDrainIfArmed` sits after `zetatss.Setup` in `start.go`, so while `Setup` fails the drain cannot start either. This PR restores that lever.

## Hotfix

Targets `release/zetaclient/v37` and `release/zetaclient/v38` to follow, same commit. Both touched files are byte-identical across all three branches.

https://claude.ai/code/session_01Cd1ZRjpN5br7NRYA9sCp7G

---

> [!NOTE]
> **High Risk**
> Changes TSS startup, P2P whitelisting, and restart triggers in security-critical signing path; deliberate behavior change skips scheduled keygen at startup while a finalized key exists.
> 
> **Overview**
> Fixes a mainnet-class outage where observer set changes blank the on-chain keygen record (empty grantees, pending at `MaxInt64`) while the finalized TSS key is unchanged.
> 
> **zetaclient** now treats a finalized TSS as authoritative: P2P whitelist comes from `TSS.TssParticipantList`, startup skips the keygen ceremony when a key already exists, and the TSS listener no longer watches the keygen record for restarts (only TSS pubkey changes and new entries in TSS history still trigger restart). Real key rotation behavior is unchanged.
> 
> Adds unit tests for `resolveTSSPeers` and the listener, plus an isolated e2e (`make start-keygen-reset-test` / `--keygen-reset-test`) that reproduces the blanked keygen state and asserts an outbound still signs under the same TSS key.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f705c5650078bf83e648097b88db9e833c10fc98. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR keeps zetaclient operational when observer-set changes blank the keygen record by treating the finalized TSS as authoritative.

- Derives the P2P whitelist from finalized TSS participants and skips an unnecessary keygen ceremony.
- Removes keygen-record changes as a zetaclient restart trigger while retaining current-TSS and TSS-history watchers.
- Adds unit and isolated end-to-end coverage for continued signing after a keygen reset.

The PR appears safe to merge, with no actionable defects identified in the changed signing, startup, listener, or test paths.

The finalized TSS remains authoritative after a keygen-record reset, peer selection retains the original participant set, and the remaining listener triggers continue to respond to actual TSS changes; the deliberate effect on scheduled rotations is clearly documented.

| Filename | Overview |
|----------|----------|
| zetaclient/tss/setup.go | Makes the finalized TSS authoritative for peer selection and ceremony decisions, with explicit handling for missing peer data and TSS-query failures. |
| zetaclient/maintenance/tss_listener.go | Removes keygen-record polling while preserving restart triggers for current-TSS and TSS-history changes. |
| zetaclient/tss/setup_test.go | Covers peer resolution for initial keygen, blanked and scheduled records, genesis imports, and successful keygen state. |
| zetaclient/maintenance/tss_listener_test.go | Verifies that keygen resets are ignored while TSS address and history changes still trigger shutdown. |
| e2e/e2etests/test_keygen_reset_signing.go | Reproduces the reset state through the administrative handler and verifies that an outbound remains signable with the existing TSS key. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[zetaclient Setup] --> B[Read keygen record]
    B --> C[Read finalized TSS with retry]
    C --> D{Finalized TSS exists?}
    D -->|Yes| E[Whitelist TSS participants]
    E --> F[Skip keygen ceremony]
    D -->|No| G[Whitelist keygen grantees]
    G --> H[Run keygen ceremony]
    F --> I[Verify key shares and start signing]
    H --> I
    J[Observer set changes] --> K[Keygen record reset]
    K --> L[TSS listener ignores reset]
    L --> I
    M[TSS address or history changes] --> N[Trigger zetaclient restart]
```

<sub>Reviews (1): Last reviewed commit: ["test(zetaclient): stop the listener test..."](https://github.com/zeta-chain/node/commit/ba0af06dd0995c48f9278a4ca1afeb402266bba9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52790838)</sub>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,6 +331,13 @@ start-drain-test: e2e-images
 	export ZETACLIENT_DRAIN_WINDOW="500" && \
 	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile dry up -d
 
+# start-keygen-reset-test runs ONLY the keygen reset e2e test in isolation. It leaves the
+# keygen record erased, so it must not share a localnet with tests that expect a populated one.
+start-keygen-reset-test: e2e-images
+	@echo "--> Starting keygen reset e2e test"
+	export E2E_ARGS="--keygen-reset-test" && \
+	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile dry up -d
+
 start-e2e-test-4nodes: e2e-images
 	@echo "--> Starting e2e test with 4 nodes"
 	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile stress up -d
--- a/changelog.md
+++ b/changelog.md
@@ -14,7 +14,7 @@
 
 ### Fixes
 
-* [4618](https://github.com/zeta-chain/node/pull/4618) - keep zetaclient signing when zetacore resets the keygen record on an observer set change, instead of shutting every signer down and refusing to restart.
+* [4618](https://github.com/zeta-chain/node/pull/4618) - keep zetaclient signing when zetacore resets the keygen record on an observer set change, instead of shutting every signer down and refusing to restart. Note for operators: while a finalized TSS exists, a keygen scheduled via `MsgUpdateKeygen` will no longer run — rotating requires removing the current TSS first.
 * [4532](https://github.com/zeta-chain/node/pull/4532) - remove unnessary tests from the CI and fix e2e-performance-test-1k.
 * [4538](https://github.com/zeta-chain/node/pull/4538) - fix missed inbound caused by early closure of error monitor channel
 * [4534](https://github.com/zeta-chain/node/pull/4534) - fix Sui nightly e2e deposit test failure by reducing gas budget to 500000000
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@
 
 ### Fixes
 
+* [4618](https://github.com/zeta-chain/node/pull/4618) - keep zetaclient signing when zetacore resets the keygen record on an observer set change, instead of shutting every signer down and refusing to restart.
 * [4532](https://github.com/zeta-chain/node/pull/4532) - remove unnessary tests from the CI and fix e2e-performance-test-1k.
 * [4538](https://github.com/zeta-chain/node/pull/4538) - fix missed inbound caused by early closure of error monitor channel
 * [4534](https://github.com/zeta-chain/node/pull/4534) - fix Sui nightly e2e deposit test failure by reducing gas budget to 500000000
--- a/e2e/e2etests/test_keygen_reset_signing.go
+++ b/e2e/e2etests/test_keygen_reset_signing.go
@@ -1,0 +1,107 @@
+package e2etests
+
+import (
+	"math"
+	"math/big"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zeta-chain/protocol-contracts-evm/pkg/gatewayzevm.sol"
+
+	"github.com/zeta-chain/node/e2e/runner"
+	"github.com/zeta-chain/node/e2e/utils"
+	crosschaintypes "github.com/zeta-chain/node/x/crosschain/types"
+	observertypes "github.com/zeta-chain/node/x/observer/types"
+)
+
+// TestKeygenResetSigning asserts that resetting the keygen record does not stop the network
+// from signing with the TSS key it already has.
+//
+// # HOW THIS REPRODUCES THE MAINNET INCIDENT — READ BEFORE CHANGING
+//
+// On mainnet nobody sent a transaction. An observer's validator was jailed and began
+// unbonding, the staking hook dropped it from the observer set, and the next BeginBlocker
+// (x/observer/abci.go) saw the observer count change and ran:
+//
+//	k.DisableInboundOnly(ctx)
+//	k.SetKeygen(ctx, types.Keygen{BlockNumber: math.MaxInt64})
+//
+// That struct literal zeroes every field it does not name, so the grantee list was erased and
+// the status reset to PendingKeygen. Every zetaclient then restarted and none could start
+// again: an empty grantee list meant an empty p2p whitelist.
+//
+// Driving a validator through jailing and unbonding inside an e2e test is slow and fragile.
+// MsgAddObserver with AddNodeAccountOnly is used INSTEAD because its handler runs that exact
+// same pair of calls (msg_server_add_observer.go), so it writes byte-for-byte the state the
+// BeginBlocker wrote, in one transaction. It is a stand-in for convenience.
+//
+// It is NOT what happened on mainnet, and no add-observer transaction was involved there. Do
+// not read this test as evidence of the cause.
+//
+// The message targets an observer that already exists, so SetNodeAccount rewrites an identical
+// record and the observer set is untouched. The only effects are the keygen reset and the
+// inbound pause, which is what we want to isolate.
+func TestKeygenResetSigning(r *runner.E2ERunner, args []string) {
+	require.Len(r, args, 1)
+
+	amount := utils.ParseBigInt(r, args[0])
+
+	tssBefore, err := r.ObserverClient.TSS(r.Ctx, &observertypes.QueryGetTSSRequest{})
+	require.NoError(r, err)
+	require.NotEmpty(r, tssBefore.TSS.TssPubkey, "test needs a TSS that has already been generated")
+
+	keygenBefore, err := r.ObserverClient.Keygen(r.Ctx, &observertypes.QueryGetKeygenRequest{})
+	require.NoError(r, err)
+	require.NotNil(r, keygenBefore.Keygen)
+	require.NotEmpty(
+		r,
+		keygenBefore.Keygen.GranteePubkeys,
+		"keygen record must start populated, otherwise this test proves nothing",
+	)
+
+	// Reuse an existing node account so the message adds nobody and changes no observer.
+	nodeAccounts, err := r.ObserverClient.NodeAccountAll(r.Ctx, &observertypes.QueryAllNodeAccountRequest{})
+	require.NoError(r, err)
+	require.NotEmpty(r, nodeAccounts.NodeAccount)
+
+	existing := nodeAccounts.NodeAccount[0]
+	require.NotNil(r, existing.GranteePubkey)
+
+	msgReset := observertypes.NewMsgAddObserver(
+		r.ZetaTxServer.MustGetAccountAddressFromName(utils.AdminPolicyName),
+		existing.Operator,
+		existing.GranteePubkey.Secp256k1.String(),
+		true,
+	)
+	_, err = r.ZetaTxServer.BroadcastTx(utils.AdminPolicyName, msgReset)
+	require.NoError(r, err)
+
+	// The record is now in the state that took mainnet down.
+	keygenAfter, err := r.ObserverClient.Keygen(r.Ctx, &observertypes.QueryGetKeygenRequest{})
+	require.NoError(r, err)
+	require.NotNil(r, keygenAfter.Keygen)
+	require.Empty(r, keygenAfter.Keygen.GranteePubkeys, "grantee list should have been erased")
+	require.Equal(r, observertypes.KeygenStatus_PendingKeygen, keygenAfter.Keygen.Status)
+	require.Equal(r, int64(math.MaxInt64), keygenAfter.Keygen.BlockNumber)
+
+	// The same handler pauses inbound, so restore it before expecting a deposit to flow.
+	reEnableInbound(r)
+
+	// The real assertion: a withdraw needs a TSS keysign, so completing one proves the
+	// signers stayed up and kept using the key they already held.
+	r.ApproveETHZRC20(r.GatewayZEVMAddr)
+	tx := r.ETHWithdraw(r.EVMAddress(), amount, gatewayzevm.RevertOptions{OnRevertGasLimit: big.NewInt(0)})
+
+	cctx := utils.WaitCctxMinedByInboundHash(r.Ctx, tx.Hash().Hex(), r.CctxClient, r.Logger, r.CctxTimeout)
+	r.Logger.CCTX(*cctx, "withdraw after keygen reset")
+	utils.RequireCCTXStatus(r, cctx, crosschaintypes.CctxStatus_OutboundMined)
+
+	// Signed by the original key, not a replacement.
+	tssAfter, err := r.ObserverClient.TSS(r.Ctx, &observertypes.QueryGetTSSRequest{})
+	require.NoError(r, err)
+	require.Equal(r, tssBefore.TSS.TssPubkey, tssAfter.TSS.TssPubkey, "TSS key must not have rotated")
+
+	// Still reset: nothing repaired the record, so signing survived it rather than dodged it.
+	keygenEnd, err := r.ObserverClient.Keygen(r.Ctx, &observertypes.QueryGetKeygenRequest{})
+	require.NoError(r, err)
+	require.Empty(r, keygenEnd.Keygen.GranteePubkeys)
+}
--- a/e2e/e2etests/test_keygen_reset_signing.go
+++ b/e2e/e2etests/test_keygen_reset_signing.go
@@ -45,6 +45,13 @@ func TestKeygenResetSigning(r *runner.E2ERunner, args []string) {
 
 	amount := utils.ParseBigInt(r, args[0])
 
+	// Fund the deployer while the network is still healthy, so the withdraw later is the only
+	// thing under test. This doubles as the "before" half of the comparison: an inbound that
+	// works now, and an outbound that must still work once the record is erased.
+	depositHash := r.DepositEtherToDeployer()
+	depositCCTX := utils.WaitCctxMinedByInboundHash(r.Ctx, depositHash.Hex(), r.CctxClient, r.Logger, r.CctxTimeout)
+	utils.RequireCCTXStatus(r, depositCCTX, crosschaintypes.CctxStatus_OutboundMined)
+
 	tssBefore, err := r.ObserverClient.TSS(r.Ctx, &observertypes.QueryGetTSSRequest{})
 	require.NoError(r, err)
 	require.NotEmpty(r, tssBefore.TSS.TssPubkey, "test needs a TSS that has already been generated")
--- a/zetaclient/maintenance/tss_listener.go
+++ b/zetaclient/maintenance/tss_listener.go
@@ -32,6 +32,12 @@ func NewTSSListener(client ZetacoreClient, logger zerolog.Logger) *TSSListener {
 }
 
 // Listen listens for any maintenance regarding TSS and calls action specified. Works in the background.
+//
+// Both watchers key off the TSS itself: the address changing, or a new key landing in history.
+// The keygen record is deliberately not watched. It is reset to "pending at block MaxInt64" on
+// any observer set change, and restarting on that reset is what turns a routine validator
+// unbonding into an outage — every signer shuts down at once and none of them can start again.
+// A real rotation still lands a new key in TSS history, so waitForNewKeyGeneration covers it.
 func (tl *TSSListener) Listen(ctx context.Context, action func()) {
 	var (
 		withLogger = bg.WithLogger(tl.logger)
@@ -40,7 +46,6 @@ func (tl *TSSListener) Listen(ctx context.Context, action func()) {
 
 	bg.Work(ctx, tl.waitForUpdate, bg.WithName("tss.wait_for_update"), withLogger, onComplete)
 	bg.Work(ctx, tl.waitForNewKeyGeneration, bg.WithName("tss.wait_for_generation"), withLogger, onComplete)
-	bg.Work(ctx, tl.waitForNewKeygen, bg.WithName("tss.wait_for_keygen"), withLogger, onComplete)
 }
 
 // waitForUpdate listens for TSS updates. Returns `nil` when the TSS address is updated
@@ -122,51 +127,6 @@ func (tl *TSSListener) waitForNewKeyGeneration(ctx context.Context) error {
 			return nil
 		case <-ctx.Done():
 			tl.logger.Info().Msg("stopped waiting for new key generation in the TSS listener")
-			return nil
-		}
-	}
-}
-
-// waitForNewKeygen is a background thread that listens for new keygen; it returns when a new keygen is set
-func (tl *TSSListener) waitForNewKeygen(ctx context.Context) error {
-	// Initial Keygen retrieval
-	keygen, err := tl.client.GetKeyGen(ctx)
-	if err != nil {
-		return errors.Wrap(err, "failed to get initial TSS history")
-	}
-
-	ticker := time.NewTicker(tssListenerTicker)
-	defer ticker.Stop()
-
-	for {
-		select {
-		case <-ticker.C:
-			keygenUpdated, err := tl.client.GetKeyGen(ctx)
-			switch {
-			case err != nil:
-				tl.logger.Warn().Err(err).Msg("unable to get keygen")
-				continue
-			// Keygen is not pending it has already been successfully generated, continue loop
-			case keygenUpdated.Status == observertypes.KeygenStatus_KeyGenSuccess:
-				continue
-			// Keygen failed we to need to wait until a new keygen is set, continue loop
-			case keygenUpdated.Status == observertypes.KeygenStatus_KeyGenFailed:
-				continue
-			// Keygen is pending but block number is not updated, continue loop.
-			// Most likely the zetaclient is waiting for the keygen block to arrive.
-			case keygenUpdated.Status == observertypes.KeygenStatus_PendingKeygen &&
-				keygenUpdated.BlockNumber <= keygen.BlockNumber:
-				continue
-			}
-
-			// Trigger restart only when the following conditions are met:
-			// 1. Keygen is pending
-			// 2. Block number is updated
-
-			tl.logger.Info().Int64("block_number", keygenUpdated.BlockNumber).Msg("got new keygen")
-			return nil
-		case <-ctx.Done():
-			tl.logger.Info().Msg("stopped waiting for new keygen in the TSS listener")
 			return nil
 		}
 	}
--- a/zetaclient/maintenance/tss_listener.go
+++ b/zetaclient/maintenance/tss_listener.go
@@ -37,7 +37,10 @@ func NewTSSListener(client ZetacoreClient, logger zerolog.Logger) *TSSListener {
 // The keygen record is deliberately not watched. It is reset to "pending at block MaxInt64" on
 // any observer set change, and restarting on that reset is what turns a routine validator
 // unbonding into an outage — every signer shuts down at once and none of them can start again.
-// A real rotation still lands a new key in TSS history, so waitForNewKeyGeneration covers it.
+//
+// Note this also removes the only trigger that restarted zetaclient for a scheduled keygen, so
+// while a finalized key exists a rotation ceremony will not start. That is deliberate; see the
+// step 5 comment in zetaclient/tss/setup.go.
 func (tl *TSSListener) Listen(ctx context.Context, action func()) {
 	var (
 		withLogger = bg.WithLogger(tl.logger)
--- a/zetaclient/maintenance/tss_listener.go
+++ b/zetaclient/maintenance/tss_listener.go
@@ -13,7 +13,9 @@ import (
 	observertypes "github.com/zeta-chain/node/x/observer/types"
 )
 
-const tssListenerTicker = 5 * time.Second
+// tssListenerTicker is how often the watchers re-query zetacore. A var rather than a const so
+// tests can shrink it; nothing in production writes to it.
+var tssListenerTicker = 5 * time.Second
 
 // TSSListener is a struct that listens for TSS updates, new keygen, and new TSS key generation.
 type TSSListener struct {
--- a/zetaclient/maintenance/tss_listener_test.go
+++ b/zetaclient/maintenance/tss_listener_test.go
@@ -2,6 +2,7 @@ package maintenance
 
 import (
 	"context"
+	"io"
 	"math"
 	"testing"
 	"time"
@@ -24,7 +25,9 @@ func waitForListenerTick() {
 }
 
 func TestTSSListener(t *testing.T) {
-	logger := zerolog.New(zerolog.NewTestWriter(t))
+	// Deliberately not zerolog.NewTestWriter(t): Listen's workers keep running after the test
+	// returns, and logging into a finished t panics with "Log in goroutine after test completed".
+	logger := zerolog.New(io.Discard)
 
 	oldTSS := observertypes.TSS{TssPubkey: tssPubkeyOld}
 
@@ -89,7 +92,9 @@ func TestTSSListenerIgnoresKeygen(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	logger := zerolog.New(zerolog.NewTestWriter(t))
+	// Deliberately not zerolog.NewTestWriter(t): Listen's workers keep running after the test
+	// returns, and logging into a finished t panics with "Log in goroutine after test completed".
+	logger := zerolog.New(io.Discard)
 	oldTSS := observertypes.TSS{TssPubkey: tssPubkeyOld}
 
 	client := mocks.NewZetacoreClient(t)
--- a/zetaclient/maintenance/tss_listener_test.go
+++ b/zetaclient/maintenance/tss_listener_test.go
@@ -1,0 +1,113 @@
+package maintenance
+
+import (
+	"context"
+	"math"
+	"testing"
+	"time"
+
+	"github.com/rs/zerolog"
+
+	observertypes "github.com/zeta-chain/node/x/observer/types"
+	"github.com/zeta-chain/node/zetaclient/testutils/mocks"
+)
+
+const (
+	tssPubkeyOld = "zetapub1addwnpepqtadxdyt037h86z60nl98t6zk56mw5zpnm79tsmvspln3hgt5phdc79kvfc"
+	tssPubkeyNew = "zetapub1addwnpepqglunjrgl3qg08duxq9pf28jmvrer3crwnnfzp6m0u0yh9jk9mnn5p76utc"
+)
+
+// waitForListenerTick gives the listener room for at least one tick so a shutdown it was
+// going to signal has actually had the chance to fire.
+func waitForListenerTick() {
+	time.Sleep(2 * tssListenerTicker)
+}
+
+func TestTSSListener(t *testing.T) {
+	logger := zerolog.New(zerolog.NewTestWriter(t))
+
+	oldTSS := observertypes.TSS{TssPubkey: tssPubkeyOld}
+
+	// A blanked keygen record must not restart the client.
+	//
+	// zetacore writes exactly this on any observer set change (x/observer/abci.go
+	// BeginBlocker): status back to pending, grantees erased, block set to MaxInt64. On
+	// mainnet that reset restarted every signer at once, and none could start again because
+	// the erased grantee list left them with an empty p2p whitelist.
+	t.Run("blanked keygen record does not trigger a shutdown", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		client := mocks.NewZetacoreClient(t)
+		client.Mock.On("GetTSS", ctx).Return(oldTSS, nil)
+		client.Mock.On("GetTSSHistory", ctx).Return([]observertypes.TSS{oldTSS}, nil)
+
+		complete := make(chan interface{})
+		NewTSSListener(client, logger).Listen(ctx, func() { close(complete) })
+
+		waitForListenerTick()
+		assertChannelNotClosed(t, complete)
+	})
+
+	t.Run("TSS address change still triggers a shutdown", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		client := mocks.NewZetacoreClient(t)
+		client.Mock.On("GetTSSHistory", ctx).Return([]observertypes.TSS{oldTSS}, nil)
+		client.Mock.On("GetTSS", ctx).Return(oldTSS, nil).Once()
+		client.Mock.On("GetTSS", ctx).Return(observertypes.TSS{TssPubkey: tssPubkeyNew}, nil)
+
+		complete := make(chan interface{})
+		NewTSSListener(client, logger).Listen(ctx, func() { close(complete) })
+
+		<-complete
+	})
+
+	t.Run("new key in TSS history still triggers a shutdown", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		newTSS := observertypes.TSS{TssPubkey: tssPubkeyNew}
+
+		client := mocks.NewZetacoreClient(t)
+		client.Mock.On("GetTSS", ctx).Return(oldTSS, nil)
+		client.Mock.On("GetTSSHistory", ctx).Return([]observertypes.TSS{oldTSS}, nil).Once()
+		client.Mock.On("GetTSSHistory", ctx).Return([]observertypes.TSS{oldTSS, newTSS}, nil)
+
+		complete := make(chan interface{})
+		NewTSSListener(client, logger).Listen(ctx, func() { close(complete) })
+
+		<-complete
+	})
+}
+
+// TestTSSListenerIgnoresKeygen pins that the listener never reads the keygen record. The mock
+// fails the test on any unexpected call, so a reintroduced keygen watcher shows up here rather
+// than as an outage.
+func TestTSSListenerIgnoresKeygen(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	logger := zerolog.New(zerolog.NewTestWriter(t))
+	oldTSS := observertypes.TSS{TssPubkey: tssPubkeyOld}
+
+	client := mocks.NewZetacoreClient(t)
+	client.Mock.On("GetTSS", ctx).Return(oldTSS, nil)
+	client.Mock.On("GetTSSHistory", ctx).Return([]observertypes.TSS{oldTSS}, nil)
+
+	NewTSSListener(client, logger).Listen(ctx, func() {})
+	waitForListenerTick()
+
+	client.Mock.AssertNotCalled(t, "GetKeyGen", ctx)
+
+	// Guard against the assertion above passing for the wrong reason: the record a caller
+	// would have read is the blanked one, and it must remain irrelevant.
+	blanked := observertypes.Keygen{
+		Status:      observertypes.KeygenStatus_PendingKeygen,
+		BlockNumber: math.MaxInt64,
+	}
+	if len(blanked.GranteePubkeys) != 0 {
+		t.Fatal("expected the blanked keygen record to carry no grantees")
+	}
+}
--- a/zetaclient/maintenance/tss_listener_test.go
+++ b/zetaclient/maintenance/tss_listener_test.go
@@ -3,7 +3,6 @@ package maintenance
 import (
 	"context"
 	"io"
-	"math"
 	"testing"
 	"time"
 
@@ -18,10 +17,19 @@ const (
 	tssPubkeyNew = "zetapub1addwnpepqglunjrgl3qg08duxq9pf28jmvrer3crwnnfzp6m0u0yh9jk9mnn5p76utc"
 )
 
+// useFastTicker shrinks the watcher poll interval for the duration of one test. The assertions
+// are about which events cause a shutdown, not about the real 5s cadence, so waiting on it only
+// bought ~31s of sleep across this file.
+func useFastTicker(t *testing.T) {
+	original := tssListenerTicker
+	tssListenerTicker = 10 * time.Millisecond
+	t.Cleanup(func() { tssListenerTicker = original })
+}
+
 // waitForListenerTick gives the listener room for at least one tick so a shutdown it was
 // going to signal has actually had the chance to fire.
 func waitForListenerTick() {
-	time.Sleep(2 * tssListenerTicker)
+	time.Sleep(20 * tssListenerTicker)
 }
 
 func TestTSSListener(t *testing.T) {
@@ -38,6 +46,8 @@ func TestTSSListener(t *testing.T) {
 	// mainnet that reset restarted every signer at once, and none could start again because
 	// the erased grantee list left them with an empty p2p whitelist.
 	t.Run("blanked keygen record does not trigger a shutdown", func(t *testing.T) {
+		useFastTicker(t)
+
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
@@ -53,6 +63,8 @@ func TestTSSListener(t *testing.T) {
 	})
 
 	t.Run("TSS address change still triggers a shutdown", func(t *testing.T) {
+		useFastTicker(t)
+
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
@@ -68,6 +80,8 @@ func TestTSSListener(t *testing.T) {
 	})
 
 	t.Run("new key in TSS history still triggers a shutdown", func(t *testing.T) {
+		useFastTicker(t)
+
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
@@ -89,6 +103,8 @@ func TestTSSListener(t *testing.T) {
 // fails the test on any unexpected call, so a reintroduced keygen watcher shows up here rather
 // than as an outage.
 func TestTSSListenerIgnoresKeygen(t *testing.T) {
+	useFastTicker(t)
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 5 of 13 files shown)


zetaclient/tss/setup.go
```diff
@@ -86,8 +86,25 @@ func Setup(ctx context.Context, p SetupProps, logger zerolog.Logger) (*Service,
 
 	setupLogger.Info().Msg("fetched TSS keygen info")
 
-	whitelistedPeers := make([]peer.ID, len(tssKeygen.GranteePubkeys))
-	for i, pk := range tssKeygen.GranteePubkeys {
+	// A finalized TSS takes precedence over the keygen record; see resolveTSSPeers.
+	currentTSS, tssErr := p.Zetacore.GetTSS(ctx)
+	if tssErr != nil {
```
ws4charlie: Might be missing something, but this treats any `GetTSS` error as "no TSS yet" — so a plain timeout here would send us back to the keygen record, which on mainnet today is the blanked one, and we'd be right back at an empty whitelist waiting on the MaxInt64 block. Could we retry with backoff like the listener does and only treat the not-found case as absent?

zetaclient/tss/setup.go
```diff
@@ -86,8 +86,25 @@ func Setup(ctx context.Context, p SetupProps, logger zerolog.Logger) (*Service,
...
+		// Absent on a chain that has never completed a keygen, which is a valid state to
+		// start from, so fall through to the ceremony rather than failing here.
+		setupLogger.Info().Err(tssErr).Msg("no finalized TSS available")
+		currentTSS = observertypes.TSS{}
```
julianrubino: On a `GetTSS` error this silently falls back to the keygen record via `resolveTSSPeers`. If the record is blanked (the exact state this PR fixes) and `GetTSS` transiently errors, you're back to an empty whitelist **and** a ceremony scheduled for `MaxInt64` — both failure modes the PR removes. It's strictly better than the old code (which never consulted `GetTSS`), and a transient error → crash → restart → retry is survivable, but since the whole premise is that the keygen record is unreliable, consider retrying `GetTSS` a few times or failing loudly here rather than trusting the record. At minimum, log this fallback at WARN, not INFO — it means the whitelist source is degraded.

zetaclient/tss/setup.go
```diff
@@ -185,6 +213,27 @@ func Setup(ctx context.Context, p SetupProps, logger zerolog.Logger) (*Service,
...
+// That says nothing about a key generated long ago and whose shares this node still holds, so
+// a finalized TSS wins. Reading the record instead would leave a healthy signer with an empty
+// whitelist and a ceremony scheduled for a block that never arrives.
+//
+// The participant list is the set that produced the key we are about to sign with, which makes
+// it the correct whitelist: it must stay a superset of the participants recorded in the local
+// key share, or peers are refused at the p2p layer during keysign.
+func resolveTSSPeers(keygen observertypes.Keygen, currentTSS observertypes.TSS) (pubkeys []string, finalized bool) {
+	if currentTSS.TssPubkey == "" {
+		return keygen.GranteePubkeys, false
```
julianrubino, line 297: This now makes the p2p whitelist depend on `TssParticipantList` being populated. It is, for any TSS created via `msg_server_vote_tss.go` (set from the keygen grantees). But a TSS imported straight from genesis (`x/observer/genesis.go:94`) with an empty `TssParticipantList` would resolve `finalized=true` with an empty pubkey slice → empty whitelist → the same crash. Not the mainnet path, but cheap to guard: if `currentTSS.TssPubkey != ""` but `TssParticipantList` is empty, either fall back to the keygen grantees or fail with a clear error rather than starting with an empty whitelist.

julianrubino (commented):
Full review — read the diff plus the surrounding TSS setup/listener/keygen paths and the zetacore keygen→TSS flow. **Core fix is correct and well-tested.** One framing point worth correcting in the description, and two robustness notes. Details below + two inline comments.

## Verified sound
- **`resolveTSSPeers` whitelist source is right.** `TSS.TssParticipantList` is populated from `keygen.GetGranteePubkeys()` at TSS creation (`x/observer/keeper/msg_server_vote_tss.go:124`), so for any key made via the normal ceremony it's a non-empty snapshot of the exact participants that produced the key — i.e. a correct superset of the local key-share participants. Empty-TSS → keygen-record path preserves genesis/first-keygen. 
- **Stopping the loop is correct.** Removing `waitForNewKeygen` + skipping the ceremony when a finalized key exists breaks the blanked-record restart storm cleanly; `waitForUpdate` (TSS addr) and `waitForNewKeyGeneration` (TSS history) still cover a genuine key change.
- **Tests are strong** — mutation-checked units, the listener finally has tests, and the isolated e2e reproduces the *exact* zetacore state. The comment that `MsgAddObserver` is a stand-in (not the cause) is exactly the right kind of honesty.
- **Drain compatibility is not just preserved — this is a prerequisite for it.** `startDrainIfArmed` sits after `zetatss.Setup` in `start.go`, so while Setup was failing on mainnet the drain couldn't have started either. This restores that lever. 👍

## 1. Important (framing, not code): "Rotation is unchanged" is not accurate — this disables scheduled keygen rotation
`KeygenCeremony` is only ever called from `Setup` (`zetaclient/tss/setup.go`), and the **removed** `waitForNewKeygen` was the trigger that restarted the client on a newly-scheduled keygen → which is what re-ran `Setup`→`KeygenCeremony` for the new key. With both gone, while a finalized key exists: nothing runs a scheduled rotation ceremony, so no new key lands in TSS history, so `waitForNewKeyGeneration` never fires. Genesis/first keygen still works (empty-TSS path). 

For the current wind-down (deliberately avoiding keygen, moving funds via the drain) this is fine and probably the intent — but the body says *"Rotation is unchanged … the listener still restarts on it,"* which reads as if rotation still works. It doesn't, until a finalized key is gone. Suggest rewording to *"scheduled keygen rotation is disabled while a finalized key exists"* so a future operator doesn't assume otherwise, and note the deliberate re-enable path if a real rotation is ever needed.

## 2. Robustness: the `GetTSS` error fallback re-opens the exact failure mode (inline)
See inline on `setup.go`. On a `GetTSS` error you fall back to the keygen record; if that record is blanked (the incident state) and `GetTSS` blips, you get both symptoms back (empty whitelist + ceremony waiting on `MaxInt64`). Strictly better than before, but since the premise is "the keygen record is unreliable," consider retry/fail-loud instead of silently trusting it.

## 3. Minor: genesis-imported TSS without a participant list (inline)
See inline on `resolveTSSPeers`. The whitelist now hinges on `TssParticipantList` being populated; a TSS imported via `x/observer/genesis.go` without that field would yield an empty whitelist = same crash. Not the mainnet case, worth a guard or a note.

## 4. Nit
`WithMetrics` (`zetaclient/tss/service.go:109`) reads `keygen.GranteePubkeys` for blame labels — during the blanked window those labels go empty for a cycle. Harmless, just flagging.

## 5. `codecov/patch` is red
The isolated e2e path isn't counted in unit coverage; confirm that's the reason and not a genuinely uncovered branch in `resolveTSSPeers`/listener.

Net: approve-worthy once the description reflects that rotation-keygen is disabled (#1). #2/#3 are cheap hardening on a security-critical path and worth doing while we're here.

---

## Precision notes (independently re-verified)

Two clarifications so the above is airtight — neither changes any verdict:

- **#1 mechanism:** the listener's trigger is `graceful.ShutdownNow()` + a **supervised restart** (the process manager brings zetaclient back up), not an in-process restart. Functionally identical for the argument — the restarted `Setup` is what would re-run the ceremony — but worth stating precisely: the removed `waitForNewKeygen` caused a shutdown, and a scheduled rotation now triggers no shutdown at all.
- **#2 crash path:** an empty whitelist doesn't fail at `NewServer` on its own — `NewServer` hard-errors only when the whitelist **and** the bootstrap peers are both empty. On mainnet bootstrap peers are configured, so Setup passes through the empty whitelist and reaches `KeygenCeremony`, which then hangs on the `MaxInt64` block. That matches the observed incident (Setup crash, not a `NewServer` bootstrap-peer error). If bootstrap peers were also empty you'd instead get the "whitelisted peers missing" abort at `NewServer` — still a Setup failure, different line.

zetaclient/tss/setup.go
kingpinXD: Good catch, fixed in `e8f89fbd2`. `GetTSS` is now retried with backoff, and only the not-found code counts as absent — anything else fails loudly rather than falling back. You were right that the old fallback handed us straight back to the blanked record.

zetaclient/tss/setup.go
kingpinXD: Fixed in `e8f89fbd2`, same change as Charlie's note. Retry with backoff, absence matched on the not-found code alone, and a persistent error now returns instead of trusting the record. Went with fail-loud rather than a WARN fallback, since as you say the whole premise is that the record is unreliable.

morde08 (commented):
Approach looks right to me. Whitelisting from `TSS.TssParticipantList` is the correct source — it is written once from the grantee list at `msg_server_vote_tss.go:124`, is never touched by the BeginBlocker, and is exactly the set that produced the key in use (9 entries on mainnet today), so it is a correct superset for p2p admission. This restores full function without a governance action or a ceremony, which is strictly better than the `MsgUpdateKeygen` route in gov-ops#1184.

Two things before this goes to the release branches — one substantive, one a description fix.

---

**1. `GetTSS` failure falls open into the original outage**

```go
currentTSS, tssErr := p.Zetacore.GetTSS(ctx)
if tssErr != nil {
    setupLogger.Info().Err(tssErr).Msg("no finalized TSS available")
    currentTSS = observertypes.TSS{}
}
```

Any error — including a transient RPC blip — is treated as "no TSS", so `resolveTSSPeers` falls back to `keygen.GranteePubkeys`. On mainnet today that is empty, which means an empty whitelist and the exact crash this PR fixes. `GetTSSHistory` two lines down fails closed and is loud; this one fails open into the bad state, silently, at `Info`.

Suggest wrapping it in `retry.DoTypedWithBackoffAndRetry` (as `waitForNewKeyGeneration` already does for `GetTSSHistory`) and only treating a genuine not-found as the empty case. This matters most during a coordinated restart against a busy zetacore — which is precisely the rollout for this fix.

Minor, same function: `resolveTSSPeers` uses `TssPubkey != ""` as the proxy for "usable participant list". Guarding on `len(currentTSS.TssParticipantList) > 0` instead makes it total — otherwise a finalized TSS with an empty participant list yields an empty whitelist and the same crash.

---

**2. "Rotation is unchanged" contradicts the code**

`KeygenCeremony` is called from exactly one place (`zetaclient/tss/setup.go:122`) and `Setup` from exactly one (`cmd/zetaclientd/start.go:146`, at startup). With `waitForNewKeygen` removed, a running client never restarts on a scheduled keygen; and with the ceremony skipped whenever a finalized TSS exists, a manually restarted one will not run it either. The two remaining watchers are both reactive — `TssPubkey` changing, and TSS history length growing — and both only fire *after* a ceremony has already succeeded. So `MsgUpdateKeygen` becomes inert and there is no remaining path to rotate the key.

The tests make clear this is intended — `"finalized key wins over a populated keygen record"` codifies it, and the reasoning in that comment is sound. But the summary says the opposite, and the summary is what reviewers and the v37/v38 backport approvers will read. Worth rewording to state the removal explicitly.

Then the real question: if a rotation is ever needed while this is deployed, what is the path? As written it appears to be "roll the clients back first". If that is the accepted answer it is worth writing down. If not, an escape hatch — honour a scheduled keygen when the record has grantees *and* an operator opts in — would keep the option without reintroducing the strand this PR is removing.

---

Neither blocks the fix. Nice catch on `waitForNewKeygen` being the thing that took every running signer down at once — that explains why operators who never touched their nodes were also reporting crashes.

zetaclient/tss/setup.go
kingpinXD: Fixed in `e8f89fbd2`. `resolveTSSPeers` now falls back to the keygen grantees when the participant list is empty, while still reporting finalized so an imported key is never regenerated. Confirmed the case is real — `x/observer/genesis.go:91` imports the TSS verbatim. Added a unit test for it, plus a guard in `Setup` so an empty whitelist fails with a clear message instead of surfacing as missing bootstrap peers from go-tss.

kingpinXD:
Thanks both — all three points were real. Pushed in `e8f89fbd2`.

**Rotation framing (@julianrubino #1).** You're right and the old wording was circular: nothing would have produced the new key the listener was supposed to react to. The description now leads with "scheduled keygen rotation is disabled while a finalized key exists", and both `setup.go` step 5 and the `Listen` comment say the same, including how to rotate on purpose.

**`GetTSS` fallback (@ws4charlie, @julianrubino #2).** Retried with backoff; only the not-found code counts as absent; anything else fails loudly instead of falling back to a record we've just declared unreliable.

**Genesis TSS without participants (@julianrubino #3).** Confirmed real — `x/observer/genesis.go:91` imports the TSS verbatim. Falls back to the keygen grantees while staying finalized so an imported key is never regenerated. Also added a guard so an empty whitelist fails with a clear message rather than as missing bootstrap peers.

**Nit #4 — `WithMetrics`.** Confirmed: it reads `keygen.GranteePubkeys` for blame labels, so during a blanked window those labels are empty until the record is repopulated. Metrics-only, no signing impact. Left alone to keep this diff to the outage.

**#5 — codecov.** The uncovered lines are the new block inside `Setup` itself, which has no unit test because it needs a real p2p server, key files and pre-params. The decision logic is factored into `resolveTSSPeers` and `tssNotFound`, both unit-tested and mutation-checked. Not an uncovered branch in either.

New tests: genesis-TSS-without-participants, and `tssNotFound` across not-found / transient / nil. Both mutation-checked — pointing `tssNotFound` at the wrong code fails two subtests, dropping the participant-list fallback fails the genesis one.

Verification on localnet is unchanged and still green: `start-keygen-reset-test` passes with the record blank at the end, and `start-drain-test` still moves funds.

zetaclient/tss/setup.go
kingpinXD: Correction — the reply above describes `e8f89fbd2`, which has since been replaced. Matching the gRPC code is gone as of `1795af8e0`.

Two reasons it was the wrong tool. zetacore answers an absent record with `InvalidArgument` rather than `NotFound`, so the match depended on that being wrong, and this ships to v37/v38 clients that talk to a zetacore of a different version — correcting the code upstream would have stopped fresh chains starting. `errors.Is` cannot help either, since the error crosses a gRPC boundary and only a code and a string survive.

It turned out nothing needed the distinction. A failed query falls back to the record, and the empty-whitelist guard is what keeps that safe: a blanked record has nothing to whitelist, so startup stops with a clear error and the supervisor retries. An intact record carries the same grantee set and a succeeded keygen makes the ceremony a noop, so a blip heals itself. `GetTSS` is retried 5s x 10 first.

zetaclient/tss/setup.go
kingpinXD: Correction — the reply above describes `e8f89fbd2`, which has since been replaced. Matching the gRPC code is gone as of `1795af8e0`.

Two reasons it was the wrong tool. zetacore answers an absent record with `InvalidArgument` rather than `NotFound`, so the match depended on that being wrong, and this ships to v37/v38 clients that talk to a zetacore of a different version — correcting the code upstream would have stopped fresh chains starting. `errors.Is` cannot help either, since the error crosses a gRPC boundary and only a code and a string survive.

It turned out nothing needed the distinction. A failed query falls back to the record, and the empty-whitelist guard is what keeps that safe: a blanked record has nothing to whitelist, so startup stops with a clear error and the supervisor retries. An intact record carries the same grantee set and a succeeded keygen makes the ceremony a noop, so a blip heals itself. `GetTSS` is retried 5s x 10 first.

zetaclient/tss/setup.go
kingpinXD: Still accurate: `resolveTSSPeers` falls back to the keygen grantees when the participant list is empty, and keeps reporting finalized so an imported key is never regenerated. The empty-whitelist guard in `Setup` is also still there. Only the `GetTSS` error-classification part of the earlier replies was withdrawn — see the other two threads.

kingpinXD:
Thanks @morde08 — both points were right, and both are addressed as of `1795af8e0`. Point 1 landed differently from your suggestion though, so it is worth saying why.

**1. `GetTSS` failing open.** Fixed, but without the not-found classification.

I did build that first, matching the status code. It was the wrong tool: zetacore answers an absent record with `InvalidArgument` rather than `NotFound`, so the match depended on that being wrong, and this ships to v37/v38 clients talking to a zetacore of a different version — correcting the code upstream would have stopped fresh chains starting. `errors.Is` is no help either, since the error crosses a gRPC boundary and only a code and a string survive.

Then it turned out nothing needed the distinction:

- **Blanked record** — nothing to whitelist, so the guard stops startup with a clear message and the supervisor retries. Fails closed and loud, which is what you asked for.
- **Intact record** — the grantees are the same set, and a keygen that already succeeded makes the ceremony a noop, so a blip heals itself instead of taking the node down.
- **Fresh chain** — the record is the correct source and the ceremony is the right next step.

`GetTSS` is retried first, constant 5s x 10, sized to outlast an RPC restart rather than the sub-second exponential I had before. Your coordinated-restart-against-a-busy-zetacore scenario is exactly what moved that number.

Your minor point is fixed as you described: `resolveTSSPeers` falls back to the keygen grantees when the participant list is empty, still reporting finalized so an imported key is never regenerated. It is real — `x/observer/genesis.go:91` imports the TSS verbatim.

**2. Rotation.** Description corrected before your review landed; it now leads with rotation being disabled while a finalized key exists, and both code comments say the same.

On your real question — the honest answer today is "roll the clients back first", and I would rather write that down than pretend otherwise. Your escape hatch is the better answer, but it adds an opt-in path to an outage hotfix going to two release branches, so I would rather land this and do it deliberately as a follow-up than smuggle it in here. Happy to be overruled if you would rather it went in now.

**Verification.** Localnet keygen-reset e2e and the drain regression both pass; the record is still blank at the end and both signers stay clean, and a manual restart into the blanked state comes up on the existing key. Re-running both now against the current head, since the last few commits changed the startup path.

kingpinXD:
@morde08 on the rotation path: we are not planning a TSS keygen at any point in the foreseeable future, so I am leaving the escape hatch out rather than adding an opt in path to a hotfix going to two release branches.

If that ever changes, the route is to roll the clients back first, and the step 5 comment in `setup.go` spells that out so nobody has to rediscover it under pressure.

zetaclient/maintenance/tss_listener.go
```diff
@@ -32,6 +32,15 @@ func NewTSSListener(client ZetacoreClient, logger zerolog.Logger) *TSSListener {
...
 // Listen listens for any maintenance regarding TSS and calls action specified. Works in the background.
+//
+// Both watchers key off the TSS itself: the address changing, or a new key landing in history.
+// The keygen record is deliberately not watched. It is reset to "pending at block MaxInt64" on
+// any observer set change, and restarting on that reset is what turns a routine validator
+// unbonding into an outage — every signer shuts down at once and none of them can start again.
+//
+// Note this also removes the only trigger that restarted zetaclient for a scheduled keygen, so
+// while a finalized key exists a rotation ceremony will not start. That is deliberate; see the
+// step 5 comment in zetaclient/tss/setup.go.
```
ws4charlie, line 45: With the keygen watcher gone the `start-tss-migration-*` suites will hang rather than fail — `waitKeygenHeight` (cmd/zetae2e/local/local.go:914) loops until the record flips to KeyGenSuccess and nothing runs the ceremony any more, and they are label-gated so this PR's green run never exercised them. Could we skip/disable them here and fix the comment at cmd/zetae2e/local/tss_migration.go:109 that still says the TSSListener manages those restarts, so the v37/v38 backports do not ship with a silently dead migration path?

zetaclient/tss/setup.go
ws4charlie: @kingpinXD this reads as resolved to me — `currentTSS = TSS{}` on exhaustion still swallows the error, but the guard at setup.go:127 now turns that into a hard "no TSS peers to whitelist" rather than booting with an empty whitelist, which was the part that actually worried me. Appreciated the self-correction when the gRPC-code approach came back out.

zetaclient/maintenance/tss_listener.go
kingpinXD: Good catch, and confirmed — thanks. Both suites go through `waitKeygenHeight`, which loops with no bound until the record flips to `KeyGenSuccess`, and nothing produces that flip any more. One refinement: in CI they do not hang, they burn the 40 minute `timeout-minutes` and fail. Locally they hang. Either way they run on the nightly `schedule` event, not just on the label, so this would have shown up as a permanently red nightly on main and on both release branches.

Disabled both matrix entries in `711644040` with a comment saying why and what to restore when rotation comes back, and fixed the stale comment at `tss_migration.go:109`. Also added a doc comment on `triggerTSSMigration` itself, since someone running the make target locally gets no CI timeout to save them.

ws4charlie: @kingpinXD perfect, that is exactly the shape I was hoping for — both matrix rows off with the restore condition written down, and the comment now says what the TSSListener actually still does versus what it stopped doing. The new `triggerTSSMigration` doc naming the forever-block is a nice touch for whoever re-enables rotation.

ws4charlie (approved):
lgtm — fail-closed throughout (a blanked keygen record now ends in the "no TSS peers to whitelist" error rather than an empty whitelist), and the migration suites are off with the restore condition documented. Only the freshly-queued CI is outstanding; `publish-release-legacy` is red on every branch in the repo, not this PR.

zetaclient/maintenance/tss_listener.go
ws4charlie: Fair correction on my wording — in CI they burn `timeout-minutes` and fail, only a local run actually hangs. The nightly `schedule` detail makes it worse than I framed it, so thanks for that; the disable plus the corrected comment covers what I was worried about.

ws4charlie (approved):
LGTM. The `GetTSS` retry plus the empty-whitelist guard covers my earlier comment nicely — worst case now is a loud startup failure instead of a signer hanging on a block that never comes, and the genesis-TSS-with-no-participants case is a good catch. Rotation staying off is clearly documented in all three places and the migration suites are disabled with a note, so nothing is silently broken.

Makefile
```diff
@@ -331,6 +331,13 @@ start-drain-test: e2e-images
 	export ZETACLIENT_DRAIN_WINDOW="500" && \
 	cd contrib/localnet/ && $(DOCKER_COMPOSE) --profile dry up -d
 
+# start-keygen-reset-test runs ONLY the keygen reset e2e test in isolation. It leaves the
+# keygen record erased, so it must not share a localnet with tests that expect a populated one.
+start-keygen-reset-test: e2e-images
```
morde08, line 336: **Medium — this target is never executed by any workflow.**

`start-keygen-reset-test` lands here, in `cmd/zetae2e/local/local.go`, and in `e2e/e2etests/e2etests.go`, but there is no matrix row in `.github/workflows/e2e.yml` and no `matrix-conditionals` output for it. `git grep -l keygen-reset-test` returns only this file — compare `start-drain-test`, which appears in both the Makefile and the workflow matrix.

Net effect of this PR on automated coverage of the TSS-startup path: two suites turned off (`start-tss-migration-*`, correctly), one added that nothing ever runs. The test that proves the mainnet fix only executes if someone types the make target by hand — which is precisely the thing that will not happen six months from now when someone refactors `resolveWhitelist`.

Suggest a matrix row plus a `KEYGEN_RESET_TESTS` conditional set `true` on `schedule` and mapped under `workflow_dispatch`, mirroring `DRAIN_TESTS`.

changelog.md
```diff
@@ -14,6 +14,7 @@
 
 ### Fixes
 
+* [4618](https://github.com/zeta-chain/node/pull/4618) - keep zetaclient signing when zetacore resets the keygen record on an observer set change, instead of shutting every signer down and refusing to restart.
```
morde08: **Medium — the rotation consequence isn't here.**

It's documented in three code comments and in the PR body, but this line is the operator-facing surface and it mentions only the half that restores service. This ships to `release/zetaclient/v37` and `v38`; an operator who broadcasts `MsgUpdateKeygen` on those branches gets silence and no ceremony, with nothing in the changelog to explain why.

Worth a second sentence here, e.g. "note: while a finalized TSS exists, a keygen scheduled via `MsgUpdateKeygen` will not run — rotating requires removing the current TSS first."

Separately: a tracking issue linked from the setup.go step-5 comment would give "re-enable rotation" an owner. Right now the only record that rotation is off lives in code comments, and the two e2e suites that would have failed loudly about it are disabled in this same PR.

zetaclient/tss/setup.go
```diff
@@ -185,6 +198,109 @@ func Setup(ctx context.Context, p SetupProps, logger zerolog.Logger) (*Service,
...
+// constant backoff is deliberate: sub-second retries give up inside an RPC restart, and the
+// only node that pays the full wait is one on a chain with no key, which then waits for the
+// keygen block anyway.
+func resolveWhitelist(
+	ctx context.Context,
+	client tssFetcher,
+	tssKeygen observertypes.Keygen,
+	setupLogger zerolog.Logger,
+) (observertypes.TSS, []peer.ID, bool, error) {
+	currentTSS, tssErr := retry.DoTypedWithBackoffAndRetry(
```
morde08, line 239: **Medium — the hardening covers one of the three RPCs on this path.**

`GetTSS` now gets 10×5s, but `GetKeyGen` at line 83 and `GetTSSHistory` at line 140 are still one-shot fatals. The scenario this retry exists for — every signer restarting at once with zetacore RPC as the contended resource — hits all three calls equally, and `GetKeyGen` fails *before* the hardening can apply.

Worth noting that once a TSS is finalized, the keygen record is only needed for the genesis-import fallback in `resolveTSSPeers`, so `GetKeyGen` failing arguably shouldn't be fatal at all — it could warn and pass a zero `Keygen`, letting the empty-whitelist guard below decide.

At minimum, wrapping both in the same `retry.DoTypedWithBackoffAndRetry(..., retry.DefaultConstantBackoff())` is a two-line change and makes the startup path uniformly tolerant of the blip it was designed around.

zetaclient/tss/setup.go
```diff
@@ -185,6 +198,109 @@ func Setup(ctx context.Context, p SetupProps, logger zerolog.Logger) (*Service,
...
+	peerSource, keyAlreadyFinalized := resolveTSSPeers(tssKeygen, currentTSS)
+	if keyAlreadyFinalized {
+		setupLogger.Info().
+			Str("tss_pubkey", currentTSS.TssPubkey).
+			Int("grantees_in_keygen_record", len(tssKeygen.GranteePubkeys)).
+			Int("participants_in_tss", len(currentTSS.TssParticipantList)).
+			Msg("TSS key already finalized; whitelisting its participants")
+	}
+
+	// Starting with nothing to whitelist is the failure this function guards against, so say
```
morde08, line 258: **Minor — the guard is right, but the stated reason understates it.**

"instead of letting go-tss report it as missing bootstrap peers" — `NewServer` only errors on `len(whitelistPeers) == 0 && len(bootstrapPeers) == 0`. With bootstrap peers present (the normal case, and the mainnet case) an empty whitelist doesn't produce *any* error: the server starts cleanly and then silently refuses every peer at keysign. That's a worse failure than the comment implies and a better argument for this guard, so it's worth saying.

While here: `WithMetrics` (`zetaclient/tss/service.go:109`) still seeds `NodeBlamePerPubKey` label values from `keygen.GranteePubkeys`, so on a blanked record it pre-registers no labels. Purely cosmetic — labels appear on first blame anyway — but it's the last keygen-record read left in the signing path, and consistency with the premise of this PR would argue for `TssParticipantList` or a comment saying why it doesn't matter.

zetaclient/maintenance/tss_listener_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+const (
+	tssPubkeyOld = "zetapub1addwnpepqtadxdyt037h86z60nl98t6zk56mw5zpnm79tsmvspln3hgt5phdc79kvfc"
+	tssPubkeyNew = "zetapub1addwnpepqglunjrgl3qg08duxq9pf28jmvrer3crwnnfzp6m0u0yh9jk9mnn5p76utc"
+)
+
+// waitForListenerTick gives the listener room for at least one tick so a shutdown it was
+// going to signal has actually had the chance to fire.
+func waitForListenerTick() {
+	time.Sleep(2 * tssListenerTicker)
```
morde08: **Medium — this adds 31s of wall clock to the `zetaclient/maintenance` package.**

Measured on this branch:

```
--- PASS: TestTSSListener (20.00s)
    --- PASS: .../blanked_keygen_record_does_not_trigger_a_shutdown (10.00s)
    --- PASS: .../TSS_address_change_still_triggers_a_shutdown (5.00s)
    --- PASS: .../new_key_in_TSS_history_still_triggers_a_shutdown (5.00s)
--- PASS: TestTSSListenerIgnoresKeygen (10.00s)
ok  github.com/zeta-chain/node/zetaclient/maintenance  31.011s
```

Roughly 30s of that is `time.Sleep` against the hardcoded `tssListenerTicker = 5 * time.Second`. Making the ticker a package-level `var` and shrinking it to ~10ms in a test setup gets the identical assertions in under a second. Given the tests themselves are good and should stay, this is worth doing so nobody is tempted to delete them later for being slow.

zetaclient/maintenance/tss_listener_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	client.Mock.AssertNotCalled(t, "GetKeyGen", ctx)
+
+	// Guard against the assertion above passing for the wrong reason: the record a caller
+	// would have read is the blanked one, and it must remain irrelevant.
+	blanked := observertypes.Keygen{
+		Status:      observertypes.KeygenStatus_PendingKeygen,
+		BlockNumber: math.MaxInt64,
+	}
+	if len(blanked.GranteePubkeys) != 0 {
```
morde08: **Minor — this check can't fail.**

`blanked` is constructed four lines above without setting `GranteePubkeys`, so `len(...) != 0` is false by construction regardless of what the listener does. It doesn't guard the `AssertNotCalled` above from passing for the wrong reason, contrary to the comment.

The `AssertNotCalled` is already doing the real work — I confirmed it fails when a keygen watcher is reintroduced. If you want a genuine belt-and-braces here, registering `client.Mock.On("GetKeyGen", ctx).Return(blanked, nil)` and then asserting it was never called would at least exercise the mock path; otherwise this block is safe to drop.

e2e/e2etests/test_keygen_reset_signing.go
```diff
@@ -0,0 +1,114 @@
...
+	// The record is now in the state that took mainnet down.
+	keygenAfter, err := r.ObserverClient.Keygen(r.Ctx, &observertypes.QueryGetKeygenRequest{})
+	require.NoError(r, err)
+	require.NotNil(r, keygenAfter.Keygen)
+	require.Empty(r, keygenAfter.Keygen.GranteePubkeys, "grantee list should have been erased")
+	require.Equal(r, observertypes.KeygenStatus_PendingKeygen, keygenAfter.Keygen.Status)
+	require.Equal(r, int64(math.MaxInt64), keygenAfter.Keygen.BlockNumber)
+
+	// The same handler pauses inbound, so restore it before expecting a deposit to flow.
+	reEnableInbound(r)
```
morde08, line 94: **Observation — this line is the other half of the incident, and it stays unfixed.**

The same BeginBlocker calls `DisableInboundOnly` (`x/observer/keeper/crosschain_flags.go:52-58`), which sets `IsInboundEnabled = false` chain-wide until an admin `MsgUpdateCrosschainFlags`. This test has to call `reEnableInbound(r)` to get past it, which makes the dependency nicely explicit.

Out of scope for a zetaclient hotfix — it's a zetacore change and shouldn't be bundled here. But after this merges, a routine validator unbonding still halts every inbound deposit until a human notices and files a governance tx. Worth its own issue so the incident isn't treated as closed by this PR alone.

morde08 (commented):
Second pass, at `4e3fbd935` (the `resolveWhitelist` extraction). I re-derived the claims rather than reading them, and the core fix holds up.

**Verified independently**

- `TssParticipantList` is written once at TSS creation and is untouched by `x/observer/abci.go:37`'s `SetKeygen(Keygen{BlockNumber: MaxInt64})`, so it survives the blanking. The struct literal does zero `GranteePubkeys` exactly as described.
- The e2e stand-in claim checks out: `MsgAddObserver` with `AddNodeAccountOnly=true` runs `DisableInboundOnly` + the same `SetKeygen` (`x/observer/keeper/msg_server_add_observer.go:37,51`). The comment's insistence that this is a convenience stand-in and *not* what happened on mainnet is worth keeping exactly as written.
- The "ceremony is a noop" reasoning is sound: `zetaclient/tss/keygen.go:104` returns `(false, nil)` on `KeyGenSuccess` and line 89 returns `zc.GetTSS(ctx)`, so the GetTSS-failed-but-record-intact fallback really does self-heal.
- **I ran the mutation check myself.** Re-adding a keygen watcher to `Listen` fails `TestTSSListenerIgnoresKeygen` *and* all three `TestTSSListener` subtests on the unexpected `GetKeyGen`. The guard is real, not decorative. All 10 new subtests pass locally.

Nothing below blocks the hotfix. The CI wiring gap is the one I'd want closed before this goes to v37/v38, since it is the only thing that would catch a regression of this exact bug.

Makefile
kingpinXD: Fixed in `3f95b978a`. You were right that `git grep` told the whole story — `keygen-reset` was in the Makefile, `local.go` and `e2etests.go`, and nowhere in the workflow. Wired `KEYGEN_RESET_TESTS` through all five sites mirroring `DRAIN_TESTS`: the outputs block, the label check, `true` on `schedule`, the `workflow_dispatch` mapping to `keygen-reset-test`, and a matrix row with a 40 minute timeout.

zetaclient/tss/setup.go
kingpinXD: Fixed in `3f95b978a`, and it was worse than two calls. There is a third: `WithMetrics` (`zetaclient/tss/service.go:109`) makes its own `GetKeyGen` and returns an error on failure, which propagates out of `NewService` and kills `Setup`. Easy to miss because it sits in an option rather than in `Setup`, and it does not show up when you read the function top to bottom. All three now use `retry.DoTypedWithBackoffAndRetry(..., retry.DefaultConstantBackoff())`.

Your point about ordering is the one that made this worth doing: `GetKeyGen` runs first, so a one-shot fatal there meant the hardening downstream could never apply. That is in the comment now.

I left all three fatal rather than taking your suggestion of letting `GetKeyGen` warn and pass a zero `Keygen`. It is a bigger behaviour change than a hotfix heading to two release branches wants, and with the retry in place the failure it protects against is much narrower.

zetaclient/maintenance/tss_listener_test.go
kingpinXD: Fixed in `3f95b978a`. `tssListenerTicker` is a `var` now, shrunk to 10ms per test via a helper with `t.Cleanup` to restore it. `TestTSSListener` plus `TestTSSListenerIgnoresKeygen` went from ~30s to 0.42s, and the package from 36.8s to 6.7s — the rest is the pre-existing `TestShutdownListener` suite, not these.

I re-ran your mutation afterwards, since faster timing is exactly how a test goes quietly vacuous. Re-adding a keygen watcher still fails `TestTSSListenerIgnoresKeygen` and two `TestTSSListener` subtests.

zetaclient/maintenance/tss_listener_test.go
kingpinXD: You are right, and this is the exact failure mode I should have caught. `blanked` is constructed four lines above without `GranteePubkeys`, so the check was false by construction and guarded nothing — a comment claiming to protect an assertion while doing no work is worse than no comment.

Dropped it in `3f95b978a` rather than taking the `client.Mock.On("GetKeyGen", ...)` route: `mocks.NewZetacoreClient(t)` registers `AssertExpectations` on cleanup, so a registered-but-uncalled expectation would fail the test for the opposite reason. `AssertNotCalled` was always the real guard.

e2e/e2etests/test_keygen_reset_signing.go
kingpinXD: Filed as https://github.com/zeta-chain/node/issues/4622, and I agree it stays out of this PR. The point that after this merges a routine validator unbonding still halts every inbound until someone files a governance tx is the part worth not losing, so the issue leads with it.

Also noted there that the `reEnableInbound(r)` call in the test is the evidence — it exists because the test cannot get past the pause otherwise.

changelog.md
kingpinXD: Both done. Changelog now carries the second sentence in `3f95b978a`: while a finalized TSS exists, a keygen scheduled via `MsgUpdateKeygen` will not run, and rotating requires removing the current TSS first.

Tracking issue is https://github.com/zeta-chain/node/issues/4623, linked from the step 5 comment in `setup.go` so it has an owner. It also records that re-enabling rotation means re-enabling the two `start-tss-migration-*` matrix rows this PR turns off, since you are right that those were the loud failure and they are gone in the same change.

zetaclient/tss/setup.go
kingpinXD: Checking this one I think the conclusion goes the other way, so flagging rather than changing it.

You are right about the `NewServer` switch — it is `len(whitelistPeers) == 0 && len(bootstrapPeers) == 0`, an AND. But that is only a pre-check. The real gate is downstream in go-tss: `resolveBootstrapPeers` (`go-tss@v0.6.3/tss/tss.go:243-247`) opens with

```go
if len(net.WhitelistedPeers) == 0 {
    return nil, errors.New("whitelisted peers missing")
}
```

with no reference to bootstrap peers. That is the mainnet crash string — the full message was `unable to create TSS server: fail to resolve bootstrap peers: whitelisted peers missing`, which is this error wrapped by its caller. So an empty whitelist does fail hard with bootstrap peers present, and the "starts cleanly, then silently refuses every peer at keysign" case does not arise. Leaving the comment as is.

On `WithMetrics`: the label seeding is cosmetic as you say, but the `GetKeyGen` call it makes is not — it returns an error that kills `Setup`. Retried in `3f95b978a` along with the other two. I have left the labels reading `GranteePubkeys`, since on a blanked record it pre-registers nothing and they appear on first blame anyway.

julianrubino (commented):
**Update after independent re-verification** (two reviewers re-derived the claims against the current head `b000039`, which has advanced past my earlier review). Two corrections + one addition:

### Correction 1 — my earlier `codecov/patch` note is stale
It's **green** on the current head, not red. The coverage gap I flagged was closed by the commits added since. Disregard that point.

### Correction 2 — the `GetTSS` fallback is worse than I rated it (elevating: minor → **Medium-High**, worth closing before merge)
My inline on `setup.go` treated the `GetTSS`-error fallback as defense-in-depth. On closer look it re-opens the *exact* outage this PR fixes, under the same storm:
- `GetTSS` here is a **single-shot call, no retry** — contrast the listener (`tss_listener.go`) which wraps the identical call in `retry.DoTypedWithBackoffAndRetry`. The grpc handler errors on any transport failure, not just genesis "not found".
- On a healthy node whose keygen record is **currently blanked** (the incident state), a transient `GetTSS` hiccup at startup → `currentTSS = TSS{}` → `resolveTSSPeers` returns `finalized=false` → whitelist = empty grantees **and** `KeygenCeremony` runs against the blanked record and **blocks forever** at the `zetaHeight < MaxInt64` wait.
- And it **hangs rather than crash-loops**: `NewServer` only errors when whitelist *and* bootstrap peers are both empty; with bootstrap peers configured (mainnet) it passes through to the stuck ceremony, so a k8s restart does **not** recover it.
- Trigger = mass restart → zetacore RPC pressure → transient `GetTSS` failure while the record is blanked — precisely the storm this PR targets.

Fix: wrap this `GetTSS` in the same retry the listener uses, and/or distinguish a genuine not-found (genesis → ceremony) from a transient error (→ return it, let the supervisor retry) instead of silently degrading. The genesis empty-participant-list guard I flagged inline would double as the fast-fail here (`if finalized && len(participantList)==0 → error loudly`).

### Addition — no test restarts a client against a blanked record end-to-end
The e2e reproduces the *state* and proves *keep-signing*, but never restarts a zetaclient, so the actual mainnet failure mechanism — cold-start whitelist recovery from `TssParticipantList` against a blanked record — rests entirely on the `resolveWhitelist` unit test. Solid coverage, but if you want the e2e to truly "reproduce the incident," it needs a client restart against the blanked record.

Everything else in my original review stands and was confirmed: the whitelist-source change is a **tightening** not a weakening (`TssParticipantList` = the actual share holders; admission gate ≠ signing threshold), rotation-keygen is genuinely disabled while a finalized key exists (the "rotation unaffected" wording is circular — nothing produces the new key to land in history), and the tests are mutation-resistant. Net: ship-worthy, but I'd close the `GetTSS`-retry gap first since it re-opens the same outage.

kingpinXD:
@julianrubino thanks for re-running this. Taking the three points in turn.

**Correction 2 — the premise is stale.** `GetTSS` has been retried since `3f95b978a`, which landed before the head you reviewed. On `b000039` it reads:

```go
currentTSS, tssErr := retry.DoTypedWithBackoffAndRetry(
    func() (observertypes.TSS, error) { return client.GetTSS(ctx) },
    retry.DefaultConstantBackoff(),
)
```

Same helper and same backoff as the listener — 11 attempts over roughly 50s. `GetKeyGen` and `GetTSSHistory` got the same treatment in that commit, plus a third `GetKeyGen` hiding inside `WithMetrics` that also killed startup.

**And the scenario terminates earlier than you traced it.** Blanked record plus a `GetTSS` failure gives `currentTSS = TSS{}`, so `resolveTSSPeers` returns the empty grantee list with `finalized=false` — and then hits this, before `NewServer` and before any ceremony:

```go
if len(peerSource) == 0 {
    return observertypes.TSS{}, nil, false, errors.New("no TSS peers to whitelist: ...")
}
```

Startup stops with that error. It never reaches `KeygenCeremony`, so the forever-block at `zetaHeight < MaxInt64` is not reachable by this route. `TestResolveWhitelist/query_failure_on_a_blanked_record_stops_startup` pins exactly that, and mutating the guard fails exactly that test.

**On "hangs rather than crash-loops":** the `NewServer` condition you quote is real but it is only a pre-check. The binding gate is downstream in go-tss — `resolveBootstrapPeers` (`go-tss@v0.6.3/tss/tss.go:243-247`) opens with `if len(net.WhitelistedPeers) == 0 { return nil, errors.New("whitelisted peers missing") }`, with no reference to bootstrap peers. That is the mainnet crash string. @morde08 raised the same point and this was the answer there too.

I also built and then removed the not-found-vs-transient classification you suggest, in an earlier round. It could not be done reliably: the status code crosses an RPC boundary, so `errors.Is` does not survive it. The guard above is what makes the unconditional fallback safe instead.

**Correction 1 — agreed,** and thanks. Codecov is green at 77.55%.

**Addition — you are right, and it is not fixable in the test.** The e2e proves keep-signing but never restarts a client, so cold-start recovery rests on the unit tests. I did verify it by hand during development: a `docker restart` against the blanked state comes up with `TSS key already finalized; whitelisting its participants grantees_in_keygen_record=0`, then `TSS service created`.

Automating it is blocked on infrastructure. The e2e runs inside the `orchestrator` container, which has no docker socket — only `promtail` gets one, under the `monitoring` profile — and there is no docker client dependency in the repo. So a test cannot restart a sibling container. Filed as https://github.com/zeta-chain/node/issues/4625, since it blocks any future test of zetaclient startup, not just this one.

julianrubino (approved):
Approving. The core reasoning holds up: the keygen record describes a key yet to be generated, so it cannot be authoritative about one generated long ago, and `resolveTSSPeers` puts that in one readable place. I verified the two mutation claims rather than taking them on trust — reverting the `currentTSS.TssPubkey == ""` guard fails 5 subtests across `TestResolveTSSPeers` and `TestResolveWhitelist`, and reintroducing any read of the keygen record from `Listen` fails all four listener cases. The rotation consequence is now stated in three places, tracked in #4623, and carried into the changelog as an operator note, which is what I asked for last round.

Also checked that `tssInfo` is only ever read as `.TssPubkey` downstream (setup.go:147, :156), so `tssInfo := currentTSS` is genuinely the same value `KeygenCeremony` returns via `GetTSS` on its success path — including the genesis-import case where the participant list is empty.

Nothing blocking. Four notes, one of which I think belongs in #4623 before someone picks it up.

**setup.go:307 — the whitelist source will bite the rotation work.** `TssParticipantList` is frozen at generation, while `MsgUpdateKeygen` rebuilds `GranteePubkeys` from *all* node accounts (msg_server_update_keygen.go:37-44). So an observer added after the last keygen is a grantee but not a participant, and existing signers no longer whitelist it. Harmless today — it holds no key share, so it could never keysign anyway, and refusing it at p2p is arguably tighter than before. But a rotation ceremony has to include exactly those new members, so re-enabling rotation means the whitelist has to come from the grantees for the ceremony's duration, not just restoring a trigger. #4623 talks about the trigger and the e2e rows; it doesn't mention this. Worth adding there, since whitelisting happens once in `Setup` and a new member that isn't whitelisted fails as "missing bootstrap peers" rather than as anything that points here.

**e2e — the half that made it unrecoverable isn't covered end to end.** `TestKeygenResetSigning` blanks the record and proves signing continues, which is the "signers stayed up" half. Nothing in the suite restarts a zetaclient, so `resolveWhitelist` never runs full-stack against a blanked record — and "none could start again" is the half that turned an incident into an outage. The unit tests do cover it and they do bite (I mutated them), so this is a coverage note, not a request. If a container restart is cheap to add to that make target, it would close the loop.

**setup.go:217 — comment slightly overstates the fallback.** "a keygen that already succeeded makes the ceremony a noop" is true for `KeyGenSuccess` (keygen.go:104), but a `GetTSS` failure against a record left `PendingKeygen` at a future block sends `Setup` into the ceremony wait instead. Moot while rotation is off and only reachable after 10x5s of retries, but the code is narrower than the sentence.

**Retry ceilings, informational.** `GetKeyGen` already retries internally (clients_observer.go:102, ~1.1s), so the new outer `DefaultConstantBackoff` nests to ~65s per call and it now runs twice on startup — `resolveWhitelist` plus `WithMetrics` (service.go:112). `GetTSS` and `GetTSSHistory` have no inner retry, so they cap at ~50s each. Worst-case startup is therefore a few minutes before a hard fail rather than the ~50s the comments imply. That still beats the one-shot fatal it replaces; just worth knowing the real number.

**tss_listener.go:18 — nit.** `tssListenerTicker` is a package var that `useFastTicker` writes while `Listen`'s workers outlive the test. Safe only because the workers read it once at ticker construction and unit tests don't run under `-race`. If a race job ever lands, this is a flake waiting to happen; an injectable interval on the struct would remove it.

The `MsgAddObserver` stand-in being labelled that emphatically in the test doc is the right call — that framing is what stops the next person reading the repro as the cause.

morde08 (approved):

PR #4624, opened 2026-08-13
feat(zetatool): add rehearsal value caps to the TSS drain payload

skosito:
## Summary

Follow-up to #4612. The drain moves the **full** native balance of every chain in one shot. Before committing that on mainnet, we want to rehearse the whole path — payload → signature verify → TSS ceremony → broadcast — with a small value.

`--only-chains` already narrows *which* chains move. This adds two flags to `zetatool drain-payload` that bound *how much* moves:

- **`--evm-max-amount <wei>`** — caps the per-chain transfer. Only the transfer is lowered; gas price, gas limit and fee are computed exactly as in a real drain, so the pinned tx is shaped identically and the remainder stays at the TSS address.
- **`--btc-max-sats <sats>`** — caps the total swept, by selecting a subset of UTXOs.

## Why BTC can't just take a smaller number

A UTXO is indivisible and the drain sweep has **no change output** — the poller's `validateBTCFee` requires `output + fee == sum(inputs)` precisely so a malformed payload can't burn the remainder to miners. Lowering an output would either break that invariant or need a second output back to the TSS address, which is a new redirect surface in the poller. Neither is worth it for a rehearsal.

So the cap selects UTXOs instead:
- **largest-first** among those that fit, and — critically — only if the resulting group still satisfies the poller's own bound `fee <= total/MaxBTCFeeFraction`, priced at the input count that input would produce. Fitting under the cap is not sufficient: a UTXO too big for the cap is skipped and selection continues down the list, so a value-only test lets a small cap slide past every large UTXO and fill the group with dust, whose fee then dwarfs its value — the group is dropped as uneconomical and the rehearsal sweeps no BTC at all. The same fixed ~68 vB per input kills viable sweeps from the other side: one 300-sat input added to a viable 90k-sat sweep at 50 sat/vB pushes the fee past the bound. Inputs that do cover their own marginal share are still taken, so this is not a blanket dust exclusion;
- limited to **one tx worth of inputs** so a capped run emits exactly one sweep;
- ordering stays amount → TxID → Vout, preserving the cross-node determinism the keysign ceremony depends on.

## Scope

Both caps are **generator-side only**. The payload schema, the zetaclient poller and the receiver anchors are untouched — a rehearsal exercises the same client binary and the same code path as the real drain. (The poller never validated the EVM amount against the live balance; it validates receiver, gas-price ceiling and nonce, all unchanged.)

A capped payload is still `final` — clients only ever sign finals — so the generator prints a loud `WARN REHEARSAL PAYLOAD` banner to stderr. The caps are otherwise visible only as smaller numbers inside the JSON, which is easy to miss when reading a payload under pressure.

Also included: `GenerateEVMTx` now refuses to pin a **zero-amount** tx, which would pay gas to move nothing. Previously possible when a balance was exactly its fee.

## Operator flow

```
# 1. rehearsal, small value, one trigger height
zetatool drain-payload mainnet --trigger-height H1 --serve \
  --evm-max-amount 1000000000000000 --btc-max-sats 100000

# 2. confirm funds landed at the receivers, then the real drain at a higher height
zetatool drain-payload mainnet --trigger-height H2 --serve
```

**Wait for the rehearsal txs to confirm before the real payload freezes.** This is a correctness gate, not tidiness: the generator pins the *confirmed* nonce (`NonceAt`) while the poller compares it against the *pending* nonce and treats "pinned nonce already consumed" as a hard stop — so a still-unmined rehearsal tx at freeze time silently costs that chain the whole firing window, recoverable only by a third republish. The rehearsal is what creates this condition; it deliberately breaks the quiescence the drain otherwise assumes. The generator now warns per chain (`NONCE NOT QUIESCED`) whenever pending and confirmed differ.

Otherwise the second run needs no cleanup: the poller re-arms on any payload newer than `lastFiredHeight`, and the generator re-reads live balances, nonces and UTXOs each tick.

### What a small `--btc-max-sats` can and cannot do

The sweep spends whole UTXOs, so a viable rehearsal needs one UTXO **both** above `MinViableSweepSats` (≈`10 × feeRate × 171` sats — enough to out-earn its own fee) **and** under the cap. On holdings shaped like mainnet's — a handful of large UTXOs and a long dust tail — there may be no cap in between, in which case BTC can only be drained uncapped. The `--btc-max-sats 100000` above is illustrative, not a working value for any particular wallet.

A capped run that sweeps no BTC now reports `REHEARSAL SWEPT NO BTC` with the minimum viable cap and the smallest UTXO that would work, so an empty BTC section can't pass for a rehearsal that covered BTC.

## Tests

- EVM cap: nil/zero/below/equal/above the computed amount; gas price, gas limit and nonce unaffected.
- EVM zero-amount refusal.
- BTC cap: subset selection within the cap, prefers economical inputs over dust, one-tx bound, cap below the smallest UTXO, determinism under shuffled input, caller slice not mutated.
- `parseEVMMaxAmount` flag parsing, including >int64 values and rejected forms.

`go build -tags drain ./...`, `go vet` and the `pkg/drain` + `cmd/zetatool` suites pass.

## Notes

- Draft for review — changelog entry to follow with this PR number.
- Targets `main` like #4612; backport to `develop` after.

<h3>Greptile Summary</h3>

The PR adds generator-side EVM and BTC value caps so operators can rehearse the emergency TSS drain before performing the full sweep.
- Caps each EVM transfer while preserving its nonce, gas price, gas limit, and fee calculation.
- Selects a deterministic, economical, single-transaction BTC UTXO subset within the requested cap.
- Warns operators about rehearsal payloads, empty BTC rehearsals, and non-quiesced EVM nonces.
- Rejects zero-value EVM transfers and oversized EVM cap inputs.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| cmd/zetatool/cli/drain.go | Adds cap parsing and propagation, rehearsal diagnostics, BTC viability reporting, and EVM nonce-quiescence warnings; the previously reported oversized-cap panic is fixed. |
| pkg/drain/generate.go | Applies EVM transfer caps, rejects zero-value transfers, and implements deterministic economical BTC UTXO selection within the cap. |
| cmd/zetatool/clients/evm.go | Adds pending-nonce lookup used to warn when a pinned confirmed nonce is not safe for the drain payload. |
| cmd/zetatool/cli/drain_cap_test.go | Covers EVM cap parsing boundaries and operator-facing rehearsal and nonce diagnostics. |
| pkg/drain/generate_test.go | Exercises EVM cap behavior and BTC cap determinism, economics, transaction limits, and caller-slice preservation. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Read live balances, nonces, and UTXOs] --> B{Rehearsal cap configured?}
    B -->|EVM cap| C[Limit transfer amount]
    B -->|BTC cap| D[Select deterministic economical UTXO subset]
    B -->|No cap| E[Build full drain]
    C --> F[Preserve gas and nonce parameters]
    D --> G[Build one capped BTC sweep]
    F --> H[Build and sign payload]
    G --> H
    E --> H
    H --> I[Publish draft or final payload]
```

<sub>Reviews (5): Last reviewed commit: ["test(drain): fix a wrong claim in the ca..."](https://github.com/zeta-chain/node/commit/c0e0a3e7faabb0ed66898b032d43a60e735a2c73) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53177344)</sub>


skosito:
@greptile

skosito:
bugbot run

skosito:
@greptile

pkg/drain/generate.go
```diff
@@ -142,6 +165,47 @@ func GenerateBTCTxs(in BTCInput) ([]draintx.BTCTx, error) {
...
+// MaxBTCFeeFraction, and the group would be dropped as uneconomical. The subset is limited to
+// one tx worth of inputs so a capped run emits exactly one sweep.
+func selectCappedUTXOs(utxos []UTXO, maxSats int64) []UTXO {
+	selected := make([]UTXO, 0, btccommon.MaxNoOfInputsPerTx)
+	var total int64
+	for _, u := range utxos {
+		if len(selected) == btccommon.MaxNoOfInputsPerTx {
+			break
+		}
+		if u.AmountSats <= 0 || total+u.AmountSats > maxSats {
```
kingpinXD, line 218: The top-up here can silently kill the rehearsal sweep. Largest-first stops us picking dust *instead of* the big UTXO, but we keep scanning and add every smaller one that still fits, so we pick dust *in addition to* it.

The group check needs `fee <= inputs/10`, and the fee is close to fixed (171 vB for one input, +68 vB per extra), so a capped sweep only survives when the selected inputs total at least `1710 * feeRate` sats. A small cap sits right on that line. One 90,000 sat UTXO at 50 sat/vB gives fee 8,550 against an allowance of 9,000, and adding a single 300 sat input takes it to 11,950 against 9,030, so the group is dropped:

```
dust= 0 -> inputs=1 fee=8550 output=81450
dust= 1 -> NO SWEEP
dust=19 -> NO SWEEP
```

The operator gets an empty BTC section and one stderr line among many. The `--btc-max-sats 100000` in the PR description works at 10 sat/vB and produces nothing above roughly 58 sat/vB.

Two things would fix it: skip inputs that aren't worth their own marginal fee during selection, and make a capped run that yields zero BTC txs say so loudly with the minimum cap for the current fee rate. Worth doing because BTC is the leg with no Athens coverage, so a rehearsal that quietly skips it leaves us where we started.

cmd/zetatool/cli/drain.go
```diff
@@ -464,6 +573,7 @@ func buildEVMTxs(
 			Balance:        sdkmath.NewUintFromString(balance.String()),
 			MedianGasPrice: median,
 			Nonce:          nonce,
+			MaxAmount:      maxAmount,
```
kingpinXD, line 581: The rehearsal breaks nonce quiescence on purpose, and the real drain has no way back from it. The generator reads `NonceAt` (confirmed) at line 563, the poller reads `PendingNonceAt`. If the real payload freezes while a rehearsal tx is still unmined, we pin `C` and the poller sees `C+1`, which `executeEVM` treats as a hard stop, so that chain drops out for the whole window and only a third republish at a higher height recovers it.

The operator flow in the description has this right by saying to confirm the funds landed first, but "the second run needs no cleanup" undersells it. Waiting for confirmation is the gate, not tidiness.

Since we already call `GetEVMNonce` here, it would be good to also read the pending nonce and warn per chain when the two differ. Turns an invisible whole-window failure into a line the operator sees before signing.

pkg/drain/generate.go
kingpinXD: Correcting myself here, this is worse than "sits right on the line". I ran the caps against a UTXO set shaped like the one we measured on real mainnet (500 UTXOs, top 20 holding 99.75% of 1.77 BTC, the other 480 dust at roughly 931 sats):

```
feeRate= 10 cap=   100000 -> NO SWEEP
feeRate= 10 cap=  1000000 -> NO SWEEP
feeRate= 10 cap=  5000000 -> NO SWEEP
feeRate= 10 cap=  8839767 -> txs=1 inputs=1 in=8839767 out=8838057 fee=1710
feeRate= 50 cap=   100000 -> NO SWEEP
feeRate= 50 cap=  8839767 -> txs=1 inputs=1 in=8839767 out=8831217 fee=8550
UNCAPPED   feeRate=10     -> txs=1 inputs=20 out=176780710
```

The reason is the `continue` rather than a `break`. A big UTXO that blows the cap is skipped and the walk carries on down the list, so a small cap slides past all 20 large UTXOs and fills with dust instead. The loop only ever exits on the 20 input count, never on value. Those 20 dust inputs bring 18,620 sats and cost 15,310 in fee, against an allowance of 1,862, so the group is dropped and no BTC tx comes out.

That leaves two regimes and nothing useful between them. Below the smallest large UTXO you get no BTC transaction at all. At or above it, one whole large UTXO moves, which on those holdings is about 0.088 BTC. Neither one is a small value rehearsal.

Caveat on the evidence: I could not fetch the live UTXO list, so this is a synthetic set built from the figures we recorded during the mainnet live run, with the large UTXOs all equal. The real set will have more spread, so a mid sized cap may well catch something. The mechanism is the part that holds: the selection takes whatever sits below the cap, and if that is dust the sweep is dropped.

Skipping inputs worth less than the fee of adding them (around 680 sats at 10 sat/vB) fixes the dust half. Worth pairing it with a loud message when a capped run yields no BTC sweep, so this cannot pass for a rehearsal that covered BTC.

skosito: Confirmed both, and the second diagnosis is the right one — it's the `continue`, not a borderline threshold. Reproduced on a synthetic mainnet-shaped set (20 large ~8.83M, 480 dust at 931) before changing anything:

```
feeRate= 10 cap=  100000 -> NO SWEEP
feeRate= 10 cap= 1000000 -> NO SWEEP
feeRate= 10 cap= 5000000 -> NO SWEEP
feeRate= 50 cap=  100000 -> NO SWEEP
90k sat UTXO, dust= 0 -> inputs=1 fee=8550 output=81450
90k sat UTXO, dust= 1 -> NO SWEEP
90k sat UTXO, dust=19 -> NO SWEEP
```

Fixed in e8cddd4 by taking your first suggestion, generalised slightly: rather than a fixed "worth its own marginal fee" threshold, `selectCappedUTXOs` now admits an input only if the resulting group still satisfies the poller's own bound, `fee <= total/MaxBTCFeeFraction`, priced with `EstimateOutboundSize` at the count that input would produce. Tying it to the poller's constant rather than a separate heuristic means a non-empty result is economical by construction — the group check afterwards can no longer drop what selection produced.

Worth noting it isn't a blanket dust exclusion: an input that does cover its marginal share is still taken, so a 10M-sat anchor plus dust still rides along, and the 90k case now keeps its single input. Tests cover both directions, plus a sweep over `{10,50} sat/vB × {100k … 20M}` on the mainnet-shaped set asserting nothing emitted can violate `validateBTCFee`.

I did **not** apply the rule to the uncapped path. There, dust packed behind large UTXOs is swept for free while the ratio holds, which is what you want in a real drain — the capped rule is about guaranteeing a viable small sweep, not maximising coverage.

Second suggestion also in: `MinViableSweepSats(feeRate, receiver)` gives the threshold (≈`10 × feeRate × 171`), and a capped run yielding zero BTC txs now prints `REHEARSAL SWEPT NO BTC` with that number plus the smallest UTXO that could be swept alone — or, when no UTXO clears the threshold, states outright that no cap can rehearse BTC on these holdings and it can only be drained uncapped.

On your caveat about the synthetic set: agreed the real spread will be wider, and I've kept the fix mechanism-level rather than tuned to those figures. But it does mean the honest answer for a wallet shaped like that one is that the smallest BTC rehearsal available is one whole large UTXO (~0.088 BTC). The tool now says that instead of implying a smaller cap works — which is the part I think matters, given no Athens coverage on this leg.

cmd/zetatool/cli/drain.go
skosito: You're right, and I checked the two reads: `clients.GetEVMNonce` is `NonceAt(ctx, addr, nil)` (confirmed), while the poller's `executeEVM` compares against `PendingNonce`, with `live > tx.Nonce` returning the "pinned nonce already consumed" error. So an unmined rehearsal tx at freeze time costs that chain the entire firing window, and only a third republish recovers it.

Added `warnIfNonceNotQuiesced` in e8cddd4: since `GetEVMNonce` is already being called per chain, it now also reads the pending nonce and warns when they differ, naming the chain, both values and the number of in-flight txs — before the payload is signed. Non-fatal, since the confirmed nonce is already in hand and this is advisory.

Also fixed the wording. "The second run needs no cleanup" was wrong to imply confirmation is tidiness; the RFC now states it as a correctness gate, with the mechanism, and notes the rehearsal is what creates the non-quiescent condition the drain otherwise assumes. `--help` says it too.

One thing I deliberately did not do: switch the pinned nonce to `PendingNonceAt` so it matches what the poller checks. It would paper over the rehearsal case, but it changes real-drain behaviour by pinning a value derived from one node's mempool view while each poller validates against its own — a divergence that's invisible under true quiescence and hard to reason about when not. A warning keeps the failure visible without moving that risk into the uncapped path. Happy to switch if you'd rather, but I'd want it as its own change.

skosito:
@greptile

skosito:
@greptile

kingpinXD (approved):

ws4charlie (approved):

cmd/zetatool/cli/drain.go
```diff
@@ -88,6 +102,17 @@ tss-balances.`,
 	cmd.Flags().Duration(FlagInterval, 10*time.Second, "serve-mode republish interval")
 	cmd.Flags().String(FlagOnlyChains, "", "comma-separated chain IDs to drain (default: all with funds)")
 	cmd.Flags().String(FlagExcludeChains, "", "comma-separated chain IDs to skip")
+	cmd.Flags().String(
+		FlagEVMMaxAmount,
+		"",
+		"rehearsal only: cap the per-chain EVM transfer at this many wei (default: full balance)",
+	)
+	cmd.Flags().Int64(
+		FlagBTCMaxSats,
```
julianrubino, line 111: Minor inconsistency between the two cap flags: an explicit `--btc-max-sats 0` silently means "sweep ALL BTC" (0 = uncapped), while an explicit `--evm-max-amount 0` errors with "must be positive". An operator rehearsing under pressure who types `0` meaning "no BTC" gets the full BTC drain instead. `cmd.Flags().Changed(FlagBTCMaxSats)` would let you reject an explicit 0 the same way the EVM flag does. Non-blocking — funds still go to the controlled receiver either way — fine as a follow-up.

cmd/zetatool/cli/drain.go
```diff
@@ -542,9 +658,16 @@ func buildBTCTxs(
...
 	})
 	if err != nil {
-		return nil, err
+		// Same policy as every other BTC failure in this function: skip BTC loudly rather than
+		// abort, so a BTC-side problem never denies the EVM drain. Only the capped path can get
+		// here (the uncapped one skips uneconomical groups instead of erroring), and it would take
+		// a receiver address whose output size can't be estimated — impossible for the hardcoded
+		// anchors — but the two paths should not diverge on how they fail.
+		fmt.Fprintf(os.Stderr, "ERROR chain %d skipped: build sweeps failed: %v\n", btcChainID, err)
+		return nil, nil
```
julianrubino, line 670: Just confirming intent: this also changes the *uncapped* real drain — a `GenerateBTCTxs` error now publishes a final payload with an empty BTC section instead of aborting the tick. The comment's argument that the error is unreachable for the hardcoded anchors looks right to me, so this is fine; just making sure skip-not-abort is deliberate for the real-drain path too, not only the rehearsal.

cmd/zetatool/clients/evm.go
```diff
@@ -214,6 +214,19 @@ func GetEVMNonce(ctx context.Context, rpcURL string, address ethcommon.Address)
 	return client.NonceAt(ctx, address, nil)
 }
 
+// GetEVMPendingNonce fetches the account nonce including mempool txs on an EVM chain. The drain
+// poller checks the pinned nonce against this value, so a difference from GetEVMNonce means a TSS
+// tx is still in flight and a payload pinned now would be rejected at fire time.
+func GetEVMPendingNonce(ctx context.Context, rpcURL string, address ethcommon.Address) (uint64, error) {
+	client, err := ethclient.Dial(rpcURL)
```
julianrubino, line 221: nit: this is a second `ethclient.Dial` per chain per tick just to read the pending nonce — the call site already dialed for `GetEVMNonce`. Matches the existing one-dial-per-call style in this file, so fine as-is; only worth touching if serve-mode RPC load ever matters.

julianrubino (approved):
Read through the cap-selection logic and the earlier threads — the marginal-fee rule from e8cddd4 plus the boundary tests pin the BTC selection down nicely, and the nonce-quiescence warning covers the rehearsal→real-drain gap. LGTM.

A few minor notes inline; the `--btc-max-sats 0` handling is the only one I'd treat as a real (non-blocking) follow-up.

cmd/zetatool/cli/drain.go
skosito: Good catch, and you're right that it's the asymmetry that makes it dangerous — `0` is a plausible thing to type under pressure and it means the opposite of what the operator intended.

Fixed in #4630 (d4b1213) rather than a separate follow-up, since that PR was already open against the same file. `cmd.Flags().Changed()` distinguishes "typed 0" from "not passed", so an explicit `--btc-max-sats 0` now errors and points at `--exclude-chains` for actually skipping BTC:

```
--btc-max-sats must be positive; omit the flag to sweep all BTC, or use --exclude-chains to skip the BTC chain
```

Pulled it out as `parseBTCMaxSats(v, changed)` so the rule is unit-tested the same way `parseEVMMaxAmount` is, including the omitted-flag case that must stay uncapped.

cmd/zetatool/cli/drain.go
skosito: Deliberate, and worth pinning down precisely because the reachability argument is what makes it safe.

`GenerateBTCTxs` can only return a non-nil error from `selectCappedUTXOs`, which is only called when `in.MaxSats > 0`. In the uncapped real drain the error is not merely unlikely — it is unreachable, so that path still behaves exactly as before: `ComputeBTCSweep` failures are swallowed per group by the existing `continue`, and a group that can't be priced is skipped rather than aborting the tick. So skip-not-abort changes only capped runs today.

That said, your framing is the right one for *if* that ever changes. Aborting the tick is the safer default for the real drain, because `RunCron` retries and will keep retrying until the final publishes, whereas skipping publishes a final with an empty BTC section and only a stderr line to say so — and the empty-payload guard only refuses when EVM *and* BTC are both empty. I chose consistency with the surrounding failure policy ("a BTC-side problem must not deny the EVM drain") on the grounds that it's unreachable for the real drain; if a future change makes `GenerateBTCTxs` fallible when uncapped, this should flip back to aborting. Noted in the comment.

cmd/zetatool/clients/evm.go
skosito: Agreed on both counts — it's a second dial per chain per tick, and it matches the one-dial-per-call style of the file, so I've left it.

For the record on the load: in `--serve` mode that's 2 dials per EVM chain per `--interval` (10s default), so ~2N dials every 10 seconds. Not enough to matter for a drain window, and the alternative is either threading a client through `buildEVMTxs` or reworking this file's dial-per-call convention, neither of which I wanted to do in a change about value caps. Happy to revisit if serve-mode RPC pressure ever shows up.

PR #4635, opened 2026-08-25
test: assert vesting-account creation is rejected via group proposal

kingpinXD:
## Summary

We block vesting-account creation on ZetaChain, but the ante-level block (`VestingAccountDecorator`) only catches top-level txs. Governance-style wrappers (`authz.MsgExec`, `group.MsgSubmitProposal`) execute their inner messages through the msg router, which bypasses that decorator.

These tests pin the behavior that actually closes the `group` route: zeta's custom `AuthzLimiterDecorator` recurses into `group.MsgSubmitProposal` (and `authz.MsgExec`) inner messages and rejects any disabled type — including `MsgCreateVestingAccount` — at ante time, before the group module runs.

**What we check:**

- **Unit** (`app/ante/authz_test.go`): `AuthzLimiterDecorator` blocks `MsgCreateVestingAccount` / `MsgCreatePermanentLockedAccount` when wrapped in a group proposal, an authz exec, and a nested authz→group combination; a non-disabled message (bank send) in the same group proposal passes; a top-level vesting msg is not blocked here (that's `VestingAccountDecorator`'s job).
- **E2E** (`disallow_vesting_via_group_proposal`, admin test group): broadcasting a `group.MsgSubmitProposal` that wraps `MsgCreateVestingAccount` fails with `found disabled msg type`. No real group needs to exist — the ante rejects the tx first.

---

> [!NOTE]
> **Low Risk**
> Test-only changes; no modifications to ante handlers or production behavior.
> 
> **Overview**
> Adds **regression coverage** for the existing `AuthzLimiterDecorator` path that rejects disabled vesting-account creation messages when they are wrapped in **`group.MsgSubmitProposal`** or **`authz.MsgExec`** (including nested authz → group), without changing chain logic.
> 
> **Unit tests** in `app/ante/authz_test.go` table-drive ante handling: blocked cases expect `found disabled msg type`; allowed cases include a group proposal with a bank send and a top-level vesting msg (explicitly not this decorator’s responsibility).
> 
> **E2E** adds `disallow_vesting_via_group_proposal`: broadcasts a group proposal wrapping `MsgCreateVestingAccount` and asserts broadcast fails with the same error. The test is registered in the admin suite and local `--test-admin` run list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 375afcc79e6cb9c33745fcc84aa6806e3e988f53. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

The PR adds unit and end-to-end coverage confirming that disabled vesting-account creation messages are rejected when wrapped in group proposals or authz execution.
- Adds table-driven ante tests for direct, wrapped, and nested message shapes.
- Registers an admin E2E test that broadcasts a group proposal containing a vesting-account creation message.
- Adds the new E2E test to the local admin suite.

<h3>Confidence Score: 4/5</h3>

The PR appears safe to merge, with only a non-blocking gap in coverage for the periodic vesting-account message.

The added unit and E2E paths correctly target recursive group and authz inspection, but the unit table does not exercise one of the three disabled vesting message types declared by its fixture.

**Files Needing Attention:** app/ante/authz_test.go

| Filename | Overview |
|----------|----------|
| app/ante/authz_test.go | Adds focused recursive-wrapper tests, but omits a case for the periodic vesting message included in the disabled fixture. |
| e2e/e2etests/test_disallow_vesting_via_group_proposal.go | Adds a live broadcast assertion that specifically requires the expected ante rejection text. |
| e2e/e2etests/e2etests.go | Registers the new argument-free E2E test in the shared catalog. |
| cmd/zetae2e/local/local.go | Adds the new test to the local admin suite with no identified orchestration issue. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
    A[Signed transaction] --> B[AuthzLimiterDecorator]
    B --> C{Wrapper type}
    C -->|group proposal| D[Inspect proposal messages]
    C -->|authz exec| E[Inspect executed messages]
    D --> F{Disabled vesting type?}
    E --> F
    F -->|Yes| G[Reject during ante handling]
    F -->|No| H[Continue ante chain]
```

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
### Issue 1
app/ante/authz_test.go:30
**Periodic vesting case is untested**

The fixture includes `MsgCreatePeriodicVestingAccount` among the disabled types, but the table only wraps regular and permanent-locked vesting messages. Add a periodic-vesting case so this suite cannot pass without verifying the claimed rejection for all three disabled vesting-account creation types.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test: assert vesting-account creation is..."](https://github.com/zeta-chain/node/commit/375afcc79e6cb9c33745fcc84aa6806e3e988f53) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56542997)</sub>

> Greptile also left **1 inline comment** on this PR.

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Application composition and transaction execution](https://app.greptile.com/zetachain/-/custom-context/knowledge-base/zeta-chain/node/-/docs/application-composition.md)
- Knowledge Base — [End-to-end testing workflows](https://app.greptile.com/zetachain/-/custom-context/knowledge-base/zeta-chain/node/-/docs/end-to-end-testing.md)
- Knowledge Base — [Testing and developer tooling](https://app.greptile.com/zetachain/-/custom-context/knowledge-base/zeta-chain/node/-/docs/testing-and-developer-tooling.md)
</details>

### DIFF
--- a/app/ante/authz_integration_test.go
+++ b/app/ante/authz_integration_test.go
@@ -1,0 +1,94 @@
+package ante_test
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"cosmossdk.io/log"
+	dbm "github.com/cosmos/cosmos-db"
+	simtestutil "github.com/cosmos/cosmos-sdk/testutil/sims"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	vesting "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
+	"github.com/cosmos/cosmos-sdk/x/group"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zeta-chain/node/app"
+	"github.com/zeta-chain/node/app/ante"
+	serverconfig "github.com/zeta-chain/node/server/config"
+	zetasimulation "github.com/zeta-chain/node/simulation"
+	"github.com/zeta-chain/node/testutil/sample"
+)
+
+// TestProductionAnteHandler_RejectsVestingViaGroupProposal is an in-process integration check that
+// wires the real app keepers into the full production ante chain (via app.DisabledAuthzMsgs) and
+// runs a signed group.MsgSubmitProposal that wraps MsgCreateVestingAccount through it. It asserts
+// the tx is rejected — the same outcome the localnet e2e test (disallow_vesting_via_group_proposal)
+// verifies over the wire, but executable without Docker.
+func TestProductionAnteHandler_RejectsVestingViaGroupProposal(t *testing.T) {
+	// ARRANGE
+	zetaApp, err := zetasimulation.NewSimApp(
+		log.NewNopLogger(),
+		dbm.NewMemDB(),
+		simtestutil.AppOptionsMap{},
+	)
+	require.NoError(t, err)
+
+	encCfg := app.MakeEncodingConfig(serverconfig.DefaultEVMChainID)
+
+	// build the ante handler exactly as app.New does, with the real disabled-msg list
+	anteHandler, err := ante.NewAnteHandler(ante.HandlerOptions{
+		AccountKeeper:     zetaApp.AccountKeeper,
+		BankKeeper:        zetaApp.BankKeeper,
+		EvmKeeper:         zetaApp.EvmKeeper,
+		FeeMarketKeeper:   zetaApp.FeeMarketKeeper,
+		SignModeHandler:   encCfg.TxConfig.SignModeHandler(),
+		SigGasConsumer:    ante.DefaultSigVerificationGasConsumer,
+		MaxTxGasWanted:    0,
+		DisabledAuthzMsgs: app.DisabledAuthzMsgs(),
+		ObserverKeeper:    zetaApp.ObserverKeeper,
+	})
+	require.NoError(t, err)
+
+	testPrivKey, testAddress := sample.PrivKeyAddressPair()
+	_, testAddress2 := sample.PrivKeyAddressPair()
+
+	vestingMsg := vesting.NewMsgCreateVestingAccount(
+		testAddress, testAddress2,
+		sdk.NewCoins(sdk.NewInt64Coin("azeta", 100_000_000)),
+		time.Date(2100, 1, 1, 0, 0, 0, 0, time.UTC).Unix(),
+		false,
+	)
+	proposalMsg, err := group.NewMsgSubmitProposal(
+		sample.AccAddress(),
+		[]string{testAddress.String()},
+		[]sdk.Msg{vestingMsg},
+		"",
+		group.Exec_EXEC_UNSPECIFIED,
+		"title",
+		"summary",
+	)
+	require.NoError(t, err)
+
+	tx, err := simtestutil.GenSignedMockTx(
+		rand.New(rand.NewSource(1)),
+		encCfg.TxConfig,
+		[]sdk.Msg{proposalMsg},
+		sdk.NewCoins(),
+		simtestutil.DefaultGenTxGas,
+		"testing-chain-id",
+		[]uint64{0},
+		[]uint64{0},
+		testPrivKey,
+	)
+	require.NoError(t, err)
+
+	ctx := zetaApp.NewContext(true)
+
+	// ACT
+	_, err = anteHandler(ctx, tx, false)
+
+	// ASSERT: the ante decorator rejects the group-wrapped vesting-create before the group module runs
+	require.Error(t, err)
+	require.ErrorContains(t, err, "found disabled msg type")
+}
--- a/app/ante/authz_integration_test.go
+++ b/app/ante/authz_integration_test.go
@@ -1,3 +1,9 @@
+//go:build test
+
+// This file builds a full in-process app via zetasimulation.NewSimApp, which requires the `test`
+// build tag; without it the EVM configurator panics. The tag keeps a bare `go test ./app/ante/...`
+// (no -tags) from taking the rest of the package down with a confusing panic.
+
 package ante_test
 
 import (
@@ -36,18 +42,11 @@ func TestProductionAnteHandler_RejectsVestingViaGroupProposal(t *testing.T) {
 
 	encCfg := app.MakeEncodingConfig(serverconfig.DefaultEVMChainID)
 
-	// build the ante handler exactly as app.New does, with the real disabled-msg list
-	anteHandler, err := ante.NewAnteHandler(ante.HandlerOptions{
-		AccountKeeper:     zetaApp.AccountKeeper,
-		BankKeeper:        zetaApp.BankKeeper,
-		EvmKeeper:         zetaApp.EvmKeeper,
-		FeeMarketKeeper:   zetaApp.FeeMarketKeeper,
-		SignModeHandler:   encCfg.TxConfig.SignModeHandler(),
-		SigGasConsumer:    ante.DefaultSigVerificationGasConsumer,
-		MaxTxGasWanted:    0,
-		DisabledAuthzMsgs: app.DisabledAuthzMsgs(),
-		ObserverKeeper:    zetaApp.ObserverKeeper,
-	})
+	// build the ante handler from the exact options app.New installs, so dropping any field there
+	// (e.g. DisabledAuthzMsgs) fails this test rather than silently passing
+	anteHandler, err := ante.NewAnteHandler(
+		app.AnteHandlerOptions(zetaApp, encCfg.TxConfig.SignModeHandler()),
+	)
 	require.NoError(t, err)
 
 	testPrivKey, testAddress := sample.PrivKeyAddressPair()
--- a/app/ante/authz_test.go
+++ b/app/ante/authz_test.go
@@ -81,6 +81,15 @@ func TestAuthzLimiter_AnteHandle(t *testing.T) {
 		msg := authz.NewMsgExec(testAddress, []sdk.Msg{inner})
 		return &msg
 	}
+	authzGrant := func(t *testing.T, msgTypeURL string) sdk.Msg {
+		msg, err := authz.NewMsgGrant(
+			testAddress, testAddress2,
+			authz.NewGenericAuthorization(msgTypeURL),
+			nil,
+		)
+		require.NoError(t, err)
+		return msg
+	}
 	govProposal := func(t *testing.T, inner sdk.Msg) sdk.Msg {
 		msg, err := govv1.NewMsgSubmitProposal(
 			[]sdk.Msg{inner},
@@ -122,6 +131,12 @@ func TestAuthzLimiter_AnteHandle(t *testing.T) {
 		{
 			"authz exec wrapping a vesting msg is blocked",
 			authzExec(createVestingMsg),
+			true,
+			"found disabled msg type",
+		},
+		{
+			"authz grant of a vesting-create authorization is blocked",
+			authzGrant(t, sdk.MsgTypeURL(&vesting.MsgCreateVestingAccount{})),
 			true,
 			"found disabled msg type",
 		},
--- a/app/ante/authz_test.go
+++ b/app/ante/authz_test.go
@@ -10,6 +10,7 @@ import (
 	vesting "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
 	"github.com/cosmos/cosmos-sdk/x/authz"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	govv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
 	"github.com/cosmos/cosmos-sdk/x/group"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: app/ante/authz_test.go [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+			"found disabled msg type",
+		},
+		{
+			"group proposal wrapping a non-disabled msg is allowed",
+			groupProposal(t, bankSend),
+			false,
+			"",
+		},
+		{
+			"top-level vesting msg is not blocked here (VestingAccountDecorator owns that)",
```
skosito, line 156: Worth a line noting gov is intentionally out of scope: `AuthzLimiterDecorator` has no `gov.MsgSubmitProposal` case, so a governance proposal can still carry a vesting-create (privileged/public/self-funded). Otherwise this reads as "all proposal paths blocked" when gov isn't one of them.

app/ante/authz_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/zeta-chain/node/app"
+	"github.com/zeta-chain/node/app/ante"
+	serverconfig "github.com/zeta-chain/node/server/config"
+	"github.com/zeta-chain/node/testutil/sample"
+)
+
+// disabledMsgs mirrors app.go's DisabledAuthzMsgs for the vesting entries: the three
+// vesting-account creation messages must not be executable indirectly.
+func disabledVestingMsgs() []string {
```
skosito: This hardcodes its own disabled list, so it verifies the decorator but not the wiring: dropping the vesting entries from `app.go`'s `DisabledAuthzMsgs` would leave this green while the chain regresses. Consider asserting the real `HandlerOptions.DisabledAuthzMsgs` contains the three types (also removes the app.go mirror).

skosito (approved):
Approving — confirms group/authz-wrapped vesting creation is blocked at ante, including nested, with a negative case and a real e2e broadcast. Two non-blocking notes inline (gov path is out of scope; test hardcodes the disabled list rather than asserting app.go's wiring).

app/ante/authz_test.go
```diff
@@ -0,0 +1,151 @@
...
+	return []string{
+		sdk.MsgTypeURL(&vesting.MsgCreateVestingAccount{}),
+		sdk.MsgTypeURL(&vesting.MsgCreatePermanentLockedAccount{}),
+		sdk.MsgTypeURL(&vesting.MsgCreatePeriodicVestingAccount{}),
+	}
+}
```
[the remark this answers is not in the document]
kingpinXD: Fixed in dd63870 — added a `MsgCreatePeriodicVestingAccount` group-proposal case, so all three disabled vesting types are now exercised.

app/ante/authz_test.go
kingpinXD: Done in dd63870 — added a case asserting a `gov.MsgSubmitProposal` wrapping a vesting msg is NOT blocked by this decorator, with a comment explaining gov is intentionally out of scope (privileged, self-funded, requires a passed vote).

app/ante/authz_test.go
kingpinXD: Done in dd63870 — exported `app.DisabledAuthzMsgs()` as the single source of truth used by the ante HandlerOptions, removed the hardcoded mirror, and added `TestDisabledAuthzMsgs_WiresVestingCreation` asserting the real list contains all three vesting types. Also added an in-process integration test that runs the real production ante handler end to end.

cmd/zetae2e/local/local.go
```diff
@@ -434,6 +434,7 @@ func localE2ETest(cmd *cobra.Command, _ []string) {
 			e2etests.TestCriticalAdminTransactionsName,
 			e2etests.TestUpdateOperationalChainParamsName,
 			e2etests.TestBurnFungibleModuleAssetName,
+			e2etests.TestDisallowVestingViaGroupProposalName,
```
morde08, line 437: **The new e2e test never runs on this PR.** `gh pr checks 4635` shows `start-e2e-admin-test / e2e -> skipping` and `start-upgrade-test-admin / e2e -> skipping`; `.github/workflows/e2e.yml:160` gates `ADMIN_TESTS` on a PR label and this PR carries only `breaking:cli`. The admin suite otherwise runs only on `schedule` (nightly) and `release/*` pushes, so a broken `disallow_vesting_via_group_proposal` would first surface in a nightly run *after* merge.

Please add the `ADMIN_TESTS` label and let the admin suite go green once before merging.

Worth knowing separately: even with the label, the `merge_group` branch at `e2e.yml:176-188` fetches `pr_labels` and then never calls `core.setOutput` with them, so the label only takes effect on the `pull_request` event. That's pre-existing, not this PR.

app/ante/authz_integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+		log.NewNopLogger(),
+		dbm.NewMemDB(),
+		simtestutil.AppOptionsMap{},
+	)
+	require.NoError(t, err)
+
+	encCfg := app.MakeEncodingConfig(serverconfig.DefaultEVMChainID)
+
+	// build the ante handler exactly as app.New does, with the real disabled-msg list
+	anteHandler, err := ante.NewAnteHandler(ante.HandlerOptions{
```
morde08: This closes half the wiring gap. `app.DisabledAuthzMsgs()` is now the single source for the *list*, but the test rebuilds `ante.HandlerOptions` by hand rather than exercising the handler `app.New` actually installs — so `New` can stop passing the field entirely and nothing notices.

Mutation-verified: changing `app/app.go` to `DisabledAuthzMsgs: nil` in the `options` literal leaves the whole `app/ante` package green (`ok github.com/zeta-chain/node/app/ante 1.319s`).

Cheapest fix is to hoist the options construction so both callers share it:

```go
// app/app.go
func AnteHandlerOptions(app *App, signModeHandler *txsigning.HandlerMap) ante.HandlerOptions {
    return ante.HandlerOptions{ /* ... */ DisabledAuthzMsgs: DisabledAuthzMsgs(), /* ... */ }
}
```

then `New` and this test both call it, and dropping the field in one place fails here.

app/ante/authz_integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// TestProductionAnteHandler_RejectsVestingViaGroupProposal is an in-process integration check that
+// wires the real app keepers into the full production ante chain (via app.DisabledAuthzMsgs) and
+// runs a signed group.MsgSubmitProposal that wraps MsgCreateVestingAccount through it. It asserts
+// the tx is rejected — the same outcome the localnet e2e test (disallow_vesting_via_group_proposal)
+// verifies over the wire, but executable without Docker.
+func TestProductionAnteHandler_RejectsVestingViaGroupProposal(t *testing.T) {
+	// ARRANGE
+	zetaApp, err := zetasimulation.NewSimApp(
```
morde08, line 36: `zetasimulation.NewSimApp` requires the `test` build tag, so this makes the whole `app/ante` package tag-gated. A bare `go test ./app/ante/...` now **panics** rather than failing one test, taking `TestIsSystemTx`, `TestVesting_AnteHandle` and `TestSystemTxPriorityDecorator_AnteHandle` down with it:

```
panic: this is only implemented with the 'test' build flag. Make sure you're running your tests using the '-tags=test' flag.
  github.com/cosmos/evm/x/vm/types.(*EVMConfigurator).ResetTestConfig(...)
  github.com/zeta-chain/node/simulation.NewSimApp(...) simulation/simulation.go:29
  app/ante_test.TestProductionAnteHandler_RejectsVestingViaGroupProposal(...) authz_integration_test.go:30
```

CI is fine — `TEST_BUILD_FLAGS := -tags pebbledb,ledger,test` (Makefile:49) — and the package is green under `make test`. But `app/ante` was tag-free before this PR, so anyone running the package directly hits a confusing panic. Consider `//go:build test` on this file so it's cleanly excluded instead of panicking, or a note in the test doc comment.

app/ante/authz_integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// build the ante handler exactly as app.New does, with the real disabled-msg list
+	anteHandler, err := ante.NewAnteHandler(ante.HandlerOptions{
+		AccountKeeper:     zetaApp.AccountKeeper,
+		BankKeeper:        zetaApp.BankKeeper,
+		EvmKeeper:         zetaApp.EvmKeeper,
+		FeeMarketKeeper:   zetaApp.FeeMarketKeeper,
+		SignModeHandler:   encCfg.TxConfig.SignModeHandler(),
+		SigGasConsumer:    ante.DefaultSigVerificationGasConsumer,
+		MaxTxGasWanted:    0,
```
morde08: `MaxTxGasWanted: 0`, but `app.New` passes `TransactionGasLimit` (`app/app.go:799`). Doesn't change the outcome here (rejection happens at decorator #2, long before `GasWantedDecorator`), but it makes the "exactly as app.New does" comment on line 39 inaccurate. Using `app.TransactionGasLimit` costs nothing and keeps the claim true.

e2e/e2etests/test_disallow_vesting_via_group_proposal.go
```diff
@@ -0,0 +1,49 @@
...
+		[]sdk.Msg{vestingMsg},
+		"",
+		group.Exec_EXEC_UNSPECIFIED,
+		"create vesting account",
+		"attempt to create a vesting account via a group proposal",
+	)
+	require.NoError(r, err)
+
+	// broadcasting must fail: the ante decorator blocks the disabled inner message
+	_, err = r.ZetaTxServer.BroadcastTx(utils.OperationalPolicyName, proposalMsg)
```
morde08: This burns ~25-30s retrying a failure that is deterministic by construction. `BroadcastTx` wraps everything in `retry.DoTypedWithBackoff` with `NewConstantBackOff(5*time.Second)` + `WithMaxRetries(bo, 5)` (`e2e/txserver/zeta_tx_server.go:274-275`), and `broadcastWithBlockTimeout`'s error is marked retryable at line 315 — so the expected ante rejection is broadcast 6 times, 5s apart, with six `broadcast failed:` lines in the log before the assertion ever sees it.

This is the first e2e test in the repo to assert a *failing* broadcast, so there's no precedent to lean on. A `BroadcastTxWithoutRetry` (or `BroadcastTxExpectFailure`) helper on `ZetaTxServer` would keep the negative path fast and the logs honest, and would be reusable for the next one.

app/ante/authz_test.go
```diff
@@ -0,0 +1,186 @@
...
+			"",
+			"title",
+			"summary",
+			false,
+		)
+		require.NoError(t, err)
+		return msg
+	}
+
+	tests := []struct {
```
morde08, line 107: The `authz.MsgGrant` branch (`app/ante/authz.go:74-82`) is still untested. That's the grant-time check — it stops the authorization from ever being created, so it's the first line of defense on the authz route, and deleting it leaves this table green.

One more row closes it:

```go
{
    "authz grant of a vesting msg authorization is blocked",
    func() sdk.Msg {
        msg, err := authz.NewMsgGrant(
            testAddress, testAddress2,
            authz.NewGenericAuthorization(sdk.MsgTypeURL(&vesting.MsgCreateVestingAccount{})),
            nil,
        )
        require.NoError(t, err)
        return msg
    }(),
    true,
    "found disabled msg type",
},
```

morde08 (commented):
Reviewed at `dd63870`, after the review-address commit. I checked the branch out, ran the suite, and mutation-tested each new test against the production code.

**Verified working:** `TestAuthzLimiter_AnteHandle` genuinely binds — I removed the recursion from the `*group.MsgSubmitProposal` case in `app/ante/authz.go` and 4 subtests went red. This is the first coverage `AuthzLimiterDecorator` has had since #794 (2023). The e2e mechanics also check out: `AuthzLimiterDecorator` is ante decorator #2, ahead of `ValidateBasicDecorator`, so the tx really does fail on `found disabled msg type` and not on the arbitrary group-policy address; `BroadcastSync` surfaces `RawLog` and `errRetryable` delegates `Error()`, so `ErrorContains` will match. The gov out-of-scope case and the periodic-vesting case both landed well.

Six things below. The first is the one I'd block on — everything else is hardening.

cmd/zetae2e/local/local.go
kingpinXD: Added the `ADMIN_TESTS` label to the PR so the admin suite runs on this `pull_request` event. The `merge_group` gap you flagged is pre-existing and out of scope here.

app/ante/authz_integration_test.go
kingpinXD: Fixed in 49b8afe. Extracted `app.AnteHandlerOptions(app, signModeHandler)` as the single construction; `New` and the integration test both call it now, so dropping `DisabledAuthzMsgs` (or any field) in one place fails the test.

app/ante/authz_integration_test.go
kingpinXD: Fixed in 49b8afe. Added `//go:build test` to the file so a bare `go test ./app/ante/...` cleanly excludes it instead of panicking — confirmed `TestVesting_AnteHandle`/`TestIsSystemTx` run green without the tag.

app/ante/authz_integration_test.go
kingpinXD: Resolved by the hoist above — the test now builds options via `app.AnteHandlerOptions`, which sets `MaxTxGasWanted: TransactionGasLimit`, so the "exactly as app.New does" claim is literally true.

e2e/e2etests/test_disallow_vesting_via_group_proposal.go
kingpinXD: Fixed in 49b8afe. Added `BroadcastTxWithoutRetry` on `ZetaTxServer` (shares a new `broadcastOnce` with `BroadcastTx`) and used it here — the deterministic rejection no longer burns the 25s retry loop or floods the log.

app/ante/authz_test.go
kingpinXD: Fixed in 49b8afe. Added an `authz.MsgGrant` row that grants a `MsgCreateVestingAccount` authorization and asserts `found disabled msg type`, covering the grant-time branch.

morde08 (approved):

